PR #3, opened 2023-06-06
Use Ratios.jl for faster rationals

MilesCranmer:
Turns out the built-in `Rational{Int}` in Julia is fairly slow because it computes `gcd` constantly. This switches to @timholy's `Ratios.SimpleRatio{Int}` instead, which is faster, but at a slightly higher risk of overflow.


MilesCranmer:
>10x speedup on integer powers on physical quantities! Nice! Especially since that seemed to be a bottleneck, that is really nice to have that speedup.

PR #7, opened 2023-06-07
Regular `Rational{Int8}` for dimensions

MilesCranmer:
@oscardssmith

I'm curious to see the speed of a normal Rational{Int8}. Even though it would be computing `gcd` frequently, maybe it's worth it.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,9 +4,7 @@ authors = ["MilesCranmer <miles.cranmer@gmail.com> and contributors"]
 version = "0.1.0"
 
 [deps]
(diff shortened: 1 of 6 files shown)


MilesCranmer:
@oscardssmith another thing I realized is that `Rational{Int8}` is so small, you could literally create a lookup table of every single operation – such that the final results are already in the lowest common denominator. The lookup table for `Base.*` would be 256^2 items = 65536 items.

MilesCranmer:
Nevermind, a lookup table is literally slower...

PR #9, opened 2023-06-07
Cache `divgcd(::Int8, ::Int8)`

MilesCranmer:
This creates `cached_divgcd` with a specialized method for `Int8` which caches all results at compilation time. Can get a 4x speedup on vectorized computations with normal Rational{Int8}.

This defines a `CRational` struct that uses this `cached_divgcd` for relevant rational calculations

@oscardssmith

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,9 +4,7 @@ authors = ["MilesCranmer <miles.cranmer@gmail.com> and contributors"]
 version = "0.1.0"
 
 [deps]
-Ratios = "c84ed2f1-dad5-54f0-aa8e-dbefe2724439"
 Requires = "ae029012-a4dd-5104-9daa-d747884805df"
-SaferIntegers = "88634af6-177f-5301-88b8-7819386cfa38"
 
 [weakdeps]
 Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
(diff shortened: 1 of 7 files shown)


oscardssmith:
I've convinced myself that the right type here is actually a `Int8/6`, since the math then becomes trivial and doing a search has not lead me to any use-cases that need a power that isn't a multiple of 1/6th.

MilesCranmer:
Wow, check out the speeds here! So caching the gcd is enough to make normal `Rational{Int8}` fast. Then it is much safer from overflow too, because it always reduces to lowest coprimes.

MilesCranmer:
I do think the best option would be to parametrize the Dimensions type by this choice. And just have the default be `Rational{Int}` I can probably get away with Int8/6 as well in my stuff, though I’m sure some people will have some weird system where they need large rational powers… For example, constants associated with a polytropic index in astrophysics can pretty much have any rational power on the units you can write down.

oscardssmith:
The type piracy here isn't great but the performance is. One improvement here is that you can make the table 2x smaller by making `x` be smaller than `y` (since `gcd` commutes).

MilesCranmer:
Surprisingly it is slower. I guess the `<(::Int8,: :Int8)` incurs more of a penalty than just the lookup?

oscardssmith:
will do

MilesCranmer:
Superseded by the faster & cleaner #21

Issue #10, opened 2023-06-07
Add a scale factor dimension?


odow:
We're looking into adding this to JuMP: https://github.com/trulsf/UnitJuMP.jl/issues/18

One kicker is that it'd be nice to be able to talk about micro/milli/kilo/mega etc.

Have you thought about adding a `scale::Int` dimension? Hopefully it'd support something like:
```julia
km = Dimension(length = 1, scale = 3)
us = Dimension(time = -1, scale = -6)
km / us == Dimension(length = 1, time = -2, scale = 9)
```

j-fu:
I think one could use Unitful as a comprehensive database for defining units and provide an API like:
```
@dynquantity km us 
```
etc. I set up something like this in [LessUnitful.jl](https://j-fu.github.io/LessUnitful.jl/stable)

j-fu:
After thinking more about your proposal - may be this problem can be solved differently. After seeing some discussions around the pacakge, I think for performance reasons, is critical to keep the size of a quantity small. So the way forward may be to define the basic units  `m`, `kg`, `s` etc. as constants and define `km=1000*m` etc.

MilesCranmer:
Thanks for the interest @odow, I'm happy to hear you are considering using it!

Just to help me understand the feature request, why couldn't you implicitly include the scale parameter inside the value of the quantity? Since the conversion from Unitful to DynamicQuantities uses `upreferred`, it will already propagate scales from units (milli, kilo, etc) into the value. For example:

```julia
julia> x_km = 1.5u"km"
1.5 km

julia> x = 1.5u"m"
1.5 m

julia> x_km_dyn = convert(DynamicQuantities.Quantity, x_km)
1500.0 𝐋 ¹

julia> x_dyn = convert(DynamicQuantities.Quantity, x)
1.5 𝐋 ¹
```

MilesCranmer:
One other tricky thing about having a separate `scale` factor is that it is technically redundant information. You could have two quantities, say

```julia
x = Quantity(0.2, length=1, scale=2)
y = Quantity(0.02, length=1, scale=3)
```
and checking equality is no longer comparing `x.val == y.val`, but it instead would require `x.val * 10^(x.scale) == y.val^(y.scale)`.

If you have a `Quantity{Float64}`, then the `scale` factor would be like if you had extra bits for the exponent. But why not just use `Float128` instead? Or define another type that includes more bits for scaling?

odow:
[...]

PR #11, opened 2023-06-08
Propose some functions and macros for unitful integration.

j-fu:
In particular this PR provides a `q_str` macro which accesses `u_str` to define quantities via the corresponding data in Unitful.

I don't know how to define the macros in the extension, so at this stage it would make Unitful a full dependency. IMHO would be ok, but of course this is open for discussion.

See my proposal in #10 - this PR could be an implementation.


MilesCranmer:
Thanks for the PR! I am in principle on board with this, just need to consider things a bit more. I am wondering whether it makes sense to make Unitful.jl a hard dependency or not. According to the benchmark above, it bumps the load time to 0.6s. After #7, the load time would go down to 0.07s, so this would be a big increase if we do choose that route.

@oscardssmith thoughts?

I definitely like the idea of exporting a `q_str` though. I think you could export it from an extension by defining a macro that calls some function `_q_str`. Then overload that function inside the extension.

If we break the target audience into two groups, I see the following use-cases:

1. Internal use by package maintainers.
  - They might want to eliminate reliance on Unitful.jl completely, and just manipulate dimensions. Here, perhaps we just want to reduce load time of the package, and simplify things as much as possible.
  - I think this will be the main user base?
2. End users (people doing scientific calculations/analysis).
  - They probably want to use units.
  - They may want to try DynamicQuantities.jl for faster calculations.
  - Having them install both DynamicQuantities.jl and Unitful.jl (as well as importing both) would make things slightly less user-friendly.

In talking through this, I wonder if it makes the most sense to just have people import both packages, and convert between them. In other words, maybe it makes sense to keep Unitful.jl as an extension, without the `q_str` macro. Then they can still rely on the Unitful.jl documentation and other unit libraries, rather than treating DynamicQuantities.jl as a front-end API and expecting the relevant docs to be here. What do you think?

MilesCranmer:
@odow I'd be interested to hear your thoughts as well

j-fu:
Interesting considerations! 

My opinion here is that Unitful.jl is too rigid to be used by end users (who ase a rule are inexperienced programers and don't see programming as their foremost purpose). `km` and `cm` being different types feels IMHO is not very intuitive. I am working  with others on a PDE simulation stack for applicationsm and I really would like to provide the possibility to do calculations with units to end users and modelers. With Unitful one needs to strip units before going into the solver (and not forget upreferred - I once [tried](https://github.com/PainterQubits/Unitful.jl/pull/448) to get some convenience calls which should have made it easier for end users into Unitful). 
For me this ended up in [LessUnitful.jl](https://j-fu.github.io/LessUnitful.jl/stable/) which is just a band aid.

For `DyanamicQuantities` I indeed see the potential that this package can work well for less experienced end users. I see the potential that quanties can be passed through e.g. a PDE solver if it is well designed. This opens up great possibilities for simulation work, teaching etc.

I see the need for being able to expose end users to quantities with units  in an easy to handle consistent way, Julia and its ecosystem certainly have the potential, and DynamicQuanties could make it possible.  So indeed I would put them first...

j-fu:
Ok the wind blew the other direction, so I think we can close this.

PR #18, opened 2023-06-09
Modifications for linear algebra

j-fu:
Figure out what is needed to solve linear systems with dynamic quantities.


j-fu:
Not yet ready for merge.

We might need:
- support adding/subtracting from zero(Quantity). 
- should abs be stripped of dimensions  ? this would be needed for pivoting. However one could use RowNonZero pivoting. This needs to be decided.
-  How to route to the right implementation for dense linalg ?

j-fu:
Sparspak works with https://github.com/PetrKryslUCSD/Sparspak.jl/pull/25

src/math.jl
```diff
@@ -16,8 +16,28 @@ Base.:/(l::Number, r::Quantity) = l * inv(r)
 Base.:/(l::Dimensions, r::Number) = Quantity(inv(r), l, true)
 Base.:/(l::Number, r::Dimensions) = Quantity(l, inv(r), true)
 
-Base.:+(l::Quantity, r::Quantity) = Quantity(l.value + r.value, l.dimensions, l.valid && r.valid && l.dimensions == r.dimensions)
-Base.:-(l::Quantity, r::Quantity) = Quantity(l.value - r.value, l.dimensions, l.valid && r.valid && l.dimensions == r.dimensions)
+function Base.:+(l::Quantity, r::Quantity)
+    if iszero(l)
```
MilesCranmer, line 20: Wouldn't you want a DimensionError to still be raised even if `iszero` is true? If adding a 2D vector to a 3D vector, you would always want to raise an error, regardless of if one of them was zero, no?

It might not even be the case that `iszero` is defined for all possible `Quantity`, since the value of a `Quantity` might be an abstract object, rather than just a scalar real

MilesCranmer:
Not sure if this is relevant, but I've been experimenting with this "WildcardDimensionWrapper" type in SymbolicRegression.jl: https://github.com/MilesCranmer/SymbolicRegression.jl/blob/6ed49f166483ec130bc0350d69d6e2214666aeda/ext/SymbolicRegressionUnitfulExt.jl#L58-L69

It's basically a way of adding a regexp to a `Quantity`, so that you can solve for unknown constants (with their own physical dimensions).

The way it works is as follows. Any free constant (here: `a` or `b`) produces a "wildcard" (regexp Quantity) that can cancel out or create physical dimensions wherever necessary.
- `*`/`/`
  - Always valid, and propagates the wildcard (since `a*x` has arbitrary dimension, no matter the dimensions of `x`).
- `+`/`-`
  - No wildcards => need same dimension
  - 1 wildcard => always valid, but loses the wildcard
  - 2 wildcards => always valid, and propagates a wildcard (since `a*x + b*y` is of arbitrary physical dimensions)
- `^`
  - No wildcard in power => need zero dimension in power
  - Wildcard in power => Always valid, but loses the wildcard
  - (Always propagates wildcard from base, since (`(a*x)^(...)` if of arbitrary physical dimension)

`sqrt/cbrt/abs` propagate wildcards.

Any other operator, if undefined for `Quantity`, would consume a wildcard. So you can have things like `cos(a*x)` be dimensionally valid for any `x`, because a free constant `a` produces a wildcard.

src/math.jl
j-fu: The use case is that there are places where a value is initialized with zero, and than += or -= are used to add other quantities to it.

This can be solved also e.g. in Sparspak. But IMHO it this belongs to the algebraic properties of Quantity. What is `zero` and what is `one` ?

MilesCranmer: It looks like other unit packages also raise DimensionErrors even if the value is zero. e.g., astropy:

```python
from astropy import units as u

q = 10 * u.m # 10 meters
q2 = 10 * u.s # 10 seconds

q * 0 + q2
# ^ Raises UnitConversionsError
```

likewise with Unitful.jl

```julia
julia> using Unitful

julia> 0.0u"m" + 10.0u"s"
ERROR: DimensionError: 0.0 m and 10.0 s are not dimensionally compatible.
```

MilesCranmer: I think "zero meters" + "one second" raising an error makes the most sense as a default? Thoughts?

j-fu: I had it running like this, so for sparspak I certainly could switch to this. But sparspak may not be the only place where this functionality may be needed.

Fundamentally this leads to the question how units are working algebraically. I'll try to think more about it to find a less ad-hoc approach. 

IMHO it would be good if DynamicQuantities could be used like Dual numbers, MultiFloats, Measurements and the like as a drop-in replacement for Float64. 
But I am aware that this may be not everyone's version for this package.

j-fu:
Just got the following idea: DynamicalQuantities provides
an abstract type AbstractDynamicalQuantity such that DynanmicalQuantities.Quantity <: AbstractDynamicalQuantity. All operations in math.jl would be defined for  AbstractDynamicalQuantity,
and the modifications for linear algebra  I am proposing could be implemented  for another subtype with some room for experimentation.

MilesCranmer:
I think that’s a good idea — though perhaps just `AbstractQuantity`. And I suppose `AbstractDimension` would be nice as well in case someone wants to add or remove specific dimensions.

src/math.jl
MilesCranmer: I think even if there are a few additional packages that need this, it is better left out of the main package, because it will slow down every call to `+` with 3x as many comparisons on the dimension objects.

j-fu:
Possibly can be closed in favor of #24.

MilesCranmer:
Maybe we can put the `XQuantity` you are interested in inside a unittest for the abstract interface?

MilesCranmer:
I think most of these are superseded or already implemented now? Closing to clean the repo up

PR #22, opened 2023-06-12
Create in-house units module

MilesCranmer:
This adds an in-house units module, so now you can parse a wide variety of base and derived SI units without relying on Unitful.jl. It includes a few non SI units: min, h, day, year - but I purposefully didn't go too far beyond that, to avoid potential namespace ambiguities. I did not include any Gaussian units (like `erg` - which is used in astronomy), because Gaussian units require different formulas and I feel like mixing the unit systems might lead to confusion.

This PR also makes it so that all physical quantities are displayed in SI base units "m¹ᐟ² kg¹ mol³", rather than the unevaluated "𝐋 ¹ᐟ² 𝐌 ¹ 𝐍 ³" representation, so representation is more obvious.

Here are all the available units and their prefixes:

```julia
# SI base units
const m = Quantity(1.0, length=1)
const g = Quantity(1e-3, mass=1)
const s = Quantity(1.0, time=1)
const A = Quantity(1.0, current=1)
const K = Quantity(1.0, temperature=1)
const cd = Quantity(1.0, luminosity=1)
const mol = Quantity(1.0, amount=1)

@add_prefixes m (f, p, n, μ, u, c, m, k, M, G)
@add_prefixes g (μ, u, m, k)
@add_prefixes s (f, p, n, μ, u, m)
@add_prefixes A (n, μ, u, m, k)
@add_prefixes K (m,)
@add_prefixes cd (m,)
@add_prefixes mol (m,)

# SI derived units
const Hz = inv(s)
const N = kg * m / s^2
const Pa = N / m^2
const J = N * m
const W = J / s
const C = A * s
const V = W / A
const F = C / V
const Ω = V / A
const T = N / (A * m)

@add_prefixes Hz (k, M, G)
@add_prefixes N ()
@add_prefixes P ()
@add_prefixes J ()
@add_prefixes W (k, M, G)
@add_prefixes C ()
@add_prefixes V (m, k, M, G)
@add_prefixes F ()
@add_prefixes Ω (m,)
@add_prefixes T ()

# Assorted units
const min = 60 * s
const h = 60 * min
const day = 24 * h
const yr = 365.25 * day
```

I tried to add all of the common units. For physical constants or other weird unit systems, I think it is best we make the user define them themselves, to ensure they know what they are doing, and avoid potential namespace collisions.

You can access these units with `uparse` or `@u_str`, as in Unitful. For example:

```julia
julia> x = 0.5u"km/s"
500.0 m s⁻¹

julia> x = 100u"ug/day"
1.1574074074074076e-12 kg s⁻¹

julia> x = 100u"fm^2"
1.0000000000000001e-28 m²

julia> x = 0.5u"kW*h"
1.8e6 m² kg s⁻²
```

You can still convert from Unitful.jl if you need to, but I think this is a more robust option.

What do you think? @oscardssmith @j-fu @odow

---

TODO:

- [x] Throw an error if the user is converting from Unitful.jl with a non-SI `upreferred`.
- [x] Update README
- [x] Add docs on available units


src/types.jl
```diff
@@ -48,10 +48,11 @@ struct Dimensions{R <: Real}
     )
     Dimensions{_R}(; kws...) where {_R} = Dimensions(_R; kws...)
     Dimensions{_R}(args...) where {_R} = Dimensions(Base.Fix1(convert, _R).(args)...)
+    Dimensions{_R}(d::Dimensions) where {_R} = Dimensions{_R}(d.length, d.mass, d.time, d.current, d.temperature, d.luminosity, d.amount)
 end
 
 const DIMENSION_NAMES = Base.fieldnames(Dimensions)
-const DIMENSION_SYNONYMS = (:𝐋, :𝐌, :𝐓, :𝐈, :𝚯, :𝐉, :𝐍)
+const DIMENSION_SYNONYMS = (:m, :kg, :s, :A, :K, :cd, :mol)
```
odow, line 55: :+1:

src/units.jl
```diff
@@ -0,0 +1,106 @@
...
+    for (prefix, value) in zip(keys(all_prefixes), values(all_prefixes))
+        prefix in prefixes || continue
+        new_unit = Symbol(prefix, base_unit)
+        push!(expr.args, :(const $new_unit = $value * $base_unit))
+    end
+    return expr
+end
+
+# SI base units
+const m = Quantity(1.0, length=1)
```
oscardssmith, line 30: IMO  single leter names here are bad. I would go with `meter` `gram` etc.

test/unittests.jl
```diff
@@ -35,8 +35,8 @@ using Test
         @test ustrip(y) ≈ T(0.4)
 
         if R <: Rational
-            @test string(x) == "0.2 𝐋 ¹ 𝐌 ⁵ᐟ²"
-            @test string(inv(x)) == "5.0 𝐋 ⁻¹ 𝐌 ⁻⁵ᐟ²"
+            @test string(x) == "0.2 m kg⁵ᐟ²"
+            @test string(inv(x)) == "5.0 m⁻¹ kg⁻⁵ᐟ²"
```
odow, line 39: Not a physicist. And this is late in the piece for such a question, but in what situations are rational units used? What is `kg^{5/2}` ? So you can have (abusing notation) `y = sqrt(x)::m^{1/2}` and then `(y^2)::m`?

src/units.jl
MilesCranmer: This is just for use in `uparse`, so you can write, e.g., `10u"km/h"`. This file is wrapped in a `module Units ... end` so none of these units are actually exported.

src/units.jl
```diff
@@ -0,0 +1,106 @@
...
+@add_prefixes F ()
+@add_prefixes Ω (m,)
+@add_prefixes T ()
+
+# Assorted units
+const min = 60 * s
+const h = 60 * min
+const day = 24 * h
+const wk = 7 * day
+const yr = 365.25 * day
```
odow, line 74: This seems wrong. I don't know what other measurement systems do, but this is a pretty steep approximation.

test/unittests.jl
MilesCranmer: I can't think of any constants or units that are rational either. (when represented in SI base units). Having the dimensions be rational is more for their use in intermediate calculations.

(You could use `FixedRational{Int8,1}` as the parametric type of `Dimensions` if you are confident the dimensions will never need to be rational)

src/units.jl
MilesCranmer: This is just the SI definition of a year. The SI definition is different from a literal orbit around the sun (which has measurement uncertainty associated with it).

test/unittests.jl
MilesCranmer: (Oops, I mean `Int8`. No need for `FixedRational` if you aren't using rational numbers...)

MilesCranmer:
I'm debating whether to add c, the speed of light, as a derived unit. I'm hesitant because c feels more like a physical constant, and I wouldn't want to mislead people that physical constants are present: if c is included, people might think h (planck constant) is also included. But in actuality, h = hours.

I think I won't include it, but just wanted to hear if others have thoughts.

The other non-SI units I added are L (liters), bar, and eV - all of which are fairly common in scientific calculations. eV is technically a "constant" but due to its usage it feels more like a unit. (eV, years, and days are probably the only ones which can get away with being "units" despite technically being physical constants.)

src/units.jl
mcabbott: Should these be tied to Float64? Using a smaller number type would allow e.g. `2 * Quantity(true, length=1)` to come out as an integer.

MilesCranmer: > Should these be tied to Float64? Using a smaller number type would allow e.g. `2 * Quantity(true, length=1)` to come out as an integer.

This is fair, but (1) it would mean the units would have different types (e.g., `fm` would have a different type than `m`). And (2) I think the type of people who worry about 8-bit vs 64-bit types are likely to figure out they can use `convert(Quantity{Bool}, u"m/s")` when needed. Whereas the lay user will be left confused when they type out
```julia
julia> u"m/s"
true m s⁻¹

julia> u"km/s"
1000.0 m s⁻¹
```

MilesCranmer: Oh, btw, your example is true regardless:

```julia
julia> typeof(2 * Quantity(true, length=1))
Quantity{Int64, DynamicQuantities.FixedRational{Int32, 25200}}
```

the types of *units* (given by `uparse` and `@u_str`) are Float64. But a `Quantity()` is whatever type you pass.

mcabbott: Yea I know, sorry my point was roughly to wonder whether `2m` should give something that's an integer, not a Float64, and perhaps that `2f0 m` should be Float32. Maybe Bool is too exotic but `1` would do this.

MilesCranmer: I guess we could make some `WeakFloat64` that stores a `Float64` but eagerly converts to the user’s type upon being used in any calculation, so long as there’s no rounding error?

But might introduce some type instability.

mcabbott: Not sure I follow, how would that differ from `1::Int`? This is type-stable.

For numbers < 1, you could use Rational, or your FixedRational. Besides not promoting small types, this has the advantage of not reducing the accuracy of large ones. `(1e-3) * big(1.0)` is only Float64 accurate, while `(1//1000) * big(1.0) == big"0.001"`

MilesCranmer: e.g., femtometers (=1e-15 m)

MilesCranmer: Is this better @mcabbott?

```julia
    z = u"yr"
    @test typeof(z) == Quantity{Rational{Int64},DEFAULT_DIM_TYPE}

    z2 = 1.0u"yr"
    @test typeof(z2) == Quantity{Float64,DEFAULT_DIM_TYPE}

    z3 = 1f0u"yr"
    @test typeof(z3) == Quantity{Float32,DEFAULT_DIM_TYPE}
```

So now the default type of **all** units is Rational{Int}. Thus, promotion to floats works as you might expect.

MilesCranmer:
Hm, the one issue with this is you can get overflow errors for certain units:

```julia
julia> u"fm^2"
ERROR: LoadError: OverflowError: 1000000000000000 * 1000000000000000 overflowed for type Int64
```

(`fm^2` is actually used for scattering cross sections in particle physics)

@mcabbott

MilesCranmer:
Reverted units back to Float64.

src/units.jl
MilesCranmer: Discussion continued in main thread

MilesCranmer:
@mcabbott Maybe something like this could be an option for the type of each unit? It's a Float64 that tries to **demote** itself whenever possible:

```julia
struct WeakFloat <: AbstractFloat
    x::Float64
end

Base.promote(x::WeakFloat, y::Float64) = (x.x, y)
Base.promote(x::WeakFloat, y::AbstractFloat) = (convert(typeof(y), x.x), y)
Base.promote(x::WeakFloat, y) = promote(x.x, y)
Base.promote(x, y::WeakFloat) = reverse(promote(y, x))

function Base.promote(x::WeakFloat, y::Rational{R}) where {R}
    # If under eps * 1e5, we promote to a float:
    abs(x.x) < eps(Float64) * 1e5 && return promote(x.x, y)
    return (rationalize(R, x.x), y)
end

function Base.promote(x::WeakFloat, y::Integer)
    abs(x.x) < eps(Float64) * 1e5 && return promote(x.x, y)
    let r=rationalize(typeof(y), x.x)
        return isinteger(r) ? (convert(typeof(y), r), y) : promote(r, y)
    end
end
```

Which gives you:

```julia
julia> 2 * WeakFloat(100.0)
200

julia> (2//1) * WeakFloat(0.5)
1//1

julia> WeakFloat(1e-5) * 10000
1//10

julia> WeakFloat(1e-20) * 20
2.0e-19
```

The downside is that this is type unstable: the promoted type varies depending on how close the float is to 0.

src/fixed_rational.jl
```diff
@@ -43,6 +43,8 @@ Base.convert(::Type{Rational{R}}, x::F) where {R,F<:FixedRational} = Rational{R}
 Base.convert(::Type{Rational}, x::F) where {F<:FixedRational} = Rational{eltype(F)}(x.num, denom(F))
 Base.convert(::Type{AF}, x::F) where {AF<:AbstractFloat,F<:FixedRational} = convert(AF, x.num) / convert(AF, denom(F))
 Base.round(::Type{T}, x::F) where {T,F<:FixedRational} = div(convert(T, x.num), convert(T, denom(F)), RoundNearest)
+Base.promote(x::Integer, y::F) where {F<:FixedRational} = (F(x), y)
+Base.promote(x::F, y::Integer) where {F<:FixedRational} = (x, F(y))
 Base.promote(x, y::F) where {F<:FixedRational} = promote(x, convert(Rational, y))
 Base.promote(x::F, y) where {F<:FixedRational} = promote(convert(Rational, x), y)
```
mcabbott, line 49: Aside, but I think the recommended way to do this is to overload promote_rule. There is some logic to try combinations for you.
```
help?> Base.promote_rule
  promote_rule(type1, type2)

  Specifies what type should be used by promote when given values of types type1 and type2. This
  function should not be called directly, but should have definitions added to it for new types as
  appropriate.

julia> methods(Base.promote)
# 6 methods for generic function "promote" from Base:
 [1] promote()
     @ promotion.jl:388
 [2] promote(x::T, y::T, zs::T...) where T
     @ promotion.jl:409
 [3] promote(x, y, z)
     @ promotion.jl:397
 [4] promote(x, y, z, a...)
     @ promotion.jl:403
 [5] promote(x, y)
     @ promotion.jl:391
 [6] promote(x)
     @ promotion.jl:389

julia> methods(Base.promote_rule)
# 191 methods for generic function "promote_rule" from Base:
...
   [9] promote_rule(::Type{Float64}, ::Type{Bool})
     @ float.jl:166
  [10] promote_rule(::Type{Float64}, ::Type{UInt32})
     @ float.jl:166
  [11] promote_rule(::Type{Float64}, ::Type{Int16})
     @ float.jl:160
  [12] promote_rule(::Type{Float64}, ::Type{Int8})
     @ float.jl:160
  [13] promote_rule(::Type{Float64}, ::Type{UInt16})
     @ float.jl:166
  [14] promote_rule(::Type{Float64}, ::Type{UInt8})
     @ float.jl:166
  [15] promote_rule(::Type{Float64}, ::Type{Int32})
     @ float.jl:160
  [16] promote_rule(::Type{Float64}, ::Type{UInt128})
     @ float.jl:173
  [17] promote_rule(::Type{Float64}, ::Type{Float16})
     @ float.jl:400
  [18] promote_rule(::Type{Float64}, ::Type{UInt64})
     @ float.jl:166
  [19] promote_rule(::Type{Float64}, ::Type{Int128})
     @ float.jl:174
  [20] promote_rule(::Type{Float64}, ::Type{Float32})
     @ float.jl:401
...
```

MilesCranmer: Thanks, I didn’t realize this. I guess the only advantage of overloading `promote` instead is I could do things like calling `rationalize` rather than just `convert(Rational` (which does a fairly bad job — approximating floats for whatever reason).

oscardssmith: the problem is it does a very good job converting floats to rationals exactly.

oscardssmith: see https://github.com/JuliaLang/julia/pull/43427 which will make it so `rationalize` works for numbers that aren't abstractfloats.

MilesCranmer: It's weird that that is the default behavior:

```julia
julia> convert(Rational, 0.2)
3602879701896397//18014398509481984

julia> rationalize(0.2)
1//5
```

but both fractions evaluate to 0.2:

```julia
julia> 3602879701896397/18014398509481984
0.2

julia> 1 / 5
0.2
```

MilesCranmer: Okay I think I will just stick with `Float64` for now, and be explicit about it in the docstring.

MilesCranmer:
Thanks for the reviews, everyone!

PR #23, opened 2023-06-12
Additional utilities for identity functions, `+`, and `-`

MilesCranmer:
This fixes the behavior of `one`, `zero`, `oneunit` as pointed out by @mcabbott in #6.

This also extends `+` and `-` for `::Quantity` to `::Number`, given the quantity is dimensionless.

Could you take a look @mcabbott?


src/utils.jl
```diff
@@ -45,16 +45,26 @@ Base.iterate(::Dimensions, ::Nothing) = nothing
...
-Base.one(::Type{Dimensions{R}}) where {R} = Dimensions{R}()
-
-Base.zero(::Type{Quantity{T}}) where {T} = zero(Quantity{T,DEFAULT_DIM_TYPE})
 Base.one(::Type{Quantity{T}}) where {T} = one(Quantity{T,DEFAULT_DIM_TYPE})
-
-Base.zero(::Type{Quantity}) = zero(Quantity{DEFAULT_VALUE_TYPE})
 Base.one(::Type{Quantity}) = one(Quantity{DEFAULT_VALUE_TYPE})
+Base.one(::Type{Dimensions{R}}) where {R} = Dimensions{R}()
 Base.one(::Type{Dimensions}) = one(Dimensions{DEFAULT_DIM_TYPE})
+Base.one(q::Quantity) = one(typeof(q))
```
MilesCranmer, line 54: To be more explicit, should this throw an error pointing the user to `oneunit(::Quantity)` or `one(::Type{Quantity})` instead?

MilesCranmer: (Though it is true that `q * one(q)` is perfectly valid...)

j-fu:
Essentially, this PR seems to hit in the same direction as #18, but just to remark: this does not define an universal `zero(DynamicQuantity(T))`, and without this, DynamicQuantity is unable to act like a "normal" number.



BTW I found an intersting paper here: 

https://link.springer.com/article/10.1023/B:JOMC.0000034933.47311.93

And two other papers citing it  concerned with programmable formalization: 
https://scholar.google.de/scholar?cites=5148028135146171185&as_sdt=2005&sciodt=0,5&hl=de

Up to what I could understand, these don't work with "universal zero" as well.

MilesCranmer:
Thanks, I see. Do you want to take a look at this too?

I don't know what you mean by `zero(DynamicQuantity(T))`, could you elaborate? The `zero` definition is given here:

https://github.com/SymbolicML/DynamicQuantities.jl/blob/5779e63ce823fcdba0295f6b966ddd875c5d6994/src/utils.jl#L57-L61

As is stated, it can only be defined for `::Quantity`. Other definitions lack the dimensional information (when given as a type), or are defined on a `Dimension` themselves, and are therefore ill-posed.

(Only in Unitful.jl is the dimension information accessible from the type itself; whereas here it is stored in the value - and therefore constrains the `zero()` definition to be like this.)

MilesCranmer:
Btw, note that with this PR, you can now add numbers to dimensionless quantities:

```julia
julia> x = Quantity(0.5)
0.5 

julia> x + 1.0
1.5 
```

or multiply scalars with vectors:

```julia
julia> x = Quantity(ones(32))
[1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0  …  1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0] 

julia> x * 0.5
[0.5, 0.5, 0.5, 0.5, 0.5, 0.5, 0.5, 0.5, 0.5, 0.5  …  0.5, 0.5, 0.5, 0.5, 0.5, 0.5, 0.5, 0.5, 0.5, 0.5] 
```

src/utils.jl
mcabbott: This seems OK, although I don't love trivial `Quantity(1.0)`. It seems to me that if the units are known statically not to apply to something, it would be better not to have the type which allows for them. (It m/m cancels off, that's dynamic of course.)

What this routing via the type won't allow is `one(Quantity([1 2; 3 4], time=1))`.

src/math.jl
```diff
@@ -2,22 +2,28 @@ Base.:*(l::Dimensions, r::Dimensions) = @map_dimensions(+, l, r)
...
 Base.:*(l::Quantity, r::Dimensions) = Quantity(l.value, l.dimensions * r)
 Base.:*(l::Dimensions, r::Quantity) = Quantity(r.value, l * r.dimensions)
-Base.:*(l::Quantity, r::Number) = Quantity(l.value * r, l.dimensions)
-Base.:*(l::Number, r::Quantity) = Quantity(l * r.value, r.dimensions)
-Base.:*(l::Dimensions, r::Number) = Quantity(r, l)
-Base.:*(l::Number, r::Dimensions) = Quantity(l, r)
+Base.:*(l::Quantity, r) = Quantity(l.value * r, l.dimensions)
+Base.:*(l, r::Quantity) = Quantity(l * r.value, r.dimensions)
+Base.:*(l::Dimensions, r) = Quantity(r, l)
+Base.:*(l, r::Dimensions) = Quantity(l, r)
```
mcabbott, line 8: It seems that both Dimensions and Quantity are given many number-like methods. Might it be simpler to only do this for Quantity, and use `Quantity(true, Dimension(...))` in place of a naked Dimension?

Or some special One if `true` is too narrow. But I'm not so sure how much it's intended to support `Quantity([1,2,3], time=1)`. It would seem simplest to me to restrict Quantity to real numbers, but maybe you have other plans.

Actually, when would methods like `*(::Number, ::Dimension)` be used at all? #22 defines `const m = Quantity(1.0, length=1)`. It could instead use this method and define `m::Dimension`, but instead seems to focus on numbers.

src/utils.jl
```diff
@@ -45,16 +45,26 @@ Base.iterate(::Dimensions, ::Nothing) = nothing
...
-Base.zero(::Type{Quantity{T,R}}) where {T,R} = Quantity(zero(T), R)
+# Multiplicative identities:
 Base.one(::Type{Quantity{T,R}}) where {T,R} = Quantity(one(T), R)
-Base.one(::Type{Dimensions{R}}) where {R} = Dimensions{R}()
-
-Base.zero(::Type{Quantity{T}}) where {T} = zero(Quantity{T,DEFAULT_DIM_TYPE})
 Base.one(::Type{Quantity{T}}) where {T} = one(Quantity{T,DEFAULT_DIM_TYPE})
-
-Base.zero(::Type{Quantity}) = zero(Quantity{DEFAULT_VALUE_TYPE})
 Base.one(::Type{Quantity}) = one(Quantity{DEFAULT_VALUE_TYPE})
```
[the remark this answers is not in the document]
MilesCranmer: It does preserve what arrived. It only passes `DEFAULT_VALUE_TYPE` when the user passes a naked `one(Quantity)`.

```julia
julia> typeof(one(Quantity{Float32, Rational{Int16}}))
Quantity{Float32, Rational{Int16}}

julia> typeof(one(Quantity{Float32}))
Quantity{Float32, DynamicQuantities.FixedRational{Int32, 25200}}

julia> typeof(one(Quantity))
Quantity{Float64, DynamicQuantities.FixedRational{Int32, 25200}}
```

src/utils.jl
MilesCranmer: Oh, good point. Let me try to fix that.

j-fu:
Ok, now found a way to make an MWE for what I am asking for:
Basically, this is about linear algebra with units.
This one has no mixed units yet, but highlights some interesting points.
```
julia> using DynamicQuantities, LinearAlgebra
julia> A=rand(10,10).*Quantity(1,length=1);
julia> b=rand(10).*Quantity(1,time=1);
```
In order to make `A\b` work , we need: 
```
julia> Base.oneunit(::Type{Quantity{T,R}}) where {T,R} = Quantity(one(T))
julia> Base.zero(::Type{Quantity{T,R}}) where {T,R} = Quantity(zero(T))
julia> Base.adjoint(a::Quantity{T,R}) where {T,R} = Quantity(adjoint(a.value),a.dimensions)
```

If not for `RowNonZero` pivoting (>=1.9) , we also would need `Base.abs` stripping the dimesions, or `Base.isless`. `RowNonZero` was made for having  linear algebra working with finite fields.
```
julia> LinearAlgebra.lupivottype(::Type{Quantity{T,R}}) where {T,R} = LinearAlgebra.RowNonZero()
```

So with these settings we get 
```
julia> A\b
10-element Vector{Quantity{Float64, DynamicQuantities.FixedRational{Int32, 25200}}}:
 -0.7353181472910584 m⁻¹ s
 3.697912215371167 m⁻¹ s
 4.3285995603007335 m⁻¹ s
 5.1254898127729 m⁻¹ s
 -0.7304999810589047 m⁻¹ s
 -6.27314096327331 m⁻¹ s
 0.2909458298595246 m⁻¹ s
 -2.954423433552737 m⁻¹ s
 -2.833125030348257 m⁻¹ s
 -2.6987760225010198 m⁻¹ s
```

If I find time I als will try to make an example with mixed quantities in the matrix and the RHS.

Besides of defining these methods in this package, with AbstractQuantities, it would be possible to define a `DirtyQuantity` (possibly in a different package) which would work the way I described.

MilesCranmer:
> In order to make `A\b` work , we need:
> 
> ```
> julia> Base.oneunit(::Type{Quantity{T,R}}) where {T,R} = Quantity(one(T))
> julia> Base.zero(::Type{Quantity{T,R}}) where {T,R} = Quantity(zero(T))
> julia> Base.adjoint(a::Quantity{T,R}) where {T,R} = Quantity(adjoint(a.value),a.dimensions)
> ```

`Base.adjoint` seems like something we can definitely implement.

I'm not sure we should add `Base.zero(::Type{Quantity{T,R}})` or `Base.oneunit(::Type{Quantity{T,R}})`. See the docstring for `zero`:

```julia
help?> zero
search: zero zeros iszero set_zero_subnormals get_zero_subnormals count_zeros RoundToZero leading_zeros trailing_zeros RoundFromZero finalizer

  zero(x)
  zero(::Type)


  Get the additive identity element for the type of x (x can also specify the type itself).

  See also iszero, one, oneunit, oftype.
```

Meaning that `zero(x)` needs to be able to be added to `x`. But `zero(::Type{Quantity})` violates this as it can produce dimension errors. So I think only `zero(::Quantity)` makes sense to implement here.

Here is the docstring for `oneunit`:

```julia
help?> oneunit
search: oneunit DimensionlessQuantity

  oneunit(x::T)
  oneunit(T::Type)


  Return T(one(x)), where T is either the type of the argument or (if a type is passed) the argument. This differs from one for dimensionful quantities:
  one is dimensionless (a multiplicative identity) while oneunit is dimensionful (of the same type as x, or of type T).
```

so it needs to know about dimensions - hence it requires `::Quantity` as input (since dimensions are stored in value), rather than `::Type{Quantity}`.

> we also would need `Base.abs` stripping the dimesions

We should be very hesitant to use a different algebra than standard unit packages unless there is a very good reason to do so. Both `astropy` and `Unitful` propagate units through `abs`, so `DynamicQuantities` should probably do so as well:

Astropy:

```python
[ins] In [2]: abs(1.5*u.km/u.s)
Out[2]: <Quantity 1.5 km / s>
```

Unitful:

```julia
julia> abs(1.5u"km/s")
1.5 km s⁻¹
```
> `Base.isless`

We can definitely add this. And I guess should raise `DimensionError` when the dimensions are different.

> Besides of defining these methods in this package, with AbstractQuantities, it would be possible to define a `DirtyQuantity` (possibly in a different package) which would work the way I described.

See #24 for a first attempt at implementing `AbstractQuantity`, which might make this possible.

mcabbott:
> In order to make `A\b` work  [...] `Vector{Quantity{Float64,`

This is one path, arrays of Quantity. In Unitful.jl, arrays of uniform unit are in principle free, and can be re-interpreted, like https://github.com/PainterQubits/Unitful.jl/pull/437 . But here, they involve storing many Dimensions objects, and can't be re-interpreted to feed to BLAS. 

The other possible path is `Quantity{Vector{Float64}}` etc. To wrap a dense array, and define linear algebra by un-wrapping, calling the usual (possibly BLAS) routines, and re-wrapping the result.

j-fu:
Well, arrays of quantity are certainly interesting for heterogeneous physical systems.  I am interested in "multiphysics" PDE systems in this context. 

With the switch to `AbstractQuantity` it is quite straightforward to define  a type with working semantics for linear algebra.

With 
```
Base.zero(::Type{XQuantity{T,R}}) where {T,R} = XQuantity(zero(T))
Base.zero(::XQuantity{T,R}) where {T,R} = XQuantity(zero(T))
Base.one(::Type{XQuantity{T,R}}) where {T,R} = XQuantity(one(T))
Base.one(::XQuantity{T,R}) where {T,R} = XQuantity(one(T))
adjoint(q::XQuantity{T,R}) where {T,R} = 

function Base.:+(l::XQuantity, r::XQuantity)
    if iszero(l.value)
        r
    elseif iszero(r.value)
        l
    else
        dimension(l) == dimension(r) ? new_quantity(typeof(l), ustrip(l) + ustrip(r), dimension(l)) : throw(DimensionError(l, r))
    end
end

XQuantity(adjoint(q.value),q.dimensions)
Base.isless(q,r) = q.dimensions==r.dimensions && q.value<r.value
```
this seems to get me through dense `A\b` and through `sparspaklu(sparse(A))\b` .

Now `adjoint` and `isless` seem to be undisputed. I agree that both `zero` and `one`  are mathematially inconsistent, and that `+` is even worse. OTOH having pluggable arithmetics of this kind is really intriguing to me.

src/math.jl
MilesCranmer: (Sorry, just realized all my messages were listed as pending because they are submitted as a batch when I hit "Review changes"...)

MilesCranmer:
Going to merge this now so we can have the fixed `one`/`zero`/`oneunit` included in 0.4.0, but we can refactor these in the future if needed.

PR #24, opened 2023-06-12
Make `AbstractQuantity` and `AbstractDimensions`

MilesCranmer:
This PR adds abstract super types for quantities and dimensions. Almost every function is now defined for the abstract version, even with the constructors set up automagically

```julia
julia> struct MyDimensions{R} <: AbstractDimensions{R}
           length::R
           mass::R
           time::R
           scale::R
       end

julia> x = Quantity(1.2, MyDimensions, time=5//2, scale=3)
1.2 s⁵ᐟ² scale³

julia> x^2
1.44 s⁵ scale⁶

julia> typeof(x)
Quantity{Float64, MyDimensions{DynamicQuantities.FixedRational{Int32, 25200}}}
```

and with an `AbstractQuantity` (which we can pass our new dimensions type too)

```julia
julia> struct MyQuantity{T,D<:AbstractDimensions} <: AbstractQuantity{T,D}
           value::T
           dimensions::D
       end

julia> x = MyQuantity(1.2, MyDimensions, time=5//2, scale=3)
1.2 s⁵ᐟ² scale³

julia> x^2
1.44 s⁵ scale⁶

julia> typeof(x)
MyQuantity{Float64, MyDimensions{DynamicQuantities.FixedRational{Int32, 25200}}}
```

The individual dimensions are by default the `FixedRational` type, but you can override this with, e.g

```julia
julia> x = MyQuantity(1.2, MyDimensions{Rational{Int}}, time=5//2, scale=3)
1.2 s⁵ᐟ² scale³
```

or by conversion:

```julia
julia> x = convert(MyQuantity{Float64,MyDimensions{Rational{Int}}}, MyQuantity(1.2, MyDimensions, time=5//2, scale=3))
1.2 s⁵ᐟ² scale³

julia> typeof(x)
MyQuantity{Float64, MyDimensions{Rational{Int64}}}
```

---

Old version:

> This is a draft PR to look at adding some abstract super types for quantities and dimensions.
> 
> Nearly all functions are defined for `::AbstractQuantity` and `::AbstractDimensions` now, so you can overload anything to a particular type.
> 
> However, the current interface is a bit awkward it relies on the user defining a `quantity` function to try to create a `Quantity` of the correct type:
> ```julia
> new_quantity(::Type{<:MyQuantity}, val, dim) = MyQuantity(val, dim)
> new_quantity(::Type{<:MyDimensions}, val, dim) = MyQuantity(val, dim)
> new_dimensions(::Type{<:MyDimensions}, args...) = MyDimensions(args...)
> ```
> 
> Maybe this is fine though.
> 
> This is because otherwise it would not know what `Quantity` type to create if you construct a quantity using, e.g., `0.5 * Dimensions(length=1)`. So you have to create a new `AbstractDimensions` every time you create an `AbstractQuantity`.
> 
> ~~The other issue is it relies on the fixed `DIMENSION_NAMES` constant for defining some macros. If a user wants to define a custom dimensions struct with different fields, this would not work. So there should be some way to check this.~~
> 
> Fixed: now you need to override `dimension_name(::AbstractDimensions, k::Symbol)`, and that's it.
> 
> So in total, you only need three functions for a custom type to work:
> 
> ```julia
> new_quantity
> new_dimensions
> dimension_name
> ```
> 
> as well as optionally `convert`, `one`, `zero`, and `oneunit`. The rest should automatically carry over.


mcabbott:
> wanting to use fewer fields in Dimensions

Yes. I missed that this PR did that too, was only thinking about `AbstractQuantity` above.

The existing `Dimensions` is SI + rational, plausibly you could decide you can get away with three `Int8` powers for some problem (a subset) or might want something more exotic (like c=1 units, or some imperial or financial units).

MilesCranmer:
> The existing `Dimensions` is SI + rational, plausibly you could decide you can get away with three `Int8` powers for some problem (a subset) or might want something more exotic (like c=1 units, or some imperial or financial units).

Just to clarify - the existing `Dimensions` already permits you to use `Int8`  rather than a rational (or any other type for the power).

But yeah I agree it could be useful to have other types of `Dimensions` objects. The question is how to best design the super type to allow both user-defined `Dimensions` as well as `AbstractArray`-based `Quantities`...

src/types.jl
```diff
@@ -18,42 +27,23 @@ example, the dimensions of velocity are `Dimensions(length=1, time=-1)`.
...
 
-const DIMENSION_NAMES = Base.fieldnames(Dimensions)
-const DIMENSION_SYNONYMS = (:m, :kg, :s, :A, :K, :cd, :mol)
-const SYNONYM_MAPPING = NamedTuple(DIMENSION_NAMES .=> DIMENSION_SYNONYMS)
+(::Type{D})(::Type{R}; kws...) where {R,D<:AbstractDimensions} = D{R}((tryrationalize(R, get(kws, k, zero(R))) for k in fieldnames(D))...)
+(::Type{D})(; kws...) where {D<:AbstractDimensions} = D(DEFAULT_DIM_TYPE; kws...)
+
+(::Type{D})(args...) where {R,D<:AbstractDimensions{R}} = constructor_of(D)(Base.Fix1(convert, R).(args)...)
+(::Type{D})(; kws...) where {R,D<:AbstractDimensions{R}} = constructor_of(D)(R; kws...)
+(::Type{D})(d::AbstractDimensions) where {R,D<:AbstractDimensions{R}} = D((getfield(d, k) for k in fieldnames(D))...)
```
MilesCranmer, line 45: Is building generic constructors like this a bad practice? It means the user doesn't have to write the constructors themselves (but could always overload them if needed).

src/types.jl
```diff
@@ -1,6 +1,15 @@
...
 const DEFAULT_VALUE_TYPE = Float64
 
+abstract type AbstractQuantity{T,R} end
+abstract type AbstractDimensions{R} end
+
+constructor_of(::Type{D}) where {D<:AbstractDimensions} = D
+constructor_of(::Type{D}) where {R,D<:AbstractDimensions{R}} = D.name.wrapper
+constructor_of(::Type{Q}) where {Q<:AbstractQuantity} = Q
+constructor_of(::Type{Q}) where {T,Q<:AbstractQuantity{T}} = Q.body.name.wrapper
+constructor_of(::Type{Q}) where {T,R,Q<:AbstractQuantity{T,R}} = Q.name.wrapper
```
MilesCranmer, line 11: The tricky thing about these is they break if the user adds another type parameter. Doesn't seem like there's a general way to go from, e.g.,
```julia
AbstractArray
AbstractArray{T}
AbstractArray{T,2}
```

to `AbstractArray`, in a way that accounts for any number of type parameters.

src/types.jl
MilesCranmer: Seems like these constructors are much slower. I guess the compiler isn't able to inline some of the stuff here. (Maybe `fieldnames` is not inferrable. i.e., a `@generated function` is needed?)

j-fu:
> What subtypes are envisaged?
I already posted this in one of the other discussions:
I would like to enable linear system solution with quantities.
This would need a couple of tweaks which are (probably rightfully) rejected for `Quantity`. By introduceing e.g. `XQuantity` we could enable dense and sparse linear system solution via:
```
# needed for dense A\b
Base.zero(::Type{XQuantity{T,R}}) where {T,R} = XQuantity(zero(T))
Base.one(::Type{XQuantity{T,R}}) where {T,R} = XQuantity(one(T))
adjoint(q::XQuantity{T,R}) where {T,R} = XQuantity(adjoint(q.value),q.dimensions)

# needed for sparspaklu(A)\b
Base.zero(::XQuantity{T,R}) where {T,R} = XQuantity(zero(T))
Base.one(::XQuantity{T,R}) where {T,R} = XQuantity(one(T))
Base.isless(q,r) = q.dimensions==r.dimensions && q.value<r.value
function Base.:+(l::XQuantity, r::XQuantity)
    if iszero(l.value)
        r
    elseif iszero(r.value)
        l
    else
        dimension(l) == dimension(r) ? new_quantity(typeof(l), ustrip(l) + ustrip(r), dimension(l)) : throw(DimensionError(l, r))
    end
end
```

MilesCranmer:
I think this sounds feasible to support. Does the current abstract type interface work for implementing this? Or is there something else that is needed to get it working?

MilesCranmer:
I changed the functions a user needs to overload, to make things simpler. Now it's just these ones:

```julia
dimension_name(::Dimensions, k::Symbol) = (length="m", mass="kg", time="s", current="A", temperature="K", luminosity="cd", amount="mol")[k]
quantity_constructor(::Type{<:Union{Quantity,Dimensions}}) = Quantity
dimension_constructor(::Type{<:Union{Quantity,Dimensions}}) = Dimensions
```

With the `AbstractQuantity` type having `.value::T` and `.dimensions::AbstractDimensions{R}`

j-fu:
> Does the current abstract type interface work for implementing this? Or is there something else that is needed to get it working?

Yes - I checked this. Not yet sure where I will land that code, though.

MilesCranmer:
Okay, let me know if there’s something missing in the abstract interface.

Also I think there might be a way to make these required functions completely optional.

MilesCranmer:
Okay, now you can use the interface without even overloading any internal functions! Here's an example:

```julia
julia> struct MyDimensions{R} <: AbstractDimensions{R}
           length::R
           mass::R
           time::R
           scale::R
       end

julia> struct MyQuantity{T,R} <: AbstractQuantity{T,R}
           value::T
           dimensions::MyDimensions{R}
       end

julia> x = MyQuantity(1.2, time=5//2, scale=3)
1.2 s⁵ᐟ² scale³

julia> x^2
1.44 s⁵ scale⁶
```

This shows how to add your desired "scale" property @odow.

The only caveat is that if you want to use the operation

```julia
(*)(::AbstractDimensions, ::Number)
```

as a way of constructing a quantity, then you need to overload the quantity construction function

```julia
import DynamicQuantities: quantity_constructor

quantity_constructor(::Type{<:MyDimensions}) = MyQuantity
```
Because otherwise it doesn't know what `AbstractQuantity` to create.

Also, if you want to use different printouts (e.g., for `scale`), you need to also overload
```julia
dimension_name(::MyDimensions, field::Symbol)::String
```

MilesCranmer:
Argh, now I'm wondering if instead of creating `AbstractQuantity`, I should have done everything with SimpleTraits.jl. As it stands, we can't easily create a `QuantityArray`... because Julia doesn't have multiple inheritance, and there is no trait-based version of `AbstractArray`

SBuercklin:
From the thread where this was proposed on Discourse, there was discussion of using the `Dimensions` type to be statically inferable. I just want to point out that this wouldn't be the case, as `Val(1)` and `Val(0)` are distinct types to the compiler. 

If supporting this statically-inferable idea is part of the intention, it's not clear to me that a single type parameter in `AbstractDimensions` can capture both the dynamic and static use-cases. If permitting the use of type-inferable dimensions is part of the plan, wouldn't it make more sense to remove the type parameter on `AbstractDimensions` entirely?

MilesCranmer:
> From the thread where this was proposed on Discourse, there was discussion of using the `Dimensions{R}` type to be statically inferable. I just want to point out that this wouldn't be the case, as `Val(1)` and `Val(0)` are distinct types to the compiler.
> 
> If supporting this statically-inferable idea is part of the intention, it's not clear to me that a single type parameter in `AbstractDimensions{R}` can capture both the dynamic and static use-cases. If permitting the use of type-inferable dimensions is part of the plan, wouldn't it make more sense to remove the type parameter on `AbstractDimensions` entirely?

Ah, apologies for the confusion – that comment on discourse was just an idea, but is unrelated to the changes implemented by this PR. Indeed, statically-inferrable dimensions are not possible with the current types. This PR is for if you would like to use custom dimension spaces, but with otherwise the same types.

odow:
cc @trulsf, x-ref https://github.com/trulsf/UnitJuMP.jl/issues/18

MilesCranmer:
Do you have a specific branch I could follow along? I am interested in make the abstract interface compatible with your use case.

What would your ideal API be? e.g., in this example:

```julia
julia> using DynamicQuantities, JuMP, HiGHS

julia> m = Model(HiGHS.Optimizer);

julia> set_silent(m)

julia> @variable(m, x >= 0, u"m/s")
x [1.0 m s⁻¹]

julia> @variable(m, y >= 0, u"km/h")
y [0.2777777777777778 m s⁻¹]

julia> @variable(m, z, Bin, Dimensions())
z [1.0 ]

julia> max_speed = 4u"km/s"
4000.0 m s⁻¹

julia> @constraint(m, x + y  <= max_speed * z)
x + 0.2777777777777778 y - 4000 z <= 0 [m s⁻¹]

julia> obj = @objective(m, Max, x + 3y)
0.8333333333333334 y + x [m s⁻¹]

julia> optimize!(m)

julia> println("obj = $(value(obj))")
obj = 12000.0 m s⁻¹

julia> println("x = $(value(x))")
x = 0.0 m s⁻¹

julia> println("y = $(value(y))")
y = 4000.0 m s⁻¹

julia> println("z = $(value(z))")
z = 1.0
```

what would the preferred outputs be?

MilesCranmer:
Also - I tried defining an `IsQuantity{X}` trait with SimpleTraits.jl, so that one could inherit methods from other types instead of `AbstractQuantity`. I realized this is not going to work, because it requires running `@traitfn` on functions in Base (like `(+)` to define it for types matching `IsQuantity{X}`). Due to the way traits in Julia work, this invalidates _all generic methods on functions in Base_ by which is maybe not a great idea.

For example,

```julia
@traitfn Base.eltype(::Type{D}) where {D; IsQuantity{D}} = ...
```

would actually generate a function `Base.eltype(::Type{D}) where {D}` - thus overwriting the stdlib one.

So we have to stick to abstract types until Julia starts allowing traits or multiple inheritance.

MilesCranmer:
If there are no other issues with the way this abstract type system, I may try to merge things later this week.

I guess one question to consider would be: do we want to have

```julia
struct Quantity{T,R} <: AbstractQuantity{T,R}
    value::T
    dimensions::Dimensions{R}
end
```

which is the way it is implemented right now, or instead something like:

```julia
struct Quantity{T,D<:AbstractDimensions} <: AbstractQuantity{T,D}
    value::T
    dimensions::D
end
```

so you could use the same quantity object but with a custom dimension types.


Cheers,
Miles

MilesCranmer:
One quick update: I also just refactored all quantities types to be parametrically typed based on their entire dimension type, **rather than** the element type of the dimensions. I think this is a much more general option, and does not cost us anything because the *space* of dimensions is still fixed throughout any calculation.

So, for example,

```julia
struct Quantity{T,D<:AbstractDimensions} <: AbstractQuantity{T,D}
    value::T
    dimensions::D
end
```

whereas before it was the more rigid

```julia
struct Quantity{T,R} <: AbstractQuantity{T,R}
    value::T
    dimensions::Dimensions{R}
end
```

This means that you could use the same `Quantity` object as implemented here, but with a different `AbstractDimensions` types.

---

The downside is that you can no longer automatically create quantities by multiplying a dimension object, like:

```julia
0.5 * Dimensions(length=1)
```

But I think this is in opposition to the abstract interface anyways (how do you associate a quantity _from_ a type?), so it's better kept out.

MilesCranmer:
Thanks everyone for the feedback on this PR! Merging now.

PR #30, opened 2023-06-29
Create LazyFloat64 type for units

MilesCranmer:
This PR creates the `LazyFloat64` type for storing units. It makes units try to _demote_ their numeric type, rather than promote everything to `Float64`:

```julia
julia> typeof(1u"m/s")
Quantity{Int64, Dimensions{DynamicQuantities.FixedRational{Int32, 25200}}}

julia> typeof(1.0u"m/s")

julia> typeof(1f0u"m^2.5")
Quantity{Float32, Dimensions{DynamicQuantities.FixedRational{Int32, 25200}}}

julia> typeof(1u"fm")
ERROR: InexactError: Int64(1.0e-15)
```

@mcabbott could you review this?

### DIFF
--- a/src/units.jl
+++ b/src/units.jl
@@ -5,8 +5,9 @@ export uparse, @u_str
 import ..DEFAULT_DIM_TYPE
 import ..DEFAULT_VALUE_TYPE
 import ..Quantity
+import ..LazyFloat64
 
-@assert DEFAULT_VALUE_TYPE == Float64 "`units.jl` must be updated to support a different default value type."
+const DEFAULT_UNIT_TYPE = LazyFloat64
 
 macro add_prefixes(base_unit, prefixes)
     @assert prefixes.head == :tuple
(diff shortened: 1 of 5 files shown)


src/units.jl
```diff
@@ -126,12 +127,12 @@ Parse a string containing an expression of units and return the
 corresponding `Quantity` object with `Float64` value. For example,
 `uparse("m/s")` would be parsed to `Quantity(1.0, length=1, time=-1)`.
 """
-function uparse(s::AbstractString)
-    return as_quantity(eval(Meta.parse(s)))::Quantity{DEFAULT_VALUE_TYPE,DEFAULT_DIM_TYPE}
+function uparse(s::AbstractString)::Quantity{LazyFloat64,DEFAULT_DIM_TYPE}
```
MilesCranmer, line 130: This ensures that any unit is converted to a LazyFloat64 at the end. So even `u"1.0"` would become a LazyFloat64.

MilesCranmer:
Moved to #66

PR #27, opened 2023-06-15
Create physical constants submodule

MilesCranmer:
PR for #26.

1. This creates a `Constants` submodule. 
2. `Units` and `Constants` are now both exported by DynamicQuantities (unsure about this, but feels right)
3. `Constants` is made available within the `@u_str` macro for ease of access. (Some constants are treated like units)

For example:

```julia
julia> mass_energy_of_sun = 1.0 * u"Const.M_sun * Const.c^2"
1.7870936681139022e47 m² kg s⁻²

julia> q1 = 1.5u"C"; q2 = 1e-5u"C";

julia> F = Const.k_e * q1 * q2 / (1u"m")
134813.2768839176 m² kg s⁻²

julia> f = 500u"MHz"
5.0e8 s⁻¹

julia> (Constants.h * f) / Constants.eV
2.0678338484619293e-6
```

The following constants are included.

- `c`: Speed of light in a vacuum. Standard.
- `h`: Planck constant. Standard.
- `hbar`: Reduced Planck constant (h/2π). Standard.
- `e`: Elementary charge. Standard.
- `k_B`: Boltzmann constant. Standard.
- `N_A`: Avogadro constant. Standard.
- `eV`: Electron volt. Standard.
- `R`: Molar gas constant. Standard.
- `F`: Faraday constant. Standard.
- `sigma_sb`: Stefan-Boltzmann constant. Standard.
- `alpha`: Fine-structure constant. Measured.
- `u`: Atomic mass unit (1/12th the mass of Carbon-12). Measured.
- `G`: Newtonian constant of gravitation. Measured.
- `mu_0`: Vacuum magnetic permeability. Measured.
- `eps_0`: Vacuum electric permittivity. Measured.
- `m_e`: Electron mass. Measured.
- `m_p`: Proton mass. Measured.
- `m_n`: Neutron mass. Measured.
- `a_0`: Bohr radius. Measured.
- `k_e`: Coulomb constant (Note: SI units only!). Measured.
- `Ryd`: Rydberg frequency. Measured.

I included the most common astro ones too (It might be nice if someone added similar sets of very common constants from other fields.)

- `M_earth`: Earth mass. Measured.
- `M_sun`: Solar mass. Measured.
- `M_jup`: Jupiter mass. Measured.
- `R_earth`: Nominal Earth equatorial radius. Standard.
- `R_jup`: Nominal Jupiter equatorial radius. Standard.
- `R_sun`: Nominal solar radius. Standard.
[...]

Issue #26, opened 2023-06-15
Physical constants


MilesCranmer:
We could have a submodule that delivers common physical constants. Already I am finding myself having to type out the speed of light by hand; it would be useful if it were just built-in.

This could work as follows:

```julia
using DynamicQuantities
import DynamicQuantities: Units as U, Constants as C
```

Then, `C` would let you access constants, and `U` the units. This is similar to how `astropy` does it, to avoid mixing namespaces.

e.g., the energy of a 500nm photon

```julia
wavelength = 500*U.nm
energy = C.c*C.h/wavelength
```

This `Constants` namespace could also be exposed within the `@u_str` macro, so you could also write this as

```julia
energy = (1/500) * u"C.c*C.h/nm"
```

---

I'd be interested to know if @odow or @chrisrackauckas have any thoughts or ideas in this direction.

odow:
That seems very reasonable to me.

j-fu:
Hi, how does this square with the physical constants defined in Unitful 
or those in 
https://github.com/JuliaPhysics/PhysicalConstants.jl? 

Codes using these constants from different sources should be guaranteed to work together.  
I think it might make sense to check this in CI...

MilesCranmer:
Could you be more specific, I don't understand your question. PhysicalConstants.jl uses Unitful.jl unfortunately, so we cannot use it directly. The constants in PhysicalConstants.jl are taken from the same source as in #27 - from CODATA 2018. By CI do you mean to check the constants are equal?

MilesCranmer:
I understand, that sounds good to me! Thanks.

MilesCranmer:
Closed with #32

PR #32, opened 2023-07-09
Create (1) physical constants, and (2) symbolic dimensions object

MilesCranmer:
This PR is the continuation of #27, fixing #26. 

It also now includes a symbolic dimensions object, to help with the JuMP.jl requirements: https://github.com/trulsf/UnitJuMP.jl/issues/18. This means you can operate with units without needing to convert to base units, like:

```julia
julia> us"km/h"
1.0 km h⁻¹

julia> expand_units(us"km/h")  # equivalent of u"km/h"
0.2777777777777778 m s⁻¹
```

TODO:

- [x] Add back docs on units
- [x] Verify constants docstrings show up on documentation
- [x] Add docs page on symbolic dimensions
- [x] Enable constants with namespace collisions, equal to their expanded forms


src/units.jl
```diff
@@ -1,48 +1,58 @@
 module Units
 
-export uparse, @u_str
+import SparseArrays as SA
+import Tricks: static_fieldnames
```
MilesCranmer, line 4: I don't think these are needed

MilesCranmer:
After merging, can add some stuff to the README about this.

Issue #2, opened 2023-06-06
Tradeoff of 0 for no dimension vs `nothing`


MilesCranmer:
Would be good to evaluate performance tradeoff of this. Could even store a Bool for each dimension. It might be faster to avoid updating a 0 dimension entirely.

PR #33, opened 2023-07-09
Create `QuantityArray <: AbstractArray`

MilesCranmer:
This creates a type:
```julia
QuantityArray{
    T,
    N,
    D<:AbstractDimensions,
    Q<:AbstractQuantity{T,D},
    V<:AbstractVector{T,N}
} <: AbstractArray{Q,N}
```
that stores a single set of physical dimensions for the entire array. 

For example:

```julia
A = QuantityArray(randn(32), u"km/s”)
```

it implements a custom broadcasting interface so as to create a new `QuantityArray` with the correct output dimensions. This part is a bit involved as it needs to materialize the first element. Thanks to tictaccat on Discourse for help.

Everything is type stable.

For safety purposes, the compiler is in charge of figuring out when the dimension calculation can be skipped for each element or not. It should be able to if it inlines things correctly, which happens for simple broadcasting. For more complex broadcasting it might not be able to inline things.

@ChrisRackauckas do you think you or someone else in @SciML could review this to see if it fits your requirements?

### DIFF
--- a/ext/DynamicQuantitiesLinearAlgebraExt.jl
+++ b/ext/DynamicQuantitiesLinearAlgebraExt.jl
@@ -1,0 +1,9 @@
+module DynamicQuantitiesLinearAlgebraExt
+
+import LinearAlgebra: norm
+import DynamicQuantities: AbstractQuantity, ustrip, dimension, new_quantity
+
+norm(q::AbstractQuantity, p::Real=2) = new_quantity(typeof(q), norm(ustrip(q), p), dimension(q))
+norm(q::AbstractArray{<:AbstractQuantity}, p::Real=2) = new_quantity(eltype(q), norm(ustrip(q), p), dimension(q))
+
+end
--- a/src/arrays.jl
+++ b/src/arrays.jl
@@ -69,21 +69,45 @@ end
 @inline ustrip(A::QuantityArray) = A.value
 @inline dimension(A::QuantityArray) = A.dimensions
 
-array_type(::Type{A}) where {T,A<:QuantityArray{T}} = Array{T,1}
+array_type(::Type{A}) where {A<:QuantityArray} = Array{T,N} where {T,N}
+array_type(::Type{A}) where {T,A<:QuantityArray{T}} = Array{T,N} where {N}
 array_type(::Type{A}) where {T,N,A<:QuantityArray{T,N}} = Array{T,N}
+array_type(::Type{A}) where {T,N,D,A<:QuantityArray{T,N,D}} = Array{T,N}
+array_type(::Type{A}) where {T,N,D,Q,A<:QuantityArray{T,N,D,Q}} = Array{T,N}
 array_type(::Type{A}) where {T,N,D,Q,V,A<:QuantityArray{T,N,D,Q,V}} = V
+
 array_type(A) = array_type(typeof(A))
 
+quantity_type(::Type{A}) where {A<:QuantityArray} = DEFAULT_QUANTITY_TYPE
+quantity_type(::Type{A}) where {T,A<:QuantityArray{T}} = Quantity{T,DEFAULT_DIM_TYPE}
+quantity_type(::Type{A}) where {T,N,A<:QuantityArray{T,N}} = Quantity{T,DEFAULT_DIM_TYPE}
+quantity_type(::Type{A}) where {T,N,D,A<:QuantityArray{T,N,D}} = Quantity{T,D}
 quantity_type(::Type{A}) where {T,N,D,Q,A<:QuantityArray{T,N,D,Q}} = Q
+quantity_type(::Type{A}) where {T,N,D,Q,V,A<:QuantityArray{T,N,D,Q,V}} = Q
+
 quantity_type(A) = quantity_type(typeof(A))
 
 dim_type(::Type{A}) where {A<:QuantityArray} = DEFAULT_DIM_TYPE
+dim_type(::Type{A}) where {T,A<:QuantityArray{T}} = DEFAULT_DIM_TYPE
+dim_type(::Type{A}) where {T,N,A<:QuantityArray{T,N}} = DEFAULT_DIM_TYPE
 dim_type(::Type{A}) where {T,N,D,A<:QuantityArray{T,N,D}} = D
+dim_type(::Type{A}) where {T,N,D,Q,A<:QuantityArray{T,N,D,Q}} = D
+dim_type(::Type{A}) where {T,N,D,Q,V,A<:QuantityArray{T,N,D,Q,V}} = D
+
 dim_type(A) = dim_type(typeof(A))
 
+# TODO: Can we avoid this pattern?
 value_type(::Type{A}) where {A<:QuantityArray} = DEFAULT_VALUE_TYPE
 value_type(::Type{A}) where {T,A<:QuantityArray{T}} = T
+value_type(::Type{A}) where {T,N,A<:QuantityArray{T,N}} = T
+value_type(::Type{A}) where {T,N,D,A<:QuantityArray{T,N,D}} = T
+value_type(::Type{A}) where {T,N,D,Q,A<:QuantityArray{T,N,D,Q}} = T
+value_type(::Type{A}) where {T,N,D,Q,V,A<:QuantityArray{T,N,D,Q,V}} = T
+
+value_type(::Type{Q}) where {Q<:AbstractQuantity} = DEFAULT_VALUE_TYPE
 value_type(::Type{Q}) where {T,Q<:AbstractQuantity{T}} = T
+value_type(::Type{Q}) where {T,D,Q<:AbstractQuantity{T,D}} = T
+
 value_type(A) = value_type(typeof(A))
 
 # One field:
--- a/src/arrays.jl
+++ b/src/arrays.jl
@@ -40,7 +40,11 @@ end
 QuantityArray(v::AbstractArray; kws...) = QuantityArray(v, DEFAULT_DIM_TYPE(; kws...))
 QuantityArray(v::AbstractArray, d::AbstractDimensions) = QuantityArray(v, d, Quantity)
 QuantityArray(v::AbstractArray, q::AbstractQuantity) = QuantityArray(v .* ustrip(q), dimension(q), typeof(q))
-QuantityArray(v::QA) where {Q<:AbstractQuantity,QA<:AbstractArray{Q}} = allequal(dimension.(v)) ? QuantityArray(ustrip.(v), dimension(first(v)), Q) : throw(DimensionError(first(v), v))
+QuantityArray(v::QA) where {Q<:AbstractQuantity,QA<:AbstractArray{Q}} =
+    let
+        allequal(dimension.(v)) || throw(DimensionError(first(v), v))
+        QuantityArray(ustrip.(v), dimension(first(v)), Q)
+    end
 # TODO: Should this check that the dimensions are the same?
 
 function Base.promote_rule(::Type{QA1}, ::Type{QA2}) where {QA1<:QuantityArray,QA2<:QuantityArray}
@@ -95,8 +99,13 @@ function Base.getindex(A::QuantityArray, i...)
         return new_quantity(quantity_type(A), output_value, dimension(A))
     end
 end
-Base.setindex!(A::QuantityArray{T,N,D,Q}, v::Q, i...) where {T,N,D,Q<:AbstractQuantity} = dimension(A) == dimension(v) ? unsafe_setindex!(A, v, i...) : throw(DimensionError(A, v))
-Base.setindex!(A::QuantityArray{T,N,D,Q}, v::AbstractQuantity, i...) where {T,N,D,Q<:AbstractQuantity} = setindex!(A, convert(Q, v), i...)
+function Base.setindex!(A::QuantityArray{T,N,D,Q}, v::Q, i...) where {T,N,D,Q<:AbstractQuantity}
+    dimension(A) == dimension(v) || throw(DimensionError(A, v))
+    return unsafe_setindex!(A, v, i...)
+end
+function Base.setindex!(A::QuantityArray{T,N,D,Q}, v::AbstractQuantity, i...) where {T,N,D,Q<:AbstractQuantity}
+    return setindex!(A, convert(Q, v), i...)
+end
 
 unsafe_setindex!(A, v, i...) = setindex!(ustrip(A), ustrip(v), i...)
 
@@ -156,7 +165,11 @@ materialize_first(q::AbstractArray{Q}, ::Any) where {Q<:AbstractQuantity} = firs
 materialize_first(r::Base.RefValue) = materialize_first(r.x)
 materialize_first(x::Base.Broadcast.Extruded) = materialize_first(x.x)
 materialize_first(args::Tuple) = materialize_first(first(args), Base.tail(args))
-materialize_first(args::AbstractArray) = length(args) >= 1 ? materialize_first(args[begin], args[begin+1:end]) : error("Unexpected broadcast format. Please submit a bug report.")
+materialize_first(args::AbstractArray) =
+    let
+        length(args) >= 1 || error("Unexpected broadcast format. Please submit a bug report.")
+        materialize_first(args[begin], args[begin+1:end])
+    end
 materialize_first(::Tuple{}) = error("Unexpected broadcast format. Please submit a bug report.")
 materialize_first(::Any, rest) = materialize_first(rest)
 
--- a/src/arrays.jl
+++ b/src/arrays.jl
@@ -83,29 +83,18 @@ end
 @inline ustrip(A::QuantityArray) = A.value
 @inline dimension(A::QuantityArray) = A.dimensions
 
-array_type(::Type{QuantityArray}) = Array{DEFAULT_VALUE_TYPE,N} where {N}
-array_type(::Type{QuantityArray{T}}) where {T} = Array{T,N} where {N}
-array_type(::Type{QuantityArray{T,N}}) where {T,N} = Array{T,N}
-array_type(::Type{QuantityArray{T,N,D}}) where {T,N,D} = Array{T,N}
-array_type(::Type{QuantityArray{T,N,D,Q}}) where {T,N,D,Q} = Array{T,N}
 array_type(::Type{<:QuantityArray{T,N,D,Q,V}}) where {T,N,D,Q,V} = V
-array_type(A) = array_type(typeof(A))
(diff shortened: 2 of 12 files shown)


src/arrays.jl
```diff
@@ -0,0 +1,177 @@
...
+materialize_first(x) = x
+
+function _print_array_type(io::IO, ::Type{QA}) where {QA<:QuantityArray}
+    return print(io, "QuantityArray(::", array_type(QA), ", ::", quantity_type(QA), ")")
+end
+Base.showarg(io::IO, v::QuantityArray, _) = _print_array_type(io, typeof(v))
+Base.show(io::IO, ::MIME"text/plain", ::Type{QA}) where {QA<:QuantityArray} = _print_array_type(io, QA)
+
+# Other array operations:
+Base.copy(A::QuantityArray) = QuantityArray(copy(ustrip(A)), dimension(A), quantity_type(A))
```
MilesCranmer, line 170: Should copy dimensions too, as some dimensions (SymbolicDimensions) store their data with an array

src/utils.jl
```diff
@@ -28,56 +29,77 @@ end
...
 
-Base.isfinite(q::AbstractQuantity) = isfinite(ustrip(q))
 Base.keys(d::AbstractDimensions) = static_fieldnames(typeof(d))
-# TODO: Make this more generic.
-Base.iszero(d::AbstractDimensions) = all_dimensions(iszero, d)
-Base.iszero(q::AbstractQuantity) = iszero(ustrip(q))
 Base.getindex(d::AbstractDimensions, k::Symbol) = getfield(d, k)
+
+# Compatibility with `.*`
+Base.length(::AbstractQuantity) = 1
```
MilesCranmer, line 42: Should forward to the value

MilesCranmer:
It seems like the remaining performance issue was a type inference problem. The change:

```diff
+         return QuantityArray(output_array, dimension(first_output)::dim_type(ElType), ElType)
-         return QuantityArray(output_array, dimension(first_output), ElType)
```

seems to have fixed it.

MilesCranmer:
So it seems to be doing better than an array of quantities now. But not as good as a naked array.

@gaurav-arya I remember we discussed this at some point. Curious to hear if you have any suggestions.

```julia
julia> f(x) = @. (x^2 * 0.9 + 0.9 * x / 0.1 * x)
```

Naked array:

```julia
julia> @btime f(x) setup=(x = randn(100000));
  39.416 μs (4 allocations: 781.31 KiB)
```

Array of quantities:

```julia
julia> @btime f(x) setup=(x = randn(100000) .* u"km/s");
  328.583 μs (4 allocations: 3.81 MiB)
```

Quantity array:

```julia
julia> @btime f(x) setup=(x = QuantityArray(randn(100000), u"km/s"));
  134.000 μs (38 allocations: 783.48 KiB)
```

It's strange that there are so many allocations, even though it's about the same size as the naked array. But it is great that the compiler seems to be skipping some of the dimensionality checks!

src/math.jl
```diff
@@ -35,6 +35,21 @@ Base.:^(l::AbstractQuantity{T,D}, r::Number) where {T,R,D<:AbstractDimensions{R}
...
+for (p, ex) in [
+    (0, :(one(l))),
+    (1, :(l)),
+    (2, :(l * l)),
+    (3, :(l * l * l)),
+    (-1, :(inv(l))),
+    (-2, :((i=inv(l); i*i)))
+]
+    @eval @inline Base.literal_pow(::typeof(^), l::AbstractQuantity, ::Val{$p}) = $ex
+end
```
gaurav-arya, line 51:
```suggestion
@inline function Base.literal_pow(::typeof(^), l::AbstractQuantity, v::Val)
    new_quantity(typeof(l), Base.literal_pow(^, ustrip(l), v), Base.literal_pow(^, dimension(l), v))
end
```
This seems to do just as well in benchmarking 🙂

gaurav-arya (approved):
Indeed, the `QuantityArray` seems to offer a pretty clear performance benefit in your case. (FWIW, if we construct the function as `_f(x) = x^2 * 0.9 + 0.9 * x / 0.1 * x; f(x) = _f.(x)`, the extra allocations go away, although it's not faster.)

I did experiment with cutting out some of the fields of `Dimensions` for the zero units similar to #2, and it seemed to give a substantial performance improvement. But it's a tricky issue because we probably want a `DynamicQuantity` to be type stable even if hit with new units. Later I'd love to play around a bit more on some benchmarks and see if I can make any tweaks based on them to get things closer to zero-cost. (Speaking of which, maybe the simple broadcast benchmark you had could be included in the benchmarks?) But this looks good!

src/fixed_rational.jl
```diff
@@ -47,14 +55,28 @@ Base.convert(::Type{F}, x::Rational) where {F<:FixedRational} = F(x)
...
 Base.promote(x::F1, y::F2) where {F1<:FixedRational,F2<:FixedRational} = (@assert denom(F1) == denom(F2); (x, y))
-Base.promote(x, y::F) where {F<:FixedRational} = promote(x, convert(Rational, y))
-Base.promote(x::F, y) where {F<:FixedRational} = reverse(promote(y, x))
-Base.show(io::IO, x::F) where {F<:FixedRational} = show(io, convert(Rational, x))
+
+Base.string(x::FixedRational) =
+    let
+        isinteger(x) && return string(convert(eltype(x), x))
+        g = gcd(x.num, denom(x))
+        return string(div(x.num, g)) * "//" * string(div(denom(x), g))
```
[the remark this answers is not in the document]
MilesCranmer: Good point. But note that this printing is mostly for the exponential form, like
```julia
julia> u"m^0.5*s^2"
1.0 m¹ᐟ² s²
```
Also, if you copy and paste it exactly, it would again become a FixedRational due to the eager conversion when creating dimension objects:

```julia
julia> u"m^(1//2)*s^2"
1.0 m¹ᐟ² s²

julia> typeof(u"m^(1//2)*s^2")
```

Maybe for printing the number by itself we could annotate the numbers similar to FixedPointNumbers.jl. However I should also note that FixedRational is really only set up for this specific use-case, as it lacks many methods required for a general numeric type – so it will convert itself to a Rational if a user tries to do any manual calculations with the dimension values.

(Also, I realize I should have put things like this into a different PR, my mistake... the commits are too intertwined to do a cherry pick unfortunately, I did try)

src/fixed_rational.jl
```diff
@@ -47,14 +55,28 @@ Base.convert(::Type{F}, x::Rational) where {F<:FixedRational} = F(x)
 Base.convert(::Type{Rational{R}}, x::F) where {R,F<:FixedRational} = Rational{R}(x.num, denom(F))
 Base.convert(::Type{Rational}, x::F) where {F<:FixedRational} = Rational{eltype(F)}(x.num, denom(F))
 Base.convert(::Type{AF}, x::F) where {AF<:AbstractFloat,F<:FixedRational} = convert(AF, x.num) / convert(AF, denom(F))
-Base.convert(::Type{I}, x::F) where {I<:Integer,F<:FixedRational} = convert(I, convert(Rational, x))
+Base.convert(::Type{I}, x::F) where {I<:Integer,F<:FixedRational} = isinteger(x) ? div(x.num, denom(F)) : throw(InexactError(:convert, I, x))
 Base.round(::Type{T}, x::F) where {T,F<:FixedRational} = div(convert(T, x.num), convert(T, denom(F)), RoundNearest)
+Base.decompose(x::F) where {T,F<:FixedRational{T}} = (x.num, zero(T), denom(F))
+
+# Promotion rules:
+Base.promote_rule(::Type{<:FixedRational{T1,den1}}, ::Type{<:FixedRational{T2,den2}}) where {T1,T2,den1,den2} = FixedRational{promote_type(T1,T2),den*den2}
```
MilesCranmer, line 63: Actually this method should probably be deleted. We don't want users to ever work with two different denominators when using FixedRationals (type explosion!). If they need it they can just use Rational instead which is safer.

MilesCranmer: (Or turned into a helpful error)

src/fixed_rational.jl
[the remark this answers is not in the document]
MilesCranmer: I guess to summarize what I'm saying, I think we actually **want** the user to copy-paste that explicitly and end up with a `Rational` rather than a `FixedRational`.

We wouldn't want the user to work with `FixedRational` manually because it's not set up for general calculations.

src/math.jl
MilesCranmer: How are you benchmarking it? For some reason I get worse performance on the broadcasting benchmark:

1. With explicit expansion like in `intfuncs.jl`:

```julia
julia> f(x) = x^2 * 0.9 + 0.9 * x / 0.1 * x;

julia> @btime f.(x) setup=(x = QuantityArray(randn(100000), u"km/s"));
  97.250 μs (2 allocations: 781.30 KiB)
```

2. With the generic `@inline Base.literal_pow(::typeof(^), l::AbstractQuantity, ::Val{p}) where {p} = new_quantity(typeof(l), Base.literal_pow(^, ustrip(l), Val(p)), Base.literal_pow(^, dimension(l), Val(p)))`:

```julia
julia> @btime f.(x) setup=(x = QuantityArray(randn(100000), u"km/s"));
  1.130 ms (2 allocations: 781.30 KiB)
```

But I'll also note that if I change `f(x)` to use `x^4` (and similar for the right term), avoiding the `literal_pow`, then I also get similarly poor performance for 1.

I wonder why this is. Maybe the general power law method for quantities is too complex, and messes with the compiler inlining, thus forcing the broadcasted expression to do dimension checking for every number? (Whereas for `x * x * x` the compiler can easily inline the dimension calculations).

Maybe I need to add a power law method for `Core.Const()`??

MilesCranmer: Code in question:

```julia
_pow(l::AbstractDimensions{R}, r::R) where {R} = map_dimensions(Base.Fix1(*, r), l)
Base.:^(l::AbstractDimensions{R}, r::Number) where {R} = _pow(l, tryrationalize(R, r))
Base.:^(l::AbstractQuantity{T,D}, r::Integer) where {T,R,D<:AbstractDimensions{R}} = new_quantity(typeof(l), ustrip(l)^r, dimension(l)^r)
Base.:^(l::AbstractQuantity{T,D}, r::Number) where {T,R,D<:AbstractDimensions{R}} =
    let dim_pow = tryrationalize(R, r), val_pow = convert(T, dim_pow)
        # Need to ensure we take the numerical power by the rationalized quantity:
        return new_quantity(typeof(l), ustrip(l)^val_pow, dimension(l)^dim_pow)
    end
```

gaurav-arya: You're right, I don't know how that didn't show locally.

I also thought it was inlining, but I realized there's actually a semantics difference with how the dimensions were getting transformed. In the old code, the multiplication would end up being unrolled into repeated addition, but with the generic version it stays as multiplication.

Although it unfortunately doesn't result in any decrease in code complexity, the right way to handle this seems to be to keep the generic `literal_pow` method for quantities, but actually add a specialization for dimensions. I've implemented this in #36 (on my machine, this actually gets a speedup on your benchmark over even the original code!)

src/fixed_rational.jl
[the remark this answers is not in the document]
MilesCranmer: Sounds good!

src/math.jl
gaurav-arya: (That being said, afaiu we wouldn't usually expect repeated addition to be an improvement over multiplication... but maybe it plays better with inlining somehow)

MilesCranmer: In looking at the code trace for the higher order literal pow, it converts from `Core.Const` to `Int64` when it reaches the `^`. So I wonder if we should use the `map_domensions` call directly in the literal pow?

gaurav-arya: I just experimented with adding the following line to #36:
```julia
Base.literal_pow(::typeof(^), l::AbstractDimensions, ::Val{4}) = map_dimensions(Base.Fix1(*, 4), l)
```
but it doesn't seem to help much if at all. However,
```julia
Base.literal_pow(::typeof(^), l::AbstractDimensions, ::Val{4}) = map_dimensions(+
, l, l, l, l)
```
does help a lot. But I wasn't able to generically get the compiler to const-prop for higher powers, something to get to the bottom of...

For now I added a simple benchmark to #36

MilesCranmer:
Still no speedup on this test:

```julia
f(x) = x^4 * 0.9 - x * x / 0.3 * x * 0.9 * x
```

However there is definitely a speedup on `f(x) = x^4` relative to an array of quantities. So maybe we just leave it at that? I guess some broadcasted functions might be too complex for the compiler inlining to eliminate the dimension check.

MilesCranmer:
Wow, custom broadcasting interfaces are super complicated... So, broadcasting is able to correctly infer the output type for `(QuantityArray) .* (Float64)`, but is not able to infer it for `(QuantityArray) * (Quantity)`. We can see this with:

```julia
Base.broadcasted(*, QuantityArray(randn(32)), 1.0).args[end]
```

which is `1.0`. However,

```julia
Base.broadcasted(*, QuantityArray(randn(32)), u"m/s").args[end]
```

is `Any[1.0 m s⁻¹]`. This causes the broadcasting to infer the output type as `AbstractVector` rather than `QuantityArray`.

I wonder if we would need to define `Base.BroadcastStyle` for regular `AbstractQuantity` as well as `QuantityArray`?

ext/DynamicQuantitiesLinearAlgebraExt.jl
```diff
@@ -0,0 +1,9 @@
+module DynamicQuantitiesLinearAlgebraExt
+
+import LinearAlgebra: norm
+import DynamicQuantities: AbstractQuantity, ustrip, dimension, new_quantity
+
+norm(q::AbstractQuantity, p::Real=2) = new_quantity(typeof(q), norm(ustrip(q), p), dimension(q))
+norm(q::AbstractArray{<:AbstractQuantity}, p::Real=2) = new_quantity(eltype(q), norm(ustrip(q), p), dimension(q))
```
gaurav-arya, line 7: I read somewhere that dispatches such as this one, where there is a generic interface (`norm(q::AbstractArray)`) and one extends it in a case where one only owns some type parameters, but not the fully type, can be a bit iffy. I tried to find the discussion (I think this was said by @vtjnash) but it might have disappeared in the slack void:/ (I would love to find that discussion, as if true it would certainly be somewhat unwritten knowledge...) But I think practically speaking it can cause invalidations by pushing e.g. a number of eligible methods over `max_methods`. 

For this reason, since we don't / aren't able to make this performance optimization anywhere else, I would maybe nudge towards removing this method to avoid any invalidation headaches, unless `norm` is a particular bottleneck.

src/utils.jl
```diff
@@ -163,10 +190,10 @@ Base.copy(q::Q) where {Q<:AbstractQuantity} = new_quantity(Q, copy(ustrip(q)), c
 
 Remove the units from a quantity.
 """
-ustrip(q::AbstractQuantity) = q.value
+@inline ustrip(q::AbstractQuantity) = q.value
 ustrip(::AbstractDimensions) = error("Cannot remove units from an `AbstractDimensions` object.")
 ustrip(aq::AbstractArray{<:AbstractQuantity}) = ustrip.(aq)
```
gaurav-arya, line 195: This type parameter dispatch looks much more innocuous than the other one since you own `ustrip`. Although, even here you might be able to get away without the `<:AbstractQuantity` dispatch and just handle all `AbstractArray`s, so long as `ustrip` has a generic no-op fallback for reals?

gaurav-arya:
> So maybe we just leave it at that? I guess some broadcasted functions might be too complex for the compiler inlining to eliminate the dimension check.

Agree about leaving further performance hacking to the future:)

src/utils.jl
gaurav-arya: (I was commenting on line 195, and I realize that this wasn't really modified by the PR, so apologies for the slightly off topic note)

ext/DynamicQuantitiesLinearAlgebraExt.jl
MilesCranmer: Removal sounds good to me.

MilesCranmer: Actually I think `norm` has to stay for arrays. I just tried removing it but it immediately runs into an error due to the use of `float` in `LinearAlgebra/src/generic.jl`:

```julia
generic_normInf(x) = float(mapreduce(norm, max, x))
```

which is used at the beginning of `norm2`:

```julia
function generic_norm2(x)
    maxabs = normInf(x)
    ...
```

One alternative would be to redefine the behavior of `float(::AbstractQuantity)` to only apply `float` to the value of a quantity. Right now `float` attempts to convert a quantity into a scalar, and throws an error if there are any dimensions left. Not sure what the best strategy is.

MilesCranmer: Fixed in https://github.com/SymbolicML/DynamicQuantities.jl/pull/37

src/utils.jl
MilesCranmer: Fixed in #37

MilesCranmer:
Cool. So once #37 merges, I think all that's left is writing the unittests to get coverage back to 100%.

MilesCranmer:
Seems like a bunch of the array tests pass on 1.10 but don't on 1.9... Maybe something in the broadcasting interface changed.

MilesCranmer:
Woo! 100% testing coverage again!

I'll do another code review and then we can merge.

src/fixed_rational.jl
```diff
@@ -47,14 +60,39 @@ Base.convert(::Type{F}, x::Rational) where {F<:FixedRational} = F(x)
...
+function Base.promote_rule(::Type{<:FixedRational{T1}}, ::Type{Rational{T2}}) where {T1,T2}
+    return Rational{promote_type(T1,T2)}
+end
+function Base.promote_rule(::Type{<:FixedRational{T1}}, ::Type{T2}) where {T1,T2}
+    return promote_type(Rational{T1}, T2)
+end
+
+# Want to consume integers:
 Base.promote(x::Integer, y::F) where {F<:FixedRational} = (F(x), y)
 Base.promote(x::F, y::Integer) where {F<:FixedRational} = reverse(promote(y, x))
```
MilesCranmer, line 87: My one comment is that these should probably be turned into `promote_rule` at some point. But that's for another PR.

Also I realize this specific and unusual promotion requirement might make things a bit harder to move `FixedRational` into Ratios.jl (which I still think is a good idea, cc @oscardssmith)... But we can just figure that out later.

MilesCranmer:
Thanks again for your help @gaurav-arya!!

PR #37, opened 2023-09-04
Redefine `float` to match Unitful and return quantity

MilesCranmer:
cc @gaurav-arya

This also removes the `norm` defined on arrays, as it is no longer needed.


gaurav-arya (approved):
LGTM.

(A note on `ustrip`:  a `ustrip(::AbstractArray)` could make sense in the same spirit of auto-vectorized functions on arrays like `real`: https://github.com/JuliaLang/julia/blob/8e14322b5aa344639dd86bf9eabb84afe831fcba/base/abstractarraymath.jl#L151-L155. But reducing API surface and just requiring broadcasting sounds great to me.)

PR #38, opened 2023-09-08
Fix broadcast stack overflow

gaurav-arya:
Presumably on 1.10 the broadcast/convert logic is smart enough to not call that `convert` method in the first place. Mirrors logic in e.g. https://github.com/JuliaLang/julia/blob/b3741c01f2e2d4956be3ba41858c4065c6fbc7e8/base/array.jl#L663


MilesCranmer:
Good catch!! Thanks

PR #42, opened 2023-09-19
Change behavior of `one` to match Unitful

### LINKED ISSUE #40: Quantity currently incompatible with QuadGK due to Base.one interpretation
I've been testing out **DynamicQuantities** as a replacement for **Unitful** in some of my packages and I discovered that it currently doesn't work with **QuadGK**'s integral solver like **Unitful** does. MWE (issue also opened [here](https://github.com/JuliaMath/QuadGK.jl/issues/88)).

```julia
julia> using DynamicQuantities  # [06fc5a27] DynamicQuantities v0.7.0

julia> using QuadGK  # [1fd47b50] QuadGK v2.8.2

julia> quadgk(t -> 5u"m/s"*t, 0u"s", 10u"s")
ERROR: MethodError: no method matching cachedrule(::Type{Quantity{Float64, Dimensions{DynamicQuantities.FixedRational{Int32, 25200}}}}, ::Int64)

Closest candidates are:
  cachedrule(::Union{Type{Complex{BigFloat}}, Type{BigFloat}}, ::Integer)
   @ QuadGK C:\Users\*\.julia\packages\QuadGK\XqIlh\src\gausskronrod.jl:253
  cachedrule(::Type{T}, ::Integer) where T<:Number
   @ QuadGK C:\Users\*\.julia\packages\QuadGK\XqIlh\src\gausskronrod.jl:264
```

It looks like the issue probably spawns from [here](https://github.com/SymbolicML/DynamicQuantities.jl/blob/30ac1e8d65c7113ca4d04426375023c915de4c41/src/utils.jl#L116) where `Base.one(::Quantity)` produces a `new_quantity`.
```julia
julia> typeof(one(typeof(1u"s")))
```

**QuadGK** would expect `Base.one` for the above example to return simply a `Float64` instead. This seems consistent with the docstring for `Base.one`, excerpted:
```
If possible, one(x) returns a value of the same type as x, and one(T) returns a value of type T. However,
this may not be the case for types representing dimensionful quantities (e.g. time in days), since the
multiplicative identity must be dimensionless. In that case, one(x) should return an identity value of
the same precision (and shape, for matrices) as x.

If you want a quantity that is of the same type as x, or of type T, even if x is dimensionful, use
oneunit instead.

...

  Examples
  ≡≡≡≡≡≡≡≡≡≡

...

  julia> import Dates; one(Dates.Day(1))
  1
```

It also seems consistent with how **Unitful** handles this
```julia
julia> typeof(one(typeof(1.0u"s")))  # Unitful @u_str
Float64
```

Was the current definition of `Base.one` [here](https://github.com/SymbolicML/DynamicQuantities.jl/blob/30ac1e8d65c7113ca4d04426375023c915de4c41/src/utils.jl#L116) intended to return the dimensioned type?

MilesCranmer:
cc @mikeingold @stevengj

This is a draft PR to modify `one(q)` to return `one(ustrip(q))` rather than a dimensionless quantity, to match the behavior of Unitful.jl. This also makes `AbstractQuantity <: Number`.

Fixes #40

Also @gaurav-arya what do you think? Apparently we need this for compatibility with packages that use Unitful.

TODO:

- [ ] Turn on other unittests again (will require updating the assumed behavior of `one`)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -33,6 +33,7 @@ julia = "1.6"
 [extras]
 Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+QuadGK = "1fd47b50-473d-5c70-9696-f719f8f3bcdc"
 Ratios = "c84ed2f1-dad5-54f0-aa8e-dbefe2724439"
 SafeTestsets = "1bc83da4-3b8d-516f-aca4-4fe02f6d838f"
 SaferIntegers = "88634af6-177f-5301-88b8-7819386cfa38"
@@ -43,4 +44,4 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 
 [targets]
-test = ["Test", "Aqua", "LinearAlgebra", "Ratios", "SaferIntegers", "SafeTestsets", "ScientificTypes", "ScientificTypesBase", "StaticArrays", "Unitful"]
+test = ["Test", "Aqua", "LinearAlgebra", "QuadGK", "Ratios", "SaferIntegers", "SafeTestsets", "ScientificTypes", "ScientificTypesBase", "StaticArrays", "Unitful"]
--- a/src/math.jl
+++ b/src/math.jl
@@ -2,19 +2,19 @@ Base.:*(l::AbstractDimensions, r::AbstractDimensions) = map_dimensions(+, l, r)
 Base.:*(l::AbstractQuantity, r::AbstractQuantity) = new_quantity(typeof(l), ustrip(l) * ustrip(r), dimension(l) * dimension(r))
 Base.:*(l::AbstractQuantity, r::AbstractDimensions) = new_quantity(typeof(l), ustrip(l), dimension(l) * r)
 Base.:*(l::AbstractDimensions, r::AbstractQuantity) = new_quantity(typeof(r), ustrip(r), l * dimension(r))
-Base.:*(l::AbstractQuantity, r) = new_quantity(typeof(l), ustrip(l) * r, dimension(l))
-Base.:*(l, r::AbstractQuantity) = new_quantity(typeof(r), l * ustrip(r), dimension(r))
-Base.:*(l::AbstractDimensions, r) = error("Please use an `AbstractQuantity` for multiplication. You used multiplication on types: $(typeof(l)) and $(typeof(r)).")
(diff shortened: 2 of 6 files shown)


gaurav-arya:
Hm, I do also appreciate the property `one(::T) -> T`. This would also help with type stability when writing e.g. a multiplicative accumulator (similarly to `zero(::T) -> T` often being relied upon for type stable sums). Note that `prod` is no longer type stable with this PR, due to the empty case:
```julia
julia> q = 1u"kg/s"
1.0 kg s⁻¹

julia> arr_empty = typeof(q)[]
Quantity{Float64, Dimensions{DynamicQuantities.FixedRational{Int32, 25200}}}[]

julia> arr = [q]
1-element Vector{Quantity{Float64, Dimensions{DynamicQuantities.FixedRational{Int32, 25200}}}}:
 1.0 kg s⁻¹

julia> prod(arr_empty) |> typeof
Float64

julia> prod(arr) |> typeof

julia> @code_warntype prod(arr)
MethodInstance for prod(::Vector{Quantity{Float64, Dimensions{DynamicQuantities.FixedRational{Int32, 25200}}}})
  from prod(a::AbstractArray; dims, kw...) @ Base reducedim.jl:996
Arguments
  #self#::Core.Const(prod)
  a::Vector{Quantity{Float64, Dimensions{DynamicQuantities.FixedRational{Int32, 25200}}}}
Body::Union{Float64, Quantity{Float64, Dimensions{DynamicQuantities.FixedRational{Int32, 25200}}}}
1 ─      nothing
│   %2 = Base.:(:)::Core.Const(Colon())
│   %3 = Core.NamedTuple()::Core.Const(NamedTuple())
│   %4 = Base.pairs(%3)::Core.Const(Base.Pairs{Symbol, Union{}, Tuple{}, @NamedTuple{}}())
│   %5 = Base.:(var"#prod#831")(%2, %4, #self#, a)::Union{Float64, Quantity{Float64, Dimensions{DynamicQuantities.FixedRational{Int32, 25200}}}}
└──      return %5
```

MilesCranmer:
Great point. Do you want to repost this on #40?

gaurav-arya:
Cross-referencing https://github.com/SymbolicML/DynamicQuantities.jl/pull/42#issuecomment-1726343281: changing `one` has the potential to a function like `Base.prod` without type unstable. This wasn't a problem with `Unitful` because these functions were inherently type unstable anyway. However, `DynamicQuantities` tries very hard not to introduce type instabilities that the corresponding unitless version does not have. Thus, it seems like an alternative approach to `one(T)` might be more appropriate for interoperability with `QuadGK.jl`?

MilesCranmer:
Will cherry pick the `<: Number` but otherwise this can be closed as QuadGK is going to switch to using BaseType.jl

MilesCranmer:
Closing as we have BaseType.base_numeric_type now

PR #45, opened 2023-09-19
Ensure user-defined mutable version of dimensions always copied

MilesCranmer:
While the default `Dimensions` is immutable, and the mutable `SymbolicDimensions` (which uses a sparse array) should always create a copy anyways, the user could declare a `AbstractDimensions` that is mutable and does not copy itself.

Therefore this PR implements `copy` for a variety of operations on both `QuantityArray` and `AbstractQuantity`.

todo:
- [ ] turn off copies for broadcasting interface


MilesCranmer:
Superseded by #56

PR #46, opened 2023-09-23
Add extension for Measurements.jl

MilesCranmer:
cc @gaurav-arya 

Measurements.jl currently has an internal extension for Unitful.jl: https://github.com/JuliaPhysics/Measurements.jl/blob/21cf93c7dc25b7a1b14d5b877bc8f4884599ff5e/ext/MeasurementsUnitfulExt.jl

However since Measurements.jl is much more established and has a stabler API stable than DynamicQuantities.jl, I think it's better we set up the extension first on our side before maybe moving it there in the future (if they're open to it).

@giordano let me know what you think


ext/DynamicQuantitiesMeasurementsExt.jl
```diff
@@ -0,0 +1,21 @@
...
+using DynamicQuantities: AbstractQuantity, new_quantity, dimension, ustrip, DimensionError
+using Measurements: Measurements, measurement, value, uncertainty
+
+function Measurements.measurement(a::Q, b::Q) where {Q<:AbstractQuantity}
+    dimension(a) == dimension(b) || throw(DimensionError(a, b))
+    raw_measurement = measurement(ustrip(a), ustrip(b))
+    return new_quantity(Q, raw_measurement, dimension(a))
+end
+function Measurements.measurement(a::Q1, b::Q2) where {Q1<:AbstractQuantity, Q2<:AbstractQuantity}
+    Q = promote_type(Q1, Q2)
```
MilesCranmer, line 12: Should just use `promote`

giordano:
I'm not familiar with the internals of this package, but this looks ok as far as I can tell.

MilesCranmer:
Thanks!

PR #47, opened 2023-09-23
Some docs fixes

gaurav-arya:
- Adapt to the `Documenter.jl` breaking release `1.0.0`. This made `strict` checks on my default, which seems to be fine for most of the checks, except `:missing_docs`, which I manually excluded (that tests that every single docstring is in the public docs, which seems a bit excessive).

- Remove `index.md`, since it is auto-generated and the current file is outdated


MilesCranmer:
Nice work, LGTM!

MilesCranmer (approved):

PR #48, opened 2023-09-23
Add `uconvert` and chemistry example

gaurav-arya:
I was doing my chemistry homework using `DynamicQuantities.jl`, and thought it might make for a nice docs example! This is on top of #47, making the diff a bit noisier.

**Edit:** This PR now also implements `uconvert`, see https://github.com/SymbolicML/DynamicQuantities.jl/pull/48#discussion_r1335061100.

### DIFF
--- a/docs/src/examples.md
+++ b/docs/src/examples.md
@@ -1,0 +1,25 @@
+# Toy Examples with Code
+
+```julia
+using DynamicQuantities
+```
+
+## 1. Solving a Chemistry Homework Problem
+
+On your chemistry homework, you are faced with the following problem:
+
+> In a photoelectric effect experiment, electrons are ejected from a titanium surface (work function ``\Phi = 4.33\mathrm{eV}``) following irradition with UV light.
+> The energy of the incident UV light is ``7.2 \cdot 10^{-19} \mathrm{J}`` per photon. Calculate the wavelength of the ejected electrons.
+
+Let's solve this problem with `DynamicQuantities.jl`!
+```@repl
+using DynamicQuantities
+using DynamicQuantities.Constants: h, c, m_e
+Φ = 4.33u"Constants.eV" # work function
+E = 7.2e-19u"J" # incident energy
+p = sqrt(2 * m_e * (E - Φ)) # momentum of ejected electrons
+λ = h / p # wavelength of ejected electrons
+```
+Since units are automatically propagated, we can verify the dimension of our answer and all intermediates.
+Also, using `DynamicQuantities.Constants`, we were able to obtain the (dimensionful!) values of all necessary constants without typing them ourselves.
+
--- a/docs/src/examples.md
+++ b/docs/src/examples.md
@@ -30,7 +30,7 @@ julia> p = sqrt(2 * m_e * (E - Φ)) # momentum of ejected electrons
 julia> λ = h / p # wavelength of ejected electrons
 3.029491247878056e-9 m
 
-julia> as_units(λ, us"nm") # return answer in nanometers
+julia> as_u(λ, us"nm") # return answer in nanometers
 3.0294912478780556 nm
 ```
 Since units are automatically propagated, we can verify the dimension of our answer and all intermediates.
--- a/docs/src/examples.md
+++ b/docs/src/examples.md
@@ -11,7 +11,7 @@ On your chemistry homework, you are faced with the following problem on the phot
 [^1]: Attribution: [MIT OCW](https://ocw.mit.edu/courses/5-111sc-principles-of-chemical-science-fall-2014/resources/mit5_111f14_lec04soln/)
 
 > In a photoelectric effect experiment, electrons are ejected from a titanium surface (work function ``\Phi = 4.33\mathrm{eV}``) following irradition with UV light.
-> The energy of the incident UV light is ``7.2 \cdot 10^{-19} \mathrm{J}`` per photon. Calculate the wavelength of the ejected electrons.
+> The energy of the incident UV light is ``7.2 \cdot 10^{-19} \mathrm{J}`` per photon. Calculate the wavelength of the ejected electrons, in nanometers.
 
 Let's solve this problem with `DynamicQuantities.jl`!
 ```@repl
@@ -21,6 +21,7 @@ using DynamicQuantities.Constants: h, c, m_e
 E = 7.2e-19u"J" # incident energy
 p = sqrt(2 * m_e * (E - Φ)) # momentum of ejected electrons
 λ = h / p # wavelength of ejected electrons
+as_units(λ, us"nm") # return answer in nanometers
 ```
 Since units are automatically propagated, we can verify the dimension of our answer and all intermediates.
 Also, using `DynamicQuantities.Constants`, we were able to obtain the (dimensionful!) values of all necessary constants without typing them ourselves.
--- a/docs/src/examples.md
+++ b/docs/src/examples.md
@@ -30,7 +30,7 @@ julia> p = sqrt(2 * m_e * (E - Φ)) # momentum of ejected electrons
 julia> λ = h / p # wavelength of ejected electrons
 3.029491247878056e-9 m
 
-julia> as_u(λ, us"nm") # return answer in nanometers
+julia> uconvert(us"nm", λ) # return answer in nanometers
 3.0294912478780556 nm
 ```
 Since units are automatically propagated, we can verify the dimension of our answer and all intermediates.
--- a/docs/src/examples.md
+++ b/docs/src/examples.md
@@ -6,7 +6,9 @@ using DynamicQuantities
 
 ## 1. Solving a Chemistry Homework Problem
 
-On your chemistry homework, you are faced with the following problem:
+On your chemistry homework, you are faced with the following problem on the photoelectric effect[^1]:
+
+[^1]: Attribution: [MIT OCW](https://ocw.mit.edu/courses/5-111sc-principles-of-chemical-science-fall-2014/resources/mit5_111f14_lec04soln/)
 
 > In a photoelectric effect experiment, electrons are ejected from a titanium surface (work function ``\Phi = 4.33\mathrm{eV}``) following irradition with UV light.
 > The energy of the incident UV light is ``7.2 \cdot 10^{-19} \mathrm{J}`` per photon. Calculate the wavelength of the ejected electrons.
--- a/docs/src/examples.md
+++ b/docs/src/examples.md
@@ -1,10 +1,5 @@
 # Toy Examples with Code
 
-```jldoctest examples
-julia> using DynamicQuantities
-
-```
-
 ## 1. Solving a Chemistry Homework Problem
 
 On your chemistry homework, you are faced with the following problem on the photoelectric effect[^1]:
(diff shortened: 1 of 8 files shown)


docs/src/examples.md
```diff
@@ -0,0 +1,27 @@
...
+> The energy of the incident UV light is ``7.2 \cdot 10^{-19} \mathrm{J}`` per photon. Calculate the wavelength of the ejected electrons.
+
+Let's solve this problem with `DynamicQuantities.jl`!
+```@repl
+using DynamicQuantities
+using DynamicQuantities.Constants: h, c, m_e
+Φ = 4.33u"Constants.eV" # work function
+E = 7.2e-19u"J" # incident energy
+p = sqrt(2 * m_e * (E - Φ)) # momentum of ejected electrons
+λ = h / p # wavelength of ejected electrons
```
gaurav-arya, line 23: As one last missing piece, it would be nice to convert the final answer (`3.029491247878056e-9 m`) into nanometers, e.g. somehow convert into a symbolic `3us"nm"`. Is that possible, and if not maybe a desirable missing feature?

MilesCranmer: Not currently possible but good point that it was be a super useful feature! We could do:

```julia
function as_units(
    q::Q1, qout::Q2
) where {
    T1,T2,
    D1<:Dimensions,
    D2<:SymbolicDimensions,
    Q1<:AbstractQuantity{T1,D1},
    Q2<:AbstractQuantity{T2,D2}
}
    return new_quantity(Q2, ustrip(q) / ustrip(expand_units(qout)), dimension(qout))
end
```

MilesCranmer: I wonder if we could use some operator for this instead. Maybe like

```julia
1000u"m" |> us"km"
```

Could mean the same thing? (Would need to write a method for `(::Quantity{T,SymbolicDimensions})`).

Seems like a fairly intuitive syntax.

gaurav-arya: Hm, personally I find `as_units` already very readable! I just implemented it in this PR following your suggestion, let me know what you think

MilesCranmer: Thanks, I see. What do you think about it for the interactivity aspect? Maybe the `|>` would be nicer for quick conversions in the REPL?

MilesCranmer: (I’m assuming we could have both options maybe)

gaurav-arya: In this scenario, should the `convert` behaviour also change? Right now we have:
```julia
julia> convert(typeof(1us"nm"), 1u"m")
1.0 m
```

gaurav-arya: Oh hm, wait, I was thinking of your original method as a constructor, but now I see it's not because what is being called is the symbolic unit quantity itself rather than its type.

Hm, I'm pretty hestitant to make a quantity "callable" like that for the purpose of syntax, unless this sort of behaviour has precedent elsewhere?

MilesCranmer: Good point, I guess this is technically more common for conversion...

I’m just wondering how we could replicate the nice behavior of e.g., astropy’s unit package of `.to("<units>")` which makes unit conversions use very few key strokes. Hm.

Maybe we could have `as_units(u_out)` generate a closure function? Then you could call it like

```julia
1000u"m/s" |> as_units(us"km/s")
```

Or have `to` instead of `as_units` here...

gaurav-arya: I added a curried form of `as_units`, similar to e.g. `<=(3)` in Base. Agreed it's not as compact as `astropy`.

MilesCranmer: I wonder if we could just do `as` instead of `as_units`? Or would that be too dangerous for an exported function perhaps...

Maybe just `as_u` to keep things shorter?

MilesCranmer: I think `as_u` might also be nice because "u" is used for "units" elsewhere in the API, like `uparse`, `u"<unit>"`, etc. 

What do you think? Sorry for nitpicking here. I think this is a really great addition regardless of the function name!!

MilesCranmer: I think we could also have `as_u(s::AbstractString) = as_us(sym_uparse(s))` so there would be no need for the `us`. Then it's just

```julia
1000u"m" |> as_u("km")
```

which seems pretty nice. What do you think? Maybe its a bit wrong, I'm not sure...

gaurav-arya: No objections to either suggestions, so have implemented. I'm in two minds about `as_units` v.s. `as_u`. One thing to think about is the relation to `expand_units` -- that's a closely related function so might be worth thinking about consistency with naming. A few options off the top of my head are:

* `as_u` and `expand_units` (current)
* `as_u` and `expand_u`
* `as_units` and `expand_units`
* `uconvert` and `uexpand` (for `uconvert` we'd switch the argument order)

gaurav-arya: Hmm, apparently there's a performance penalty:
```julia
julia> @btime as_u(3u"m", us"km")
  360.081 ns (7 allocations: 288 bytes)
0.003 km

julia> @btime as_u(3u"m", "km")
  3.727 μs (61 allocations: 3.20 KiB)
0.003 km
```
Which I guess I should have expected.

MilesCranmer: > One thing to think about is the relation to `expand_units`

Ah, you are right! Great catch. Okay I take back my suggestion earlier.

What do you think is the cleanest way to approach this? I guess the goals are for it to be easy to use, but also consistent with existing API, and also semantically comparable to Unitful.jl for any new converts.

gaurav-arya: I think I personally favor `as_units` for consistency with `expand_units` and readability. I also lightly favor removing the string method for now just because the performance implications might need a bit of though, but it could easily be re-added in a separate PR. What do you think?

MilesCranmer: `uconvert` is a great idea. I did a facepalm because that was so obvious in hindsight! Sorry to keep changing my mind here.. 

I just pushed some commits to change it to `uconvert` and also change the order of the arguments to match Unitful. The existing features such as allowing strings or the curried form are still in, as I think they are really nice for interactive use.

What do you think?

(I wonder if we should take the time to also deprecate `expand_units` in favor of like `uexpand` or something... but maybe better for a different PR!)

gaurav-arya: I'm sold on `uconvert`! 

Still a little hesitant about the string method due to the possible performance implications, but I'm fine with it. On the flip side, it does significantly increase usability as people could use `uconvert` without any understanding of symbolic units. Maybe there could be a performance warning in the docstring? E.g. "For maximum performance, it is recommended to directly use a symbolic dimension as the first argument to `uconvert`, rather than the string convenience method."

Re `uexpand`: I definitely agree we should make the switch for consistency with `uconvert`. Not familiar with best practices for deprecation, so I'd appreciate if you could see that through here or subsequently as you deem appropriate, but I'm on board :) There's probably also a docs thing here where we would really like to make sure people know about `uexpand`. After all, if they can use `uconvert` with a native string to easily land with a symbolic dimension, we should make sure they know to `uexpand` it back in performance critical cases and/or cases with units of mixed scale.

MilesCranmer: I think whether we allow string or not, it should still be the same performance, right? Since the `us"m"` just expands to `sym_uparse("m")`. So it is still parsed at runtime regardless, no?

(Maybe we should make `uparse` and `sym_uparse` have an LRU cache at some point...)

MilesCranmer: But I could see `uconvert` with strings being perhaps _too_ convenient, such that new users start using it without realizing they are converting to a more expensive type... Whereas forcing them to write `us"m"` instead of `u"m"` makes it explicit.

So I guess we could delete that method anyways?

gaurav-arya: Yes, that makes sense!

src/symbolic_dimensions.jl
```diff
@@ -98,13 +98,40 @@ end
...
 end
 expand_units(q::QuantityArray) = expand_units.(q)
 
+"""
+    uconvert(qout::AbstractQuantity{<:Any, <:SymbolicDimensions}, q::AbstractQuantity{<:Any, <:Dimensions})
+    uconvert(ustr::String, q::AbstractQuantity{<:Any, <:Dimensions})
+
+Convert a quantity `q` with base SI units to the symbolic units of `qout`, for `q` and `qout` with compatible units.
+Mathematically, the result has value `q / expand_units(qout)` and units `dimension(qout)`. 
+For string input, `qout` is created by parsing `ustr` as a symbolic unit, i.e. `qout = sym_uparse(ustr)`.
```
gaurav-arya, line 116:
```suggestion
```

src/symbolic_dimensions.jl
```diff
@@ -98,13 +98,40 @@ end
...
+function uconvert(qout::AbstractQuantity{<:Any, <:SymbolicDimensions}, q::AbstractQuantity{<:Any, <:Dimensions})
+    qout_expanded = expand_units(qout)
+    dimension(q) == dimension(qout_expanded) || throw(DimensionError(q, qout_expanded))
+    return new_quantity(typeof(qout), ustrip(q) / ustrip(qout_expanded), dimension(qout))
+end
+uconvert(ustr::String, q::AbstractQuantity{<:Any, <:Dimensions}) = uconvert(sym_uparse(ustr), q) 
+
+"""
+    uconvert(qout::AbstractQuantity{<:Any, <:SymbolicDimensions})
+    uconvert(ustr::String)
```
gaurav-arya, line 127:
```suggestion
```

src/symbolic_dimensions.jl
```diff
@@ -98,13 +98,40 @@ end
...
+
+Create a function that converts an input quantity `q` with base SI units to the symbolic units of `qout`, i.e 
+a function equivalent to `q -> uconvert(qout, q)`.
+For string input, `qout` is created by parsing `ustr` as a symbolic unit, i.e. `qout = sym_uparse(ustr)`.
```
gaurav-arya, line 131:
```suggestion
```

src/symbolic_dimensions.jl
```diff
@@ -98,13 +98,40 @@ end
...
+"""
+uconvert(qout::AbstractQuantity{<:Any, <:SymbolicDimensions}) = Base.Fix1(uconvert, qout)
+uconvert(ustr::String) = uconvert(sym_uparse(ustr))
```
gaurav-arya, line 134:
```suggestion
```

test/unittests.jl
```diff
@@ -569,6 +569,9 @@ end
     @test us"Constants.h" != us"h"
     @test expand_units(us"Constants.h") == u"Constants.h"
 
+    @test uconvert(us"nm", 5e-9u"m") ≈ uconvert("nm", 5e-9u"m") ≈ (5e-9u"m" |> uconvert(us"nm")) ≈ (5e-9u"m" |> uconvert("nm")) ≈ 5us"nm"
```
gaurav-arya, line 572:
```suggestion
    @test uconvert(us"nm", 5e-9u"m") ≈ (5e-9u"m" |> uconvert(us"nm")) ≈ 5us"nm"
```

gaurav-arya (commented):
Remove string method

docs/src/examples.md
gaurav-arya: (it's also the more "conservative" thing to do, and if we find down the line that people have usability issues we could reconsider)

MilesCranmer: Oh yeah, you are right! Cool.

MilesCranmer (approved):
Looks good! Great work @gaurav-arya!!

PR #49, opened 2023-09-25
Support dispatches on `Number` via union type

### LINKED ISSUE #44: Support dispatches on `Number`, `Real`, etc.
Inspired by https://github.com/SymbolicML/DynamicQuantities.jl/issues/40#issuecomment-1725432288:

> It's too bad I can't do
> 
> ```julia
> abstract type AbstractQuantity{T,D} <: T end
> ```
> 
> which would make quantity containing a `Number` also a type `Number`.

While we can't do that, a practical solution could be to get this to work on a finite set of types `T`, e.g. `Number` and `Real`, Perhaps wrapper types could be a nice solution here, e.g. something like

```julia
struct NumberQuantityWrapper{Q} <: Number
    quantity::Q
end

struct RealQuantityWrapper{Q} <: Real
    quantity::Q
end

const QuantityWrapper = Union{NumberQuantityWrapper, RealQuantityWrapper}
ustrip(q::QuantityWrapper) = ustrip(q.quantity)
dimension(q::QuantityWrapper) = dimension(q.quantity)

function new_quantity(::Type{Q}, l, r) where {Q<:AbstractQuantity}
    quantity = constructor_of(Q)(l, r)
    if l isa Number 
        return NumberQuantityWrapper(quantity)
    elseif l isa Real
        return RealQuantityWrapper(quantity)
    else
        return quantity
    end
end
```
One difficulty here is that `NumberQuantityWrapper` and `RealQuantityWrapper` no longer live in the same type hierarchy as `AbstractQuantity`, but I think we could come up with a clean design using union types and/or metaprogramming that loops over the wrappers for easily defining dispatch rules on quantities. What do you think @MilesCranmer? 

Addendum: Although in the wrapper type approach proposed here, `AbstractQuantity` would again be generic, I am in support of subtyping `Number` as in #40 as an initial step-- as long as it does not cause any concrete immediate issues, matching `Unitful` seems like the right thing. This issue is about having our cake and eating it too:) (Supporting a few carefully selected generic types like `Real` would hopefully go a long way to increasing composability of `DynamicQuantities.jl` even further, as compared to `Unitful.jl`.)

*Edit*: We also have to think about the consequences of breaking the `AbstractQuantity` hierarchy on possible extension rules in downstream packages. Perhaps the current `AbstractQuantity` should become `AbstractGenericQuantity`, and `AbstractQuantity` should be exported and equal to `Union{AbstractGenericQuantity, QuantityWrapper}`.

MilesCranmer:
This fixes #44 and refactors much of the internals to support two abstract quantities:

- `AbstractQuantity{T,D} <: Number`. Same name, but now only `Number`
- `AbstractGenericQuantity{T,D}`. New name, supports `Any`

It does this by creating `UnionAbstractQuantity` which is a union of these two types:
```julia
const UnionAbstractQuantity{T,D} = Union{AbstractQuantity{T,D},AbstractGenericQuantity{T,D}}
```
There is also now `GenericQuantity` as a built-in `AbstractGenericQuantity`, similar to how `Quantity` is the built-in `AbstractQuantity`.

In principle one could also add `RealQuantity <: AbstractRealQuantity` and so on. I chose to not add those yet until someone needs it.

@gaurav-arya what do you think?

---

The downside of this PR is that it hurts readability due to the use of `@eval` (required to avoid method ambiguities, and if we don't want to manually write a hundred extra methods...)


src/math.jl
```diff
@@ -1,41 +1,60 @@
...
-    end
-Base.:-(l::AbstractQuantity) = new_quantity(typeof(l), -ustrip(l), dimension(l))
-Base.:-(l::AbstractQuantity, r::AbstractQuantity) = l + (-r)
+        Base.:*(l::$type, r::$base_type) = new_quantity(typeof(l), ustrip(l) * r, dimension(l))
+        Base.:/(l::$type, r::$base_type) = new_quantity(typeof(l), ustrip(l) / r, dimension(l))
+
+        Base.:*(l::$base_type, r::$type) = new_quantity(typeof(r), l * ustrip(r), dimension(r))
+        Base.:/(l::$base_type, r::$type) = new_quantity(typeof(r), l / ustrip(r), inv(dimension(r)))
+
+        Base.:*(l::$type, r::AbstractDimensions) = new_quantity(typeof(l), ustrip(l), dimension(l) * r)
```
gaurav-arya, line 12: Hm, I don't quite follow why we need to modify the `AbstractDimensions` methods to e.g. dispatch the other argument on `$type`, rather than leaving it as before. I also don't quite understand the purpose of the new methods in `disambiguities.jl`. Naively I'd hope that we could just leave all the `AbstractDimensions` methods alone, but perhaps I'm missing something?

MilesCranmer: I thought so too. But `Aqua` kept detecting various method ambiguities. A lot of the complexity in this PR is purely for the reason of eliminating different ambiguities. (Some of those are new in 1.10 so may not be visible in older versions)

MilesCranmer: Part of the issue is because `AbstractUnionQuantity` is on par with `Number` in a type hierarchy. So whenever `Number` appears in a method, I have to break the union into each type.

gaurav-arya: I did expect the ambiguities with `*(::Number, ::Number)`, those unfortunately seem avoidable when trying to make a new number type participate in the type hieracy. Here's an example of a similar flavor in `ForwardDiff` (`Dual` overloads `Real`, for reference): https://github.com/JuliaDiff/ForwardDiff.jl/blob/d3002093beb88ff0b98ed178377961dfd55c1247/src/dual.jl#L142C14-L148.

I'm mostly confused about the `AbstractDimensions` methods, rather than the `AbstractQuantity` ones. Since we have total control over how `AbstractDimensions` is used, I was surprised that the dimensions methods needed to be included in the metaprogramming loops, rather than kept outside.

gaurav-arya: To put it more simply: it seems to me that, in theory, this PR should not affect `AbstractDimensions` at all, e.g. it doesn't change its position in the type hierarchy. So why should the methods defined on it change?

MilesCranmer: Oh I think I understand your question now. The answer lies two steps away. First: there's a method:
```julia
Base.:*(l::AbstractDimensions, r::Number)
```
(whose only job is to raise a helpful error).

In this PR, before the ambiguities, I had defined

```julia
Base.:*(l::AbstractDimensions, r::AbstractUnionQuantity)
```

But because `AbstractUnionQuantity` and `Number` are at the same level of the hierarchy (one above `AbstractQuantity`), I therefore had to split the `AbstractUnionQuantity` into

```julia
Base.:*(l::AbstractDimensions, r::AbstractQuantity)
Base.:*(l::AbstractDimensions, r::AbstractGenericQuantity)
```

to get rid of the ambiguity.

src/utils.jl
```diff
@@ -26,113 +26,119 @@ end
...
+function Base.promote_rule(::Type{<:GenericQuantity{T1,D1}}, ::Type{<:GenericQuantity{T2,D2}}) where {T1,T2,D1,D2}
+    return GenericQuantity{promote_type(T1,T2),promote_type(D1,D2)}
+end
+function Base.promote_rule(::Type{<:Quantity{T1,D1}}, ::Type{<:GenericQuantity{T2,D2}}) where {T1,T2,D1,D2}
+    return GenericQuantity{promote_type(T1,T2),promote_type(D1,D2)}
+end
+function Base.promote_rule(::Type{<:Quantity{T1,D1}}, ::Type{<:Quantity{T2,D2}}) where {T1,T2,D1,D2}
+    return Quantity{promote_type(T1,T2),promote_type(D1,D2)}
+end
+function Base.promote_rule(::Type{T1}, ::Type{<:Quantity{T2,D2}}) where {T1<:Number,T2,D2}
```
gaurav-arya, line 47: If this rule is needed for `Quantity`, should there also be an analogous promotion rule for `GenericQuantity`? I think that promotion with generic quantities might be one of the most important things to add a few tests for in this PR.

gaurav-arya (commented):
This is an *extremely* cool PR: I don't know of any other number type packages that have made the effort to work around the `MyStruct{T} <: T` conundrum. (the closest might be symbolics having both generic variables and a special `Num <: Real`?). 

I have admittedly not had time yet to review line by line, but some general comments:

**Naming:** TL;DR everything seems fine:) I wondered whether `AbstractUnionQuantity` should really be `AbstractQuantity`. However, I think it's important for it to be clear that this is a union type (and can't e.g. be subtyped), so I think you've already made the right choice. Relatedly, should `AbstractQuantity <: Number` really be `AbstractNumberQuantity <: Number`. I'm fine with your choice here too: special casing `Number` as the shortest name / default type seems reasonable, as that's what `Unitful` worked with. Presumably future additions would like `AbstractRealQuantity <: Real`` etc.

**Code readability:** I don't think code readability is hurt that much in this PR (the net increase in code lines in impressively small). Further, I hope in a future PR to substantially reduce the code size of e.g.`math.jl` by classifying the overloads into their maybe 3 or 4 different kinds and then having a simple loop over functions for each (or a convenience macro to define the overload). #64 is also somewhat related. That would hopefully increase readability further, as what we're seeing now is the multiplicative explosion of the looping logic together with the original repeated-overload complexity.

**Unit parsing:** Right now, AFAIU unit strings are always parsed as an instance of `Quantity`. This will be an interesting thing to revisit if were to want e.g. an `AbstractRealQuantity` in the future, as we may want `1.0 * u"m" <: Real` for example. But I don't think there's any issue right now.

MilesCranmer:
Awesome, thanks so much @gaurav-arya!!

> Relatedly, should `AbstractQuantity <: Number` really be `AbstractNumberQuantity <: Number`. I'm fine with your choice here too: special casing `Number` as the shortest name / default type seems reasonable, as that's what `Unitful` worked with. Presumably future additions would look like `AbstractRealQuantity <: Real` etc.

Exactly my thoughts as well. i.e., keeping it `AbstractQuantity` just for `Unitful` compatibility. We could even have `AbstractNumberQuantity` as a synonym for it if needed, in case someone is confused why there is `AbstractRealQuantity` but no `Number` equivalent.

> Further, I hope in a future PR to substantially reduce the code size of e.g.`math.jl` by classifying the overloads into their maybe 3 or 4 different kinds and then having a simple loop over functions for each (or a convenience macro to define the overload).

Great idea! We could use [TypedDelegation.jl](https://github.com/JeffreySarnoff/TypedDelegation.jl) for this too if needed.

> **Unit parsing:** Right now, AFAIU unit strings are always parsed as an instance of `Quantity`. This will be an interesting thing to revisit if were to want e.g. an `AbstractRealQuantity` in the future, as we may want `1.0 * u"m" <: Real` for example. But I don't think there's any issue right now.

Sounds good, we can bring this part up again later!

src/utils.jl
MilesCranmer: Good point.

MilesCranmer: Deleted this promotion rule as I realized it was way too powerful, and would basically turn everything it touches into Quantities. Instead should just promote to a union type or `Any` (already provided by `Base`), or have the user write a promotion method for their type.

src/types.jl
```diff
@@ -75,12 +90,16 @@ end
 (::Type{D})(::Type{R}; kws...) where {R,D<:AbstractDimensions} = constructor_of(D){R}((tryrationalize(R, get(kws, k, zero(R))) for k in static_fieldnames(D))...)
 (::Type{D})(; kws...) where {R,D<:AbstractDimensions{R}} = constructor_of(D)(R; kws...)
 (::Type{D})(; kws...) where {D<:AbstractDimensions} = D(DEFAULT_DIM_BASE_TYPE; kws...)
-(::Type{D})(d::AbstractDimensions) where {R,D<:AbstractDimensions{R}} = D((getproperty(d, k) for k in static_fieldnames(D))...)
+function (::Type{D})(d::D2) where {R,D<:AbstractDimensions{R},D2<:AbstractDimensions}
+    issetequal(static_fieldnames(D), static_fieldnames(D2)) ||
+        error("Cannot create a dimensions of `$(D)` from `$(D2)`. Please write a custom method for construction.")
```
MilesCranmer, line 95: I noticed this potential bug while writing this PR. It's unrelated to the goal of this PR but it was needed to fix some of the issues that came up, so I'm including it.

MilesCranmer:
Okay I added a ton of unittests for `GenericQuantity`. This should be ready to go now. Do you want to take a look @gaurav-arya?

(The errors seem to only happen on Julia <= 1.9; I'll fix those now)

src/utils.jl
MilesCranmer: Actually, wait, this is needed, otherwise the code

```julia
[0.5, 0.5u"km"]
```

will return an error (since it will try to convert to `Float64`)... Hmm...

MilesCranmer:
Do we want this as a fallback method?

```julia
function Base.promote_rule(::Type{<:Quantity}, ::Type{<:Number})
    return Number
end
```

then you could have

```julia
julia> [0.5, 0.5u"km"]
2-element Vector{Number}:
       0.5
 500.0 m
```

which seems reasonable.

src/utils.jl
gaurav-arya: Hm, I'm not sure I follow -- what was wrong with turning numbers into quantities? Then `[0.5, 0.5u"km"]` would for example be concretly typed, which seems to fit the spirit of removing type stability footguns

MilesCranmer: My concern is that if you had this behavior in other abstract number packages, there would an ambiguity about which type takes priority. For example, say you had

```julia
[0.5 ± 0.1, 0.5u"km"]
```

would this get promoted to a `Vector{Measurement{Quantity{Float64,D}}}` or a `Vector{Quantity{Measurement{Float64},D}}`?

With this PR, it simply gets promoted to a `Vector{Number}` which I think is more conservative. Thoughts?

gaurav-arya: What if we did the more aggressive thing for types `T` satisfying `base_number_type(T) = T`?

gaurav-arya: But I'd also be happy punting this and making an issue for more careful thought about promotion and its relation to array types. The behaviour here is indeed slightly better than `main`, so I don't think anything must be addressed in this PR in any case. (Perhaps too idealistically, it would be nice to even figure out how to make something concrete in cases with measurements etc. The principle being that if `[0.5 ± 0.1, 0.2]` makes something concrete, then replacing `0.2` with a quantity should not hurt type stability.)

src/types.jl
```diff
@@ -109,27 +128,50 @@ dimensions according to the operation.
...
+
+
+for (type, base_type) in ABSTRACT_QUANTITY_TYPES
+    @eval begin
+        (::Type{Q})(x::T, ::Type{D}; kws...) where {D<:AbstractDimensions,T<:$base_type,T2,Q<:$type{T2}} = constructor_of(Q)(convert(T2, x), D(; kws...))
+        (::Type{Q})(x::$base_type, ::Type{D}; kws...) where {D<:AbstractDimensions,Q<:$type} = constructor_of(Q)(x, D(; kws...))
+        (::Type{Q})(x::T; kws...) where {T<:$base_type,T2,Q<:$type{T2}} = constructor_of(Q)(convert(T2, x), dim_type(Q)(; kws...))
+        (::Type{Q})(x::$base_type; kws...) where {Q<:$type} = constructor_of(Q)(x, dim_type(Q)(; kws...))
+    end
+    for (type2, _) in ABSTRACT_QUANTITY_TYPES
```
gaurav-arya, line 159: Can we skip this when `type2 = type1`?

gaurav-arya, line 159: Can we skip this when `type2 = type1`?

gaurav-arya:
Sorry, ignore any notifications so far, I accidentalyl submitted some pending comments.

src/arrays.jl
```diff
@@ -156,10 +156,10 @@ end
 materialize_first(bc::Base.Broadcast.Broadcasted) = bc.f(materialize_first.(bc.args)...)
 
 # Base cases
-materialize_first(q::AbstractQuantity{<:AbstractArray}) = new_quantity(typeof(q), first(ustrip(q)), dimension(q))
-materialize_first(q::AbstractQuantity) = q
+materialize_first(q::AbstractGenericQuantity{<:AbstractArray}) = new_quantity(typeof(q), first(ustrip(q)), dimension(q))
```
gaurav-arya, line 159: I wonder what our level of support of array-containing `Quantity`'s is, e.g. do we expect users to ever use them in the wild, or just something we support because we can, and we can ponder their use as time goes on? They don't seem to offer anything more than `QuantityArray`s (although maybe they could replace them -- they do at least already carry the information that allows the compiler to in theory lift out dimension calculations)

src/utils.jl
```diff
@@ -26,113 +26,119 @@ end
...
-Base.ndims(::Type{<:AbstractQuantity{T}}) where {T} = ndims(T)
-Base.ndims(q::AbstractQuantity) = ndims(ustrip(q))
-Base.broadcastable(q::AbstractQuantity) = new_quantity(typeof(q), Base.broadcastable(ustrip(q)), dimension(q))
-Base.getindex(q::AbstractQuantity, i...) = new_quantity(typeof(q), getindex(ustrip(q), i...), dimension(q))
-Base.keys(q::AbstractQuantity) = keys(ustrip(q))
+Base.ndims(::Type{<:AbstractUnionQuantity{T}}) where {T} = ndims(T)
+Base.ndims(q::AbstractUnionQuantity) = ndims(ustrip(q))
+Base.broadcastable(q::AbstractUnionQuantity) = new_quantity(typeof(q), Base.broadcastable(ustrip(q)), dimension(q))
+for (type, _) in ABSTRACT_QUANTITY_TYPES
+    @eval Base.getindex(q::$type) = new_quantity(typeof(q), getindex(ustrip(q)), dimension(q))
```
gaurav-arya, line 67: what's going on here

src/utils.jl
```diff
@@ -176,94 +182,93 @@ tryrationalize(::Type{R}, x) where {R} = isinteger(x) ? convert(R, round(Int, x)
...
 Base.convert(::Type{D}, d::AbstractDimensions) where {D<:AbstractDimensions} = d
 Base.convert(::Type{D}, d::AbstractDimensions) where {R,D<:AbstractDimensions{R}} = D(d)
 
 Base.copy(d::D) where {D<:AbstractDimensions} = map_dimensions(copy, d)
-Base.copy(q::Q) where {Q<:AbstractQuantity} = new_quantity(Q, copy(ustrip(q)), copy(dimension(q)))
+Base.copy(q::Q) where {Q<:AbstractUnionQuantity} = new_quantity(Q, copy(ustrip(q)), copy(dimension(q)))
 
 """
-    ustrip(q::AbstractQuantity)
+    ustrip(q::AbstractUnionQuantity)
```
gaurav-arya, line 196: A bit of a premature comment since there are no docs yet, but if `AbstractUnionQuantity` is used in the API description (as it probably should), then it should be clearly documented to the user (and maybe exported?)

gaurav-arya:
Overall I only had minor comments:) I made a bit of a mess of the review though, sorry about that..

src/utils.jl
MilesCranmer: > What if we did the more aggressive thing for types `T` satisfying `base_number_type(T) = T`?

That's a good idea. Although what would happen if you have `[1, 1.0, 1.0u"km"]`? Would it first try to promote `[1, 1.0]` to `[1.0, 1.0]` and then move all to quantities? And would it differ from `[1, 1.0u"km", 1.0]`?

MilesCranmer: Punting it sounds good!

src/types.jl
MilesCranmer: I don't think so, because these are abstract types. You would still want this to cover `MyQuantity1(::MyQuantity2)` even if those are **both** `AbstractQuantity`. Whether they are from different abstract types isn't important though

src/utils.jl
MilesCranmer: Good point! Maybe in the docs it can be split into separate cases for `AbstractQuantity` and `AbstractGenericQuantity` so that users don't get confused.

src/math.jl
```diff
@@ -1,41 +1,60 @@
...
+                iszero(dimension(l)) || throw(DimensionError(l, r))
+                new_quantity(typeof(l), ustrip(l) + r, dimension(l))
+            end
+        Base.:+(l::$base_type, r::$type) =
+            let
+                iszero(dimension(r)) || throw(DimensionError(l, r))
+                new_quantity(typeof(r), l + ustrip(r), dimension(r))
+            end
+
+        Base.:-(l::$type, r::$type) = l + (-r)
```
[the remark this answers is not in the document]
MilesCranmer: Done! I refactored all of this part; it's much cleaner now.

MilesCranmer:
Thanks!

I just added some more documentation examples. Let me know what you think?

MilesCranmer:
Not sure what’s going on with the automated benchmarks, but if I run the benchmarks locally it seems to have no effect on the performance. Although I’m using 1.10 locally so perhaps there’s some difference on 1.9...

src/symbolic_dimensions.jl
```diff
@@ -99,35 +99,35 @@ end
...
 
 """
-    uconvert(qout::AbstractQuantity{<:Any, <:SymbolicDimensions}, q::AbstractQuantity{<:Any, <:Dimensions})
+    uconvert(qout::AbstractUnionQuantity{<:Any, <:SymbolicDimensions}, q::AbstractUnionQuantity{<:Any, <:Dimensions})
 
 Convert a quantity `q` with base SI units to the symbolic units of `qout`, for `q` and `qout` with compatible units.
 Mathematically, the result has value `q / uexpand(qout)` and units `dimension(qout)`. 
 """
-function uconvert(qout::AbstractQuantity{<:Any, <:SymbolicDimensions}, q::AbstractQuantity{<:Any, <:Dimensions})
+function uconvert(qout::AbstractUnionQuantity{<:Any, <:SymbolicDimensions}, q::AbstractUnionQuantity{<:Any, <:Dimensions})
```
gaurav-arya, line 122: Since last review this method (along with other methods for `uconvert`, along with a method for `uexpand`) generalized from just `AbstractQuantity` to `AbstractUnionQuantity`, but it doesn't seem like corresponding unit tests were added. I guess code coverage wouldn't catch this, but would it make sense to add unit tests that test this method in a more general setting (e.g. with `qout` being a `GenericQuantity`?)

gaurav-arya (commented):
Some small comments:)

gaurav-arya:
The performance regression on 1.9 appears to be real:

```julia 

julia> q = 1u"m"
1.0 m

julia> @btime inv($q)
  254.961 ns (8 allocations: 768 bytes)
```

src/types.jl
```diff
@@ -75,12 +96,16 @@ end
 (::Type{D})(::Type{R}; kws...) where {R,D<:AbstractDimensions} = constructor_of(D){R}((tryrationalize(R, get(kws, k, zero(R))) for k in static_fieldnames(D))...)
 (::Type{D})(; kws...) where {R,D<:AbstractDimensions{R}} = constructor_of(D)(R; kws...)
 (::Type{D})(; kws...) where {D<:AbstractDimensions} = D(DEFAULT_DIM_BASE_TYPE; kws...)
-(::Type{D})(d::AbstractDimensions) where {R,D<:AbstractDimensions{R}} = D((getproperty(d, k) for k in static_fieldnames(D))...)
+function (::Type{D})(d::D2) where {R,D<:AbstractDimensions{R},D2<:AbstractDimensions}
+    issetequal(static_fieldnames(D), static_fieldnames(D2)) ||
```
gaurav-arya, line 100: This check seems to have caused the regression on 1.9; I'm not sure exactly why, but presumably the line isn't running entirely at compile time as we would expect on 1.9?

MilesCranmer: Good catch! Maybe `==` is the right operator here…

MilesCranmer: Okay I fixed it with a generated function for the fieldnames equality check. I guess 1.9 just didn't want to inline it.

test/unittests.jl
```diff
@@ -10,9 +11,9 @@ using Test
 
 @testset "Basic utilities" begin
 
-    for T in [DEFAULT_VALUE_TYPE, Float16, Float32, Float64], R in [DEFAULT_DIM_BASE_TYPE, Rational{Int16}, Rational{Int32}, SimpleRatio{Int}, SimpleRatio{SafeInt16}]
+    for Q in [Quantity, GenericQuantity], T in [DEFAULT_VALUE_TYPE, Float16, Float32, Float64], R in [DEFAULT_DIM_BASE_TYPE, Rational{Int16}, Rational{Int32}, SimpleRatio{Int}, SimpleRatio{SafeInt16}]
```
MilesCranmer, line 14: @gaurav-arya much of `AbstractUnionQuantity` should be covered by this generalization. Maybe I could do it for all the tests but since they are identical I guess we don't need too much... Just need to be careful for things that are not generalized to AbstractUnionQuantity

docs/src/examples.md
```diff
@@ -30,6 +31,356 @@ julia> λ = h / p # wavelength of ejected electrons
...
+The main reason you would use a custom quantity is if you want
+to change built-in behavior, or maybe have special methods for
+different types of quantities.
+
+Note that you can declare a method on `AbstractQuantity`, or
+`AbstractGenericQuantity` to allow their respective inputs.
+
+**Note**: In general, you should probably
+specialize on `AbstractUnionQuantity` which is
+the union of these two abstract quantities, _as well as any other future abstract quantity types_,
```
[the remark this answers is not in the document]
MilesCranmer: Done!

src/symbolic_dimensions.jl
MilesCranmer: Okay I looped over a lot of the unittests and now test both `Quantity` and `GenericQuantity`. Also added special tests for `uconvert`.

MilesCranmer:
I fixed the performance regression. Looks like it was just something that the Julia compiler was able to inline on Julia 1.10 but not the earlier versions. I just moved it to a generated function now.

MilesCranmer:
Thanks for the review! Here are the changes since your last pass over: https://github.com/SymbolicML/DynamicQuantities.jl/compare/3bb945261b0a983d40ad6ea1d8b68b77ef53cecf...union-type

Regarding test coverage, I'm a bit worried but not too much because `GenericQuantity` should be the exact same as `Quantity`. The only cases we need to worry about are: (1) where both `GenericQuantity` and `Quantity` can be used as inputs, like where there are two `AbstractUnionQuantity` arguments, and (2) where `Quantity` is the default for something (we should have a `GenericQuantity` alternative). I just added another field to `ABSTRACT_QUANTITY_TYPES` to set the default concrete type in such cases so its easier to loop.

i.e., I went through and added a `GenericQuantity` alternative default for `QuantityArray`, which is triggered when an array has non-numeric type. Let me know what you think.

Also, just to be safe, I'm looping over both quantity types in a lot of the unittests now, particularly the basic tests and the QuantityArray tests.

MilesCranmer:
Btw, should it be `UnionAbstractQuantity` instead? Because we want to have `AbstractXQuantity` be subtyped to `X`, it might have better semantics.

Also, maybe `AbstractAnyQuantity` and `AnyQuantity` instead of `Generic` - wdyt?

MilesCranmer:
Okay I changed it to `UnionAbstractQuantity`. However I'm going to leave the generic one as `GenericQuantity` since `AnyQuantity` makes me think it **is** an Any, which is bad.

Any last comments or do you think it's good to go?

src/internal_utils.jl
```diff
@@ -0,0 +1,25 @@
+"""
+This file contains utility functions that are not specific to the
+library, but are used throughout.
+"""
+
+@generated function fieldnames_equal(::Type{T1}, ::Type{T2}) where {T1,T2}
+    # Needs to be a generated function as might not be inlined
+    return Base.fieldnames(T1) == Base.fieldnames(T2)
```
gaurav-arya, line 8: Does this line make sense for `SymbolicDimensions`? It looks like `SymbolicDimensions` directly overloads `static_fieldnames` in order to indicate its fields, and this wouldn't be reflected in `fieldnames`? https://github.com/SymbolicML/DynamicQuantities.jl/blob/09a43cc4d9dd55c596d41a852d65342bdbe4b49f/src/symbolic_dimensions.jl#L42

Separately, it doesn't seem right to make `static_fieldnames` the API to the extent that it needs to be overloaded by `SymbolicDimensions`. Maybe better to make the API something like `dimension_names`, which defaults to `static_fieldnames`? I'll leave it to your discretion what (if anything) needs to be addressed here rather than in a seprate PR.

gaurav-arya (commented):
Sorry for the slow review -- nearly all looks good to me! I just have a note about the fieldnames check.

src/internal_utils.jl
MilesCranmer: Good catch!

src/internal_utils.jl
```diff
@@ -0,0 +1,25 @@
+"""
+This file contains utility functions that are not specific to the
+library, but are used throughout.
+"""
+
+@generated function fieldnames_equal(::Type{T1}, ::Type{T2}) where {T1,T2}
+    # Needs to be a generated function as might not be inlined
```
gaurav-arya, line 7: I'm not sure if "inlined" is the right word here, maybe better to simply say "to ensure the check occurs at compile time"?

src/internal_utils.jl
MilesCranmer: Going to add this in a future PR

MilesCranmer:
Thanks for the reviews!!

gaurav-arya (approved):

Issue #53, opened 2023-10-09
Make SparseArrays a weak dependency?


devmotion:
In Julia >= 1.10, SparseArrays won't be included in the system image anymore. To reduce loading times, therefore SparseArrays was or is about to be demoted from a regular to a weak dependency for many packages in the ecosystem such as FillArrays, Distances and PDMats. However, if a package in the dependency stack loads SparseArrays all these extensions will be compiled and loaded. Could the SparseArrays dependency of DynamicQuantities also be moved to an extension or removed alltogether to avoid that it triggers all these extensions?

MilesCranmer:
Thanks for raising this @devmotion. I'm not sure if there's a way around this because SparseArrays is actually used internally for the SymbolicDimensions struct: https://github.com/SymbolicML/DynamicQuantities.jl/blob/87e1eb3d0d4c139ed94a563e2fae3cababc5e512/src/symbolic_dimensions.jl#L37-L42

It is used here because there are ~100 physical symbolic dimensions (`m` and `cm` are separate dimensions), so rather than store 100 fields, one for each dimension, I store it as a single sparse vector.

Perhaps it is possible to write a small internal sparse vector library to get this working without loading all of SparseArray.

What do you think?

devmotion:
I wonder if it would be possible to replace `_data` with a dictionary or, more similar to the sparse array, with two vectors `ind::Vector{Int}` (or even `sym::Vector{Symbol}`?) and `val::Vector{R}` of indices of non-zero symbolic dimensions and their values.

MilesCranmer:
Definitely possible. But it would add to the maintenance/testing burden to need to unit test a custom sparse array package, and possibly introduce new bugs. (Whereas SparseArrays work out-of-the-box)

devmotion:
Maybe my suggestion was not completely clear: I don't think one should implement a new sparse array type but possibly a different data structure (array or two vectors) could provide the same functionality without significant regressions. And since `SparseVector` is implemented based on two vectors, I think the alternative with two vectors would be a natural candidate (https://github.com/JuliaSparse/SparseArrays.jl/blob/034d23420bba5a8688fba8d857fc7f813f121f31/src/sparsevector.jl#L38-L39).

PR #54, opened 2023-10-09
Make package a bit more lightweight and improve loading time

devmotion:
In the spirit of #53, this makes the package a bit more lightweight and faster to load, in particular on Julia < 1.9:

- Implement package extensions as described in the Pkg docs without PkgExtensionsCompat.jl. This removes an additional layer of complexity and possible source of bugs (it seems most/all issues with the package are fixed but as former issues regarding unsupported use cases and bugs show this concern is real) and significant overhead in older Julia versions where the code in PkgExtensionsCompat.jl is executed. For instance, on Julia 1.6.5 I get on the master branch
  ```julia
  julia> ] precompile

  julia> @time using DynamicQuantities
    0.359792 seconds (1.32 M allocations: 83.207 MiB, 2.38% gc time, 71.25% compilation time)
  ```
  and with this PR
  ```julia
  julia> ] precompile

  julia> @time using DynamicQuantities
    0.140393 seconds (461.62 k allocations: 29.999 MiB, 28.96% compilation time)
  ```
  On Julia 1.9, the effect is less noticeable (since the package is basically empty), on the master branch one gets
  ```julia
  julia> @time_imports using DynamicQuantities
      0.5 ms  Tricks
      0.3 ms  Compat
      0.4 ms  Compat → CompatLinearAlgebraExt
      0.4 ms  PackageExtensionCompat
     44.3 ms  DynamicQuantities
      0.4 ms  DynamicQuantities → DynamicQuantitiesLinearAlgebraExt
  ```
  whereas with this PR I obtain
  ```julia
  julia> @time_imports using DynamicQuantities
      0.4 ms  Tricks
      0.3 ms  Compat
      0.4 ms  Compat → CompatLinearAlgebraExt
     44.1 ms  DynamicQuantities
      0.4 ms  DynamicQuantities → DynamicQuantitiesLinearAlgebraExt
  ```
- Removes the TOML dependency and its `using` statement: It's only used for extracting the version number from Project.toml, which IMO is a bit questionable. This feature is not used in the package, and if desired hardcoding the version number and adding a test for its consistency with Project.toml would be much simpler and faster. Generally, I have the feeling though that such a functionality (obtaining the version of an installed package) should rather be provided by Pkg or a Pkg utility package (I would assume such a package already exists?).
- Removes the weak dependency and test dependency on ScientificTypesBase: It is re-exported by ScientificTypes, so it is safe to access it via `ScientificTypes.ScientificTypesBase` since it can't be removed from ScientificTypes in a non-breaking release. This simplifies the dependency structure and the Requires hooks.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -5,30 +5,27 @@ version = "0.7.1"
 
 [deps]
 Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
-PackageExtensionCompat = "65ce6f38-6b18-4e1d-a461-8949797d7930"
+Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
-TOML = "fa267f1f-6049-4f14-aa54-33bafae1ed76"
 Tricks = "410a4b4d-49e4-4fbc-ab6d-cb71b17b3775"
 
 [weakdeps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Measurements = "eff96d63-e80a-5855-80a2-b1b0885c5ab7"
 ScientificTypes = "321657f4-b219-11e9-178b-2701a2544e81"
-ScientificTypesBase = "30f210dd-8aff-4c5f-94ba-8e64358c1161"
 Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 
 [extensions]
 DynamicQuantitiesLinearAlgebraExt = "LinearAlgebra"
 DynamicQuantitiesMeasurementsExt = "Measurements"
-DynamicQuantitiesScientificTypesExt = ["ScientificTypes", "ScientificTypesBase"]
+DynamicQuantitiesScientificTypesExt = "ScientificTypes"
 DynamicQuantitiesUnitfulExt = "Unitful"
 
 [compat]
-Compat = "^3.42, 4"
+Compat = "3.42, 4"
 Measurements = "2"
-PackageExtensionCompat = "1"
+Requires = "1"
 ScientificTypes = "3"
-ScientificTypesBase = "3"
 Tricks = "0.1"
 Unitful = "1"
 julia = "1.6"
@@ -41,22 +38,9 @@ Ratios = "c84ed2f1-dad5-54f0-aa8e-dbefe2724439"
 SafeTestsets = "1bc83da4-3b8d-516f-aca4-4fe02f6d838f"
 SaferIntegers = "88634af6-177f-5301-88b8-7819386cfa38"
 ScientificTypes = "321657f4-b219-11e9-178b-2701a2544e81"
-ScientificTypesBase = "30f210dd-8aff-4c5f-94ba-8e64358c1161"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 
 [targets]
-test = [
-    "Aqua",
-    "LinearAlgebra",
-    "Measurements",
-    "Ratios",
-    "SaferIntegers",
-    "SafeTestsets",
-    "ScientificTypes",
-    "ScientificTypesBase",
-    "StaticArrays",
-    "Test",
-    "Unitful"
-]
+test = ["Aqua", "LinearAlgebra", "Measurements", "Ratios", "SaferIntegers", "SafeTestsets", "ScientificTypes", "StaticArrays", "Test", "Unitful"]
--- a/src/DynamicQuantities.jl
+++ b/src/DynamicQuantities.jl
@@ -1,15 +1,5 @@
 module DynamicQuantities
 
-import TOML: parsefile
-
-const PACKAGE_VERSION = try
-    let project = parsefile(joinpath(pkgdir(@__MODULE__), "Project.toml"))
-        VersionNumber(project["version"])
-    end
-catch
-    VersionNumber(0, 0, 0)
-end
-
 export Units, Constants
 export AbstractQuantity, AbstractDimensions
 export Quantity, Dimensions, SymbolicDimensions, QuantityArray, DimensionError
@@ -27,13 +17,20 @@ include("constants.jl")
 include("uparse.jl")
 include("symbolic_dimensions.jl")
 
-import PackageExtensionCompat: @require_extensions
 import .Units
 import .Constants
 import .UnitsParse: uparse, @u_str
 
+if !isdefined(Base, :get_extension)
+using Requires
+end
+@static if !isdefined(Base, :get_extension)
 function __init__()
-    @require_extensions
+    @require LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e" include("../ext/DynamicQuantitiesLinearAlgebraExt.jl")
+    @require Measurements = "eff96d63-e80a-5855-80a2-b1b0885c5ab7" include("../ext/DynamicQuantitiesMeasurementsExt.jl")
+    @require ScientificTypes = "321657f4-b219-11e9-178b-2701a2544e81" include("../ext/DynamicQuantitiesScientificTypesExt.jl")
+    @require Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d" include("../ext/DynamicQuantitiesUnitfulExt.jl")
+end
 end
 
 end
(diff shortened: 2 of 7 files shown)


MilesCranmer:
This sounds great, thanks.

However, I would vote for leaving PkgExtensionCompat as-is, since it is zero cost on new versions of Julia.

Is it truly worth the maintenance burden of manually writing extensions all to get 0.2 s startup improvement on old Julia versions that few people use (which already have very slow startups)? I definitely want DynamicQuantities to be (1) correct and (2) fast on old Julia, but startup time seems like a second order effect not worth the effort.

devmotion:
> since it is zero cost on new versions of Julia.

On new Julia versions it has almost zero loading time but it's measurable nevertheless, and it requires an additional dependency with additional compilation that can be avoided.

In my opinion, arguably the officially documented way of writing optional package extensions with backwards compatibility does not involve any significant additional maintenance. In particular, once it is set up, maintenance should be almost equivalent both with and without PackageExtensionsCompat, the only difference being that without PackageExtensionsCompat one has to duplicate modifications to the `using ...` statements.

Generally, I think it's good to improve performance even on older Julia versions (particularly since the Julia LTS is still 1.6), especially if the necessary modifications are simple. I'd argue following the Pkg docs is sufficiently simple here, but of course not everyone will agree with this opinion 🙂

MilesCranmer:
## PackageExtensionCompat

Would you be okay to split out the PackageExtensionCompat removal into a separate PR as that change is a bit more contentious? We could discuss it there. My motivation for using it – which I do in all of my Julia packages – is that I have run into bugs in the past from the manual approach due to mixing up the correct sequence of `import ..`, `isdefined(Base, :get_extension)`, and the `__init__`.

I like PackageExtensionCompat because it manages all of this for you. For me it is more about about reducing the rate of human errors (for when I eventually add another extension); I think hurting startup time by 0.2% for a 1% improvement in reliability is a pretty good tradeoff, no?

## TOML

TOML is in the Julia image so I don't follow why this needs to be removed as a dep? I don't think there is any startup cost. If you help me understand why I am open to it.

The `PACKAGE_VERSION` constant is a more robust way to set package versions rather than using Pkg, since it will get compiled into the module when even the `Project.toml` might not be available. I was recommended this strategy on discourse and have used it in libraries I contribute to ever since. It's zero cost and I find it's more reliable for testing/performance tracking.

## ScientificTypesBase

Good point, happy to make this change!

## SparseArrays

^ Also, were you thinking of including this change (to use a simple sparse array struct) in this PR or separately?

SparseArrays seems to be about 90% of the startup time so I definitely agree that that would be the biggest change, compared with these smaller tweaks!

devmotion:
I'll open a new PR with the PackageExtensionsCompat changes. In the end it's of course your call since you're the maintainer of this package, but to me it's still unclear why there should be additional maintenance burden by following the official/vanilla approach, I don't see any challenges with it in particular once the first extension is implemented. I guess I'm more worried about relying on an additional non-official tool that causes loading time regressions and has a history of introducing bugs to the workflow.

I also checked Pkg quickly and indeed there is an official way to retrieve the version of a package that imported a module:
```julia
julia> import DynamicQuantities, Pkg

julia> Pkg.pkgversion(DynamicQuantities)
v"0.7.1"

help?> Pkg.pkgversion
  pkgversion(m::Module)

  Return the version of the package that imported module m, or nothing if m was not imported from a package, or imported from a package without a version field set.

  The version is read from the package's Project.toml during package load.

  To get the version of the package that imported the current module the form pkgversion(@__MODULE__) can be used.

  │ Julia 1.9
  │
  │  This function was introduced in Julia 1.9.
```
To me that seems simpler than defining non-standard constants (many packages don't define `PACKAGE_VERSION`, so you can't query the version of an arbitrary package based on it). Moreover, `pkgversion` even works with submodules:
```julia
julia> Pkg.pkgversion(DynamicQuantities.Units)
v"0.7.1"
```

MilesCranmer:
> `Pkg.pkgversion`

Cool! This must be a new addition to the stdlib. 

Okay with this we can definitely remove the TOML business.

> PackageExtensionCompat

I would still like to leave this in if that’s okay? The only bugs I have had related to this sort of thing have been from following the official/manual approach to get compatibility for earlier Julia versions and making mistakes. I just find the official syntax to be pretty unintuitive/tedious. PackageExtensionCompat is just so much simpler I don’t really want to go back...

devmotion:
I reverted the PackageExtensionCompat change.

src/DynamicQuantities.jl
```diff
@@ -1,14 +1,6 @@
...
-import TOML: parsefile
-
-const PACKAGE_VERSION = try
-    let project = parsefile(joinpath(pkgdir(@__MODULE__), "Project.toml"))
-        VersionNumber(project["version"])
-    end
-catch
-    VersionNumber(0, 0, 0)
-end
+using PackageExtensionCompat
```
MilesCranmer, line 3:
```suggestion
import PackageExtensionCompat: @require_extensions
```

(Personal preference to use explicit imports so its easier to figure out what symbol came from where, and to avoid namespace collisions)

test/test_aqua.jl
```diff
@@ -1,4 +1,9 @@
 using DynamicQuantities
 using Aqua
 
-Aqua.test_all(DynamicQuantities; project_toml_formatting=false)
+Aqua.test_all(
+    DynamicQuantities;
+    project_toml_formatting=false,
+    # Requires is only loaded on older Julia versions that do not support extensions
+    stale_deps=(ignore=isdefined(Base, :get_extension) ? [:Requires] : Symbol[],),
```
MilesCranmer, line 8: Just wondering if this is still needed with the `@require_extensions` part?

MilesCranmer (changes requested):
Looks good! Here are some final suggestions. Let me know what you think.

Project.toml
```diff
@@ -41,22 +38,9 @@ Ratios = "c84ed2f1-dad5-54f0-aa8e-dbefe2724439"
...
-    "Ratios",
-    "SaferIntegers",
-    "SafeTestsets",
-    "ScientificTypes",
-    "ScientificTypesBase",
-    "StaticArrays",
-    "Test",
-    "Unitful"
-]
+test = ["Aqua", "LinearAlgebra", "Measurements", "Ratios", "SaferIntegers", "SafeTestsets", "ScientificTypes", "StaticArrays", "Test", "Unitful"]
```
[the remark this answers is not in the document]
devmotion: Yes, Pkg changes this automatically. I didn't perform any manual changes in this section, AFAICT running e.g. `julia> ] add PackageExtensionCompat` or `julia> ] rm TOML` modifies it in this way.

src/DynamicQuantities.jl
devmotion: Sure, I like that as well. I'd suggest
```suggestion
using PackageExtensionCompat: @require_extensions
```
though since the macro is only used but not extended in the package.

test/test_aqua.jl
devmotion: No, it's not. I thought I had removed it but I guess I messed up the git rebase.

src/DynamicQuantities.jl
MilesCranmer: > though since the macro is only used but not extended in the package.

Oh, I never knew this. Is this a standard strategy (`using` vs `import`) for indicating intended usage of a symbol?

I've been using BlueStyle for most of my packages which enforces the usage of `import` everywhere, which is why I stick to it in my packages (including internal imports; see `symbolic_dimensions.jl`). So I guess even if it doesn't match the semantic meaning I'd prefer to be consistently against style guidelines rather than inconsistently correct. (If that's okay with you?). But maybe I'll just switch everything over some time...

Project.toml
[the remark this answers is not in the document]
MilesCranmer: Got it! Okay we can leave it as-is then since I wouldn't want to manually reformat it each time.

MilesCranmer (approved):

src/DynamicQuantities.jl
MilesCranmer: Oh, weird. Not sure why I thought it did.

> So `using` is a bit safer IMO since it prevents you from unintentionally extending functions.

This is a **really** good point! Okay maybe we can try to switch them all (at once) in another PR. (And now I need to also do that in all my other libraries...)

devmotion: Reference: https://docs.julialang.org/en/v1/manual/modules/#using-and-import-with-specific-identifiers,-and-adding-methods

As an additional comment, if you would like to only load the module itself you can use `using MyPackage: MyPackage`.

BlueStyle actually explicitly notes that `using` should be prefered over `import`, even when extending functions: https://github.com/invenia/BlueStyle#module-imports

MilesCranmer:
Thanks for the PR!

PR #56, opened 2023-10-10
Remove SparseArrays dependency

### LINKED ISSUE #53: Make SparseArrays a weak dependency?

devmotion:
This PR removes the SparseArrays dependency completely.

Only the implementation of `==` and `+`/`-` required a bit more code, otherwise not too many changes were needed.

Fixes #53.

### DIFF
--- a/src/symbolic_dimensions.jl
+++ b/src/symbolic_dimensions.jl
@@ -213,7 +213,7 @@ function map_dimensions(op::O, l::SymbolicDimensions{L}, r::SymbolicDimensions{R
     while ir <= nr
         s = op(zero_L, nzvals_r[ir])
         if !iszero(s)
-            push!(I, op(nzdims_r[ir]))
+            push!(I, nzdims_r[ir])
             push!(V, s)
         end
         ir += 1
--- a/src/symbolic_dimensions.jl
+++ b/src/symbolic_dimensions.jl
@@ -82,7 +82,7 @@ function Base.convert(::Type{Quantity{T,SymbolicDimensions{R}}}, q::Quantity{<:A
     return Quantity(convert(T, ustrip(q)), dims)
 end
 function Base.convert(::Type{Q}, q::Quantity{<:Any,<:SymbolicDimensions}) where {T,D<:Dimensions,Q<:Quantity{T,D}}
-    result = one(Q) * ustrip(q)
+    result = new_quantity(Q, ustrip(q), dim_type(Q))
     d = dimension(q)
     for (idx, value) in zip(getfield(d, :nzdims), getfield(d, :nzvals))
         if !iszero(value)
--- a/src/symbolic_dimensions.jl
+++ b/src/symbolic_dimensions.jl
@@ -71,27 +71,23 @@ end
 function Base.convert(::Type{Quantity{T,SymbolicDimensions}}, q::Quantity{<:Any,<:Dimensions}) where {T}
     return convert(Quantity{T,SymbolicDimensions{DEFAULT_DIM_BASE_TYPE}}, q)
 end
-function Base.convert(::Type{Q}, q::Quantity{<:Any,<:Dimensions}) where {T,R,Q<:Quantity{T,SymbolicDimensions{R}}}
-    return new_quantity(
-        Q,
-        convert(T, ustrip(q)),
-        SymbolicDimensions{R}(;
-            m=ulength(q),
-            kg=umass(q),
-            s=utime(q),
-            A=ucurrent(q),
-            K=utemperature(q),
-            cd=uluminosity(q),
-            mol=uamount(q),
-        )
-    )
+function Base.convert(::Type{Quantity{T,SymbolicDimensions{R}}}, q::Quantity{<:Any,<:Dimensions}) where {T,R}
+    syms = (:m, :kg, :s, :A, :K, :cd, :mol)
+    vals = (ulength(q), umass(q), utime(q), ucurrent(q), utemperature(q), uluminosity(q), uamount(q))
+    I = INDEX_TYPE[ALL_MAPPING[s] for (s, v) in zip(syms, vals) if !iszero(v)]
+    V = R[tryrationalize(R, v) for v in vals if !iszero(v)]
+    p = sortperm(I)
+    permute!(I, p)
+    permute!(V, p)
+    dims = SymbolicDimensions{R}(I, V)
+    return Quantity(convert(T, ustrip(q)), dims)
 end
-function Base.convert(::Type{Q}, q::Quantity{<:Any,<:SymbolicDimensions}) where {T,D<:Dimensions,Q<:Quantity{T,D}}
-    result = new_quantity(Q, ustrip(q), dim_type(Q))
+function Base.convert(::Type{Quantity{T,D}}, q::Quantity{<:Any,<:SymbolicDimensions}) where {T,D<:Dimensions}
+    result = Quantity(T(ustrip(q)), D())
     d = dimension(q)
     for (idx, value) in zip(getfield(d, :nzdims), getfield(d, :nzvals))
         if !iszero(value)
-            result = result * convert(Q, ALL_VALUES[idx]) ^ value
+            result = result * convert(Quantity{T,D}, ALL_VALUES[idx]) ^ value
         end
     end
     return result
--- a/src/symbolic_dimensions.jl
+++ b/src/symbolic_dimensions.jl
@@ -152,9 +152,16 @@ function Base.:(==)(l::SymbolicDimensions, r::SymbolicDimensions)
     return true
 end
 Base.iszero(d::SymbolicDimensions) = iszero(getfield(d, :nzvals))
-Base.:*(l::SymbolicDimensions, r::SymbolicDimensions) = _combine_vals(+, l, r)
-Base.:/(l::SymbolicDimensions, r::SymbolicDimensions) = _combine_vals(-, l, r)
-function _combine_vals(op::O, l::SymbolicDimensions{L}, r::SymbolicDimensions{R}) where {O,L,R}
+
+# Defines `inv(::SymbolicDimensions)` and `^(::SymbolicDimensions, ::Number)`
+function map_dimensions(op::Function, d::SymbolicDimensions)
+    return SymbolicDimensions(getfield(d, :nzdims), map(op, getfield(d, :nzvals)))
+end
+
+# Defines `*(::SymbolicDimensions, ::SymbolicDimensions)` and `/(::SymbolicDimensions, ::SymbolicDimensions)`
+function map_dimensions(op::O, l::SymbolicDimensions{L}, r::SymbolicDimensions{R}) where {O<:Function,L,R}
+    zero_L = zero(L)
+    zero_R = zero(R)
     T = typeof(op(zero(L), zero(R)))
     I = Vector{INDEX_TYPE}(undef, 0)
     V = Vector{T}(undef, 0)
@@ -177,14 +184,14 @@ function _combine_vals(op::O, l::SymbolicDimensions{L}, r::SymbolicDimensions{R}
             il += 1
             ir += 1
         elseif dim_l < dim_r
-            s = nzvals_l[il]
+            s = op(nzvals_l[il], zero_R)
             if !iszero(s)
                 push!(I, dim_l)
                 push!(V, s)
             end
             il += 1
         else
-            s = op(nzvals_r[ir])
+            s = op(zero_L, nzvals_r[ir])
             if !iszero(s)
                 push!(I, dim_r)
                 push!(V, s)
@@ -194,7 +201,7 @@ function _combine_vals(op::O, l::SymbolicDimensions{L}, r::SymbolicDimensions{R}
     end
 
     while il <= nl
-        s = nzvals_l[il]
+        s = op(nzvals_l[il], zero_R)
         if !iszero(s)
             push!(I, nzdims_l[il])
             push!(V, s)
@@ -203,7 +210,7 @@ function _combine_vals(op::O, l::SymbolicDimensions{L}, r::SymbolicDimensions{R}
     end
 
     while ir <= nr
-        s = nzvals_r[ir]
+        s = op(zero_L, nzvals_r[ir])
         if !iszero(s)
             push!(I, op(nzdims_r[ir]))
             push!(V, s)
@@ -213,10 +220,6 @@ function _combine_vals(op::O, l::SymbolicDimensions{L}, r::SymbolicDimensions{R}
 
     return SymbolicDimensions(I, V)
 end
-Base.inv(d::SymbolicDimensions) = SymbolicDimensions(getfield(d, :nzdims), -getfield(d, :nzvals))
-Base.:^(l::SymbolicDimensions{R}, r::Integer) where {R} = SymbolicDimensions(getfield(l, :nzdims), r * getfield(l, :nzvals))
-Base.:^(l::SymbolicDimensions{R}, r::Number) where {R} = SymbolicDimensions(getfield(l, :nzdims), tryrationalize(R, r) * getfield(l, :nzvals))
-
 
 """
     SymbolicUnitsParse
--- a/src/symbolic_dimensions.jl
+++ b/src/symbolic_dimensions.jl
@@ -3,6 +3,7 @@ import .Constants: CONSTANT_SYMBOLS, CONSTANT_MAPPING, CONSTANT_VALUES
 
 const SYMBOL_CONFLICTS = intersect(UNIT_SYMBOLS, CONSTANT_SYMBOLS)
 
+const INDEX_TYPE = UInt8
 # Prefer units over constants:
 # For example, this means we can't have a symbolic Planck's constant,
 # as it is just "hours" (h), which is more common.
@@ -18,7 +19,7 @@ const ALL_VALUES = vcat(
         if k ∉ SYMBOL_CONFLICTS
     )...
 )
-const ALL_MAPPING = NamedTuple([s => UInt8(i) for (i, s) in enumerate(ALL_SYMBOLS)])
+const ALL_MAPPING = NamedTuple([s => INDEX_TYPE(i) for (i, s) in enumerate(ALL_SYMBOLS)])
 
 """
     SymbolicDimensions{R} <: AbstractDimensions{R}
@@ -34,7 +35,7 @@ to one which uses `Dimensions` as its dimensions (i.e., base SI units)
 `expand_units`.
 """
 struct SymbolicDimensions{R} <: AbstractDimensions{R}
-    nzdims::Vector{UInt8}
+    nzdims::Vector{INDEX_TYPE}
     nzvals::Vector{R}
 end
 
@@ -43,7 +44,11 @@ function Base.getproperty(d::SymbolicDimensions{R}, s::Symbol) where {R}
     nzdims = getfield(d, :nzdims)
     i = ALL_MAPPING[s]
     ii = searchsortedfirst(nzdims, i)
-    (ii <= length(nzdims) && nzdims[ii] == i) ? getfield(d, :nzvals)[ii] : zero(R)
+    if ii <= length(nzdims) && nzdims[ii] == i
+        return getfield(d, :nzvals)[ii]
+    else
+        return zero(R)
+    end
 end
 Base.propertynames(::SymbolicDimensions) = ALL_SYMBOLS
 Base.getindex(d::SymbolicDimensions, k::Symbol) = getproperty(d, k)
@@ -53,9 +58,9 @@ SymbolicDimensions{R}(d::SymbolicDimensions) where {R} = SymbolicDimensions{R}(g
 SymbolicDimensions(; kws...) = SymbolicDimensions{DEFAULT_DIM_BASE_TYPE}(; kws...)
 function SymbolicDimensions{R}(; kws...) where {R}
     if isempty(kws)
-        return SymbolicDimensions{R}(Vector{UInt8}(undef, 0), Vector{R}(undef, 0))
+        return SymbolicDimensions{R}(Vector{INDEX_TYPE}(undef, 0), Vector{R}(undef, 0))
     end
-    I = UInt8[ALL_MAPPING[s] for s in keys(kws)]
+    I = INDEX_TYPE[ALL_MAPPING[s] for s in keys(kws)]
     p = sortperm(I)
     V = R[tryrationalize(R, kws[i]) for i in p]
     return SymbolicDimensions{R}(permute!(I, p), V)
@@ -68,7 +73,7 @@ end
 function Base.convert(::Type{Quantity{T,SymbolicDimensions{R}}}, q::Quantity{<:Any,<:Dimensions}) where {T,R}
     syms = (:m, :kg, :s, :A, :K, :cd, :mol)
     vals = (ulength(q), umass(q), utime(q), ucurrent(q), utemperature(q), uluminosity(q), uamount(q))
-    I = UInt8[ALL_MAPPING[s] for (s, v) in zip(syms, vals) if !iszero(v)]
+    I = INDEX_TYPE[ALL_MAPPING[s] for (s, v) in zip(syms, vals) if !iszero(v)]
     p = sortperm(I)
     permute!(I, p)
     V = R[tryrationalize(R, vals[i]) for i in p]
@@ -151,7 +156,7 @@ Base.:*(l::SymbolicDimensions, r::SymbolicDimensions) = _combine_vals(+, l, r)
 Base.:/(l::SymbolicDimensions, r::SymbolicDimensions) = _combine_vals(-, l, r)
 function _combine_vals(op::O, l::SymbolicDimensions{L}, r::SymbolicDimensions{R}) where {O,L,R}
     T = typeof(op(zero(L), zero(R)))
-    I = Vector{UInt8}(undef, 0)
+    I = Vector{INDEX_TYPE}(undef, 0)
     V = Vector{T}(undef, 0)
     nzdimsl = getfield(l, :nzdims)
     nzvalsl = getfield(l, :nzvals)
--- a/src/symbolic_dimensions.jl
+++ b/src/symbolic_dimensions.jl
@@ -1,6 +1,5 @@
 import .Units: UNIT_SYMBOLS, UNIT_MAPPING, UNIT_VALUES
 import .Constants: CONSTANT_SYMBOLS, CONSTANT_MAPPING, CONSTANT_VALUES
-import SparseArrays as SA
 
 const SYMBOL_CONFLICTS = intersect(UNIT_SYMBOLS, CONSTANT_SYMBOLS)
 
@@ -19,7 +18,7 @@ const ALL_VALUES = vcat(
         if k ∉ SYMBOL_CONFLICTS
     )...
 )
-const ALL_MAPPING = NamedTuple([s => i for (i, s) in enumerate(ALL_SYMBOLS)])
+const ALL_MAPPING = NamedTuple([s => UInt8(i) for (i, s) in enumerate(ALL_SYMBOLS)])
 
 """
     SymbolicDimensions{R} <: AbstractDimensions{R}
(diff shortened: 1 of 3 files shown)


src/symbolic_dimensions.jl
```diff
@@ -19,7 +18,7 @@ const ALL_VALUES = vcat(
         if k ∉ SYMBOL_CONFLICTS
     )...
 )
-const ALL_MAPPING = NamedTuple([s => i for (i, s) in enumerate(ALL_SYMBOLS)])
+const ALL_MAPPING = NamedTuple([s => UInt8(i) for (i, s) in enumerate(ALL_SYMBOLS)])
```
devmotion, line 21: Not sure if this storage optimization (here and in the `SymbolicDimensions` struct) is worth it.

src/symbolic_dimensions.jl
```diff
@@ -35,49 +34,54 @@ to one which uses `Dimensions` as its dimensions (i.e., base SI units)
...
+function SymbolicDimensions{R}(; kws...) where {R}
+    if isempty(kws)
+        return SymbolicDimensions{R}(Vector{UInt8}(undef, 0), Vector{R}(undef, 0))
     end
+    I = UInt8[ALL_MAPPING[s] for s in keys(kws)]
+    p = sortperm(I)
+    V = R[tryrationalize(R, kws[i]) for i in p]
+    return SymbolicDimensions{R}(permute!(I, p), V)
+end
+(::Type{<:SymbolicDimensions})(::Type{R}; kws...) where {R} = SymbolicDimensions{R}(; kws...)
```
devmotion, line 63: I added this for backward compatibility but IMO it should be removed - the standard way to specify the type `R` is to call `SymbolicDimensions{R}` directly. With this function signature, unintuitive and confusing calls such as `SymbolicDimensions{Float32}(Float64)` are possible.

src/symbolic_dimensions.jl
MilesCranmer: I think it's fine to use this. It also helps readability because it indicates there we only expect 256 max indices here (which is true).

I guess you could do UInt16 just in case the user wants to add a bunch of symbols of their own, so they don't run into any weird overflow bugs. But it's not a big deal if you'd prefer UInt8.

src/symbolic_dimensions.jl
devmotion: > I don't actually remember if SymbolicDimensions{T}(T2; ...) is used anywhere, I don't think it is?

It's supported here:
https://github.com/SymbolicML/DynamicQuantities.jl/blob/87e1eb3d0d4c139ed94a563e2fae3cababc5e512/src/symbolic_dimensions.jl#L52

Similar definitions exist for other types as well:
https://github.com/SymbolicML/DynamicQuantities.jl/blob/87e1eb3d0d4c139ed94a563e2fae3cababc5e512/src/types.jl#L75
https://github.com/SymbolicML/DynamicQuantities.jl/blob/87e1eb3d0d4c139ed94a563e2fae3cababc5e512/src/types.jl#L118-L119

They all allow such inconsistent calls. IMO all these `(::Type{...})` constructors should be removed and replaced with standard constructors such as e.g. `Dimension(...)` and `Dimension{R}(...)`.

MilesCranmer: > They all allow such inconsistent calls.

The `constructor_of` part removes the type, so I think it's safe, right? It would never actually hit `SymbolicDimensions{R}(R2, ...)` unless you explicitly call that (in which case it would return a `SymbolicDimensions{R2}`.

But yeah I do agree this feels a bit dangerous to play with such generic constructors.

> IMO all these `(::Type{...})` constructors should be removed and replaced with standard constructors such as e.g. `Dimension(...)` and `Dimension{R}(...)`.

I set it up this way so you could easily create your own custom dimensions (as a struct-based alternative to the macro-based approach of Unitful):

```julia
struct MyDimensions{R} <:AbstractDimensions{R}
    length::R
    mass::R
end
```

and just start using it, like:

```julia
x1 = Quantity(0.5, MyDimensions, length=2)
x2 = Quantity(0.3, MyDimensions, mass=-1)

x1 * x2 # 0.5*0.3 length=2 mass=-1
```

I could see many applications where one simply never uses some of the physical dimensions, so I wanted to make it very easy to declare custom dimensional spaces.

But I'm open to opinions on this!

devmotion: > It would never actually hit SymbolicDimensions{R}(R2, ...) unless you explicitly call that (in which case it would return a SymbolicDimensions{R2}.

Yes, the implementation is safe. My point was that it's surprising to allow this call at all and that the return type is unexpected: By convention, you would expect that a call of `SomeType{R}(...)` creates an instance of type `SomeType{R}`.

MilesCranmer: Ah, I see what you mean. I agree we should change that. 

I think the most logical behavior would be:

1. Calling `Dimension` and passing a type `R` => Creates `Dimension{R}`.
2. Calling `Dimension{R}` and passing a type `R` => Creates `Dimension{R}`.
2. Calling `Dimension{R}` explicitly but passing a different type `R2` => Error.

Then for quantities:

1. Calling `Quantity` and passing a value of type `T` => Creates `Quantity{T,D}` (default `D`).
2. Calling `Quantity{T}` and passing a value of type `T2` => Creates `Quantity{T,D}` (no error) (default `D`).
2. Calling `Quantity{T}` and passing a value of type `T2` and specifying a dimension type `D` => Creates `Quantity{T,D}`.
3. Calling `Quantity{T,D}` and passing a value of type `T2` => Creates `Quantity{T,D}`.
3. Calling `Quantity{T,D}` and passing a value of type `T2` and specifying dimension type `D` => Creates `Quantity{T,D}`.
4. Calling `Quantity{T,D}` and passing a value of any type, but specifying a different dimension type `D2` => Error.

Wdyt?

MilesCranmer:
I'm wondering (maybe I missed it) if there could be some `filter!` or `deleteat!` somewhere? Otherwise it seems like the vector of dimensions could keep growing.

e.g., each operation on SymbolicDimensions could end with something like

```julia
remove_zeros(v) = filter!(e->!iszero(e), v)
```

(but adapted to the two vector version).

I'm not sure where in the operations SparseArrays performs this.

src/symbolic_dimensions.jl
devmotion: I think even better than erroring out in some cases would be to not allow such calls in the first place. For instance, for `Dimension` I would just define
```julia
Dimension(; kw...) = Dimension{DEFAULT_BASE_TYPE}(; kw...)
Dimension{R}(; length=zero(R), time=zero(R), ....) where {R} = Dimension{R}(R(length), R(time), ...)
```
without any specific `Type` argument.

Also for `Quantity` `Type` arguments are not needed: One could define
```julia
Quantity(val; kw...) = Quantity{typeof(val),DEFAULT_DIMENSION_TYPE}(val; kw...)
Quantity{T}(val; kw...) where {T} = Quantity{T,DEFAULT_DIMENSION_TYPE}(val; kw...)
Quantity{T,D}(val; kw...) where {T,D<:Dimension} = Quantity(T(val), D(; kw...))
```

If it's more desirable to specify the dimension type, then one could change the order of the type parameters.

For user types, the automatic keyword-based functionality (i.e., the `Quantity{T,<:CustomType}(...)` definitions) could be defined with a weak dependency on ConstructionBase.jl: https://juliaobjects.github.io/ConstructionBase.jl/dev/ If users want to use this convenience feature for their custom dimension types, they could extend `ConstructionBase.constructorof`. In the extension one would define
```julia
function DynamicQuantities.Quantity{T,D}(val; kw...) where {T,D<:AbstractDimension}
    return Quantity(T(val), ConstructionBase.constructorof(D)(; kw...))
end
```

src/symbolic_dimensions.jl
MilesCranmer, line 21:
```suggestion
const ALL_MAPPING = NamedTuple([s => INDEX_TYPE(i) for (i, s) in enumerate(ALL_SYMBOLS)])
```

src/symbolic_dimensions.jl
```diff
@@ -1,6 +1,5 @@
 import .Units: UNIT_SYMBOLS, UNIT_MAPPING, UNIT_VALUES
 import .Constants: CONSTANT_SYMBOLS, CONSTANT_MAPPING, CONSTANT_VALUES
-import SparseArrays as SA
 
 const SYMBOL_CONFLICTS = intersect(UNIT_SYMBOLS, CONSTANT_SYMBOLS)
```
MilesCranmer, line 5:
```suggestion
const INDEX_TYPE = UInt16
```

So we only need to set it in one location.

UInt16 over UInt8 to give more room for when more symbols are added in the future. 255 seems too little and could cause overflow issues if someone is not aware and adds their own set of symbols.

src/symbolic_dimensions.jl
```diff
@@ -35,49 +34,54 @@ to one which uses `Dimensions` as its dimensions (i.e., base SI units)
 `expand_units`.
 """
 struct SymbolicDimensions{R} <: AbstractDimensions{R}
-    _data::SA.SparseVector{R}
-
-    SymbolicDimensions(data::SA.SparseVector) = new{eltype(data)}(data)
-    SymbolicDimensions{_R}(data::SA.SparseVector) where {_R} = new{_R}(data)
+    nzdims::Vector{UInt8}
```
MilesCranmer, line 37:
```suggestion
    nzdims::Vector{INDEX_TYPE}
```

src/symbolic_dimensions.jl
```diff
@@ -35,49 +34,54 @@ to one which uses `Dimensions` as its dimensions (i.e., base SI units)
...
-        length(kws) == 0 && return constructor(SA.spzeros(R, length(ALL_SYMBOLS)))
-        I = [ALL_MAPPING[s] for s in keys(kws)]
-        V = [tryrationalize(R, v) for v in values(kws)]
-        data = SA.sparsevec(I, V, length(ALL_SYMBOLS))
-        return constructor(data)
+SymbolicDimensions{R}(d::SymbolicDimensions) where {R} = SymbolicDimensions{R}(getfield(d, :nzdims), convert(Vector{R}, getfield(d, :nzvals)))
+SymbolicDimensions(; kws...) = SymbolicDimensions{DEFAULT_DIM_BASE_TYPE}(; kws...)
+function SymbolicDimensions{R}(; kws...) where {R}
+    if isempty(kws)
+        return SymbolicDimensions{R}(Vector{UInt8}(undef, 0), Vector{R}(undef, 0))
```
MilesCranmer, line 56:
```suggestion
        return SymbolicDimensions{R}(Vector{INDEX_TYPE}(undef, 0), Vector{R}(undef, 0))
```

src/symbolic_dimensions.jl
```diff
@@ -35,49 +34,54 @@ to one which uses `Dimensions` as its dimensions (i.e., base SI units)
...
     end
+    I = UInt8[ALL_MAPPING[s] for s in keys(kws)]
```
MilesCranmer, line 58:
```suggestion
    I = INDEX_TYPE[ALL_MAPPING[s] for s in keys(kws)]
```

src/symbolic_dimensions.jl
```diff
@@ -35,49 +34,54 @@ to one which uses `Dimensions` as its dimensions (i.e., base SI units)
...
-    )
-    SA.dropzeros!(data(dimension(output)))
-    return output
+function Base.convert(::Type{Quantity{T,SymbolicDimensions}}, q::Quantity{<:Any,<:Dimensions}) where {T}
+    return convert(Quantity{T,SymbolicDimensions{DEFAULT_DIM_BASE_TYPE}}, q)
+end
+function Base.convert(::Type{Quantity{T,SymbolicDimensions{R}}}, q::Quantity{<:Any,<:Dimensions}) where {T,R}
+    syms = (:m, :kg, :s, :A, :K, :cd, :mol)
+    vals = (ulength(q), umass(q), utime(q), ucurrent(q), utemperature(q), uluminosity(q), uamount(q))
+    I = UInt8[ALL_MAPPING[s] for (s, v) in zip(syms, vals) if !iszero(v)]
```
MilesCranmer, line 71:
```suggestion
    I = INDEX_TYPE[ALL_MAPPING[s] for (s, v) in zip(syms, vals) if !iszero(v)]
```

src/symbolic_dimensions.jl
```diff
@@ -95,14 +99,118 @@ end
...
+    end
+
+    return true
+end
+Base.iszero(d::SymbolicDimensions) = iszero(getfield(d, :nzvals))
+Base.:*(l::SymbolicDimensions, r::SymbolicDimensions) = _combine_vals(+, l, r)
+Base.:/(l::SymbolicDimensions, r::SymbolicDimensions) = _combine_vals(-, l, r)
+function _combine_vals(op::O, l::SymbolicDimensions{L}, r::SymbolicDimensions{R}) where {O,L,R}
+    T = typeof(op(zero(L), zero(R)))
+    I = Vector{UInt8}(undef, 0)
```
MilesCranmer, line 154:
```suggestion
    I = Vector{INDEX_TYPE}(undef, 0)
```

MilesCranmer (commented):
Initial suggestions to change UInt8 to UInt16, and also make that type a global constant for ensuring consistency.

src/symbolic_dimensions.jl
```diff
@@ -35,49 +34,54 @@ to one which uses `Dimensions` as its dimensions (i.e., base SI units)
 `expand_units`.
 """
 struct SymbolicDimensions{R} <: AbstractDimensions{R}
-    _data::SA.SparseVector{R}
-
-    SymbolicDimensions(data::SA.SparseVector) = new{eltype(data)}(data)
-    SymbolicDimensions{_R}(data::SA.SparseVector) where {_R} = new{_R}(data)
+    nzdims::Vector{UInt8}
+    nzvals::Vector{R}
```
MilesCranmer, line 38: Could these be `_nzdims` and `_nzvals` to indicate they should not be accessed by the user? Since all other `AbstractDimensions` objects typically have each field corresponding to a specific physical dimension.

src/symbolic_dimensions.jl
```diff
@@ -35,49 +34,54 @@ to one which uses `Dimensions` as its dimensions (i.e., base SI units)
...
 
 static_fieldnames(::Type{<:SymbolicDimensions}) = ALL_SYMBOLS
-data(d::SymbolicDimensions) = getfield(d, :_data)
-Base.getproperty(d::SymbolicDimensions{R}, s::Symbol) where {R} = data(d)[ALL_MAPPING[s]]
-Base.getindex(d::SymbolicDimensions{R}, k::Symbol) where {R} = getproperty(d, k)
+function Base.getproperty(d::SymbolicDimensions{R}, s::Symbol) where {R}
+    nzdims = getfield(d, :nzdims)
+    i = ALL_MAPPING[s]
+    ii = searchsortedfirst(nzdims, i)
+    (ii <= length(nzdims) && nzdims[ii] == i) ? getfield(d, :nzvals)[ii] : zero(R)
```
MilesCranmer, line 46:
```suggestion
    if ii <= length(nzdims) && nzdims[ii] == i
        return getfield(d, :nzvals)[ii]
    else
        return zero(R)
    end
```

src/symbolic_dimensions.jl
```diff
@@ -35,49 +34,54 @@ to one which uses `Dimensions` as its dimensions (i.e., base SI units)
...
+SymbolicDimensions{R}(d::SymbolicDimensions) where {R} = SymbolicDimensions{R}(getfield(d, :nzdims), convert(Vector{R}, getfield(d, :nzvals)))
+SymbolicDimensions(; kws...) = SymbolicDimensions{DEFAULT_DIM_BASE_TYPE}(; kws...)
+function SymbolicDimensions{R}(; kws...) where {R}
+    if isempty(kws)
+        return SymbolicDimensions{R}(Vector{UInt8}(undef, 0), Vector{R}(undef, 0))
     end
+    I = UInt8[ALL_MAPPING[s] for s in keys(kws)]
+    p = sortperm(I)
+    V = R[tryrationalize(R, kws[i]) for i in p]
+    return SymbolicDimensions{R}(permute!(I, p), V)
```
MilesCranmer, line 61:
```suggestion
    return SymbolicDimensions{R}(I[p], V)
```

(The number of keywords is typically ~1-3 elements so `permute!` is a bit overkill)

src/symbolic_dimensions.jl
```diff
@@ -35,49 +34,54 @@ to one which uses `Dimensions` as its dimensions (i.e., base SI units)
...
+    return convert(Quantity{T,SymbolicDimensions{DEFAULT_DIM_BASE_TYPE}}, q)
+end
+function Base.convert(::Type{Quantity{T,SymbolicDimensions{R}}}, q::Quantity{<:Any,<:Dimensions}) where {T,R}
+    syms = (:m, :kg, :s, :A, :K, :cd, :mol)
+    vals = (ulength(q), umass(q), utime(q), ucurrent(q), utemperature(q), uluminosity(q), uamount(q))
+    I = UInt8[ALL_MAPPING[s] for (s, v) in zip(syms, vals) if !iszero(v)]
+    p = sortperm(I)
+    permute!(I, p)
+    V = R[tryrationalize(R, vals[i]) for i in p]
+    dims = SymbolicDimensions{R}(I, V)
```
MilesCranmer, line 75:
```suggestion
    p = sortperm(I)
    V = R[tryrationalize(R, vals[i]) for i in p]
    dims = SymbolicDimensions{R}(I[p], V)
```

src/symbolic_dimensions.jl
```diff
@@ -95,14 +99,118 @@ end
...
-Base.:(==)(l::SymbolicDimensions, r::SymbolicDimensions) = data(l) == data(r)
-Base.iszero(d::SymbolicDimensions) = iszero(data(d))
-Base.:*(l::SymbolicDimensions, r::SymbolicDimensions) = SymbolicDimensions(data(l) + data(r))
-Base.:/(l::SymbolicDimensions, r::SymbolicDimensions) = SymbolicDimensions(data(l) - data(r))
-Base.inv(d::SymbolicDimensions) = SymbolicDimensions(-data(d))
-Base.:^(l::SymbolicDimensions{R}, r::Integer) where {R} = SymbolicDimensions(data(l) * r)
-Base.:^(l::SymbolicDimensions{R}, r::Number) where {R} = SymbolicDimensions(data(l) * tryrationalize(R, r))
+Base.copy(d::SymbolicDimensions) = SymbolicDimensions(copy(getfield(d, :nzdims)), copy(getfield(d, :nzvals)))
+function Base.:(==)(l::SymbolicDimensions, r::SymbolicDimensions)
+    nzdimsl = getfield(l, :nzdims)
```
MilesCranmer, line 104: Nitpick: could these (and other variables) be `.._l` rather than just `..l`, and `.._r` rather than `..r`? Just to improve readability.

src/symbolic_dimensions.jl
```diff
@@ -95,14 +99,118 @@ end
...
+    while il <= nl && ir <= nr
+        diml = nzdimsl[il]
+        dimr = nzdimsr[ir]
+        if diml == dimr
+            if nzvalsl[il] != nzvalsr[ir]
+                return false
+            end
+            il += 1
+            ir += 1
+        elseif diml < dimr
```
MilesCranmer, line 120: Not sure I understand. What do these branches for `diml < dimr` and `diml > dimr` do?

MilesCranmer: Oh wait, I think I get it.

Wouldn't it be simpler (both conceptually and perhaps computationally) to first pop any 0s from each vector, and then we only need to check `nzdimsl == nzdimsr` and `nzvalsl == nzvalsr`?

src/symbolic_dimensions.jl
```diff
@@ -95,14 +99,118 @@ end
...
+        end
+        ir += 1
+    end
+
+    return true
+end
+Base.iszero(d::SymbolicDimensions) = iszero(getfield(d, :nzvals))
+Base.:*(l::SymbolicDimensions, r::SymbolicDimensions) = _combine_vals(+, l, r)
+Base.:/(l::SymbolicDimensions, r::SymbolicDimensions) = _combine_vals(-, l, r)
+function _combine_vals(op::O, l::SymbolicDimensions{L}, r::SymbolicDimensions{R}) where {O,L,R}
```
MilesCranmer, line 152: I think it might be better to use multiple dispatch on `map_dimensions` here instead of defining a new function name. Then some operators come for free like `*` and `/`.

MilesCranmer: Also the single-operator `map_dimensions` should be defined as well, which would give us `inv` for free.

src/symbolic_dimensions.jl
MilesCranmer: Actually wait I'm having second thoughts. Maybe `UInt8` is good! We're only at 113 symbols right now.

That being said it would still be nice to have this set in a global constant.

devmotion: Yes, I was considering adding a global constant as well, I think that'll make it easier to change the type if needed.

```suggestion

const INDEX_TYPE = UInt8
```

src/symbolic_dimensions.jl
devmotion: I was considering this as well 😅 It didn't seem necessary to me though since you can't access them accidentally anyway: You have to use `getfield(dims, :nzvals)` etc. since `dims.nzvals` will throw an error (there's no symbolic dimension `:nzvals`). So using `_nzvals` seemed mainly to increase the number of characters in the implementation.

What do you think?

src/symbolic_dimensions.jl
devmotion: Surely a micro-optimization but still saves a few allocations 🤷 Note that the computer will also allocate more memory than needed (so not only for 1-3 elements) to avoid that new memory has to be allocated for every `push!`. (With the implementation on the master branch, `I` always contains 7 elements when converting `Dimensions` to `SymbolicDimensions` but with this PR zero dimensions will be omitted.)

src/symbolic_dimensions.jl
MilesCranmer: Do you know if hitting the tab key on `dim.` in the REPL will print out the properties, or the fields? (Sent from mobile so I can’t check right now)

src/symbolic_dimensions.jl
devmotion: Popping elements will cause unnecessary allocations (if I understood your suggestion correctly). This should be the most efficient algorithm since it only loops over the dimensions once without allocating anything and with exiting early. The same algorithm is also used in SparseArrays: https://github.com/JuliaSparse/SparseArrays.jl/blob/d884072ee95d522505c4b27fa928dc3c251f7d2b/src/sparsevector.jl#L937-L966

src/symbolic_dimensions.jl
devmotion: I just defined whatever is currently defined on the master branch. I guess additional improvements might be better left for a separate PR?

src/symbolic_dimensions.jl
devmotion: Done!

src/symbolic_dimensions.jl
MilesCranmer: Oh, sorry, what I’m trying to say is that we only need to change the function name `_combine_vals` to `map_dimensions`. Rather than defining a new function name. (This signature is the same one used for `map_dimensions`)

Then the lines for `*` and `/` can be deleted as well.

devmotion: Ah, I see 👍

src/symbolic_dimensions.jl
MilesCranmer: Okay, fine with me!

src/symbolic_dimensions.jl
MilesCranmer: I see, thanks!

src/symbolic_dimensions.jl
devmotion: Done. BTW what's the reason for a separate `map_dimensions` function instead of just overloading `map`?

src/symbolic_dimensions.jl
devmotion: It depends on whether you define `Base.propertynames` (as done in this PR) or not (as currently done on the master branch):

In this PR:
```julia

julia> dims = SymbolicDimensions(; m = 3, s = -1)
m³ s⁻¹

julia> dims.
A         C         F         G         GHz       GV        GW        GeV       Gm        Gpc       Gyr       Hz        J         K         L         L_bol0    L_sun     MHz
MV        MW        M_earth   M_jup     M_sun     MeV       Mm        Mpc       Myr       N         N_A       Pa        R         R_earth   R_jup     R_sun     Ryd       T
TeV       V         W         a_0       alpha     atm       au        bar       c         cd        cm        dL        day       dm        e         eV        eps_0     fm
fs        g         h         hbar      hr        kA        kHz       kJ        kPa       kV        kW        k_B       k_e       keV       kg        km        kpc       kyr
ly        m         mA        mK        mL        mV        m_e       m_n       m_p       mcd       meV       mg        min       mm        mmol      mohm      mol       ms
mu_0      mΩ        nA        nm        ns        ohm       pc        pm        ps        s         sigma_T   sigma_sb  u         uA        ug        um        us        yr
Ω         μA        μg        μm        μs
```

On the master branch:
```julia

julia> dims = SymbolicDimensions(; m = 3, s = -1)
m³ s⁻¹

julia> dims._data # (Autocompleted to _data after pressing tab)
```

src/symbolic_dimensions.jl
```diff
@@ -95,15 +104,122 @@ end
...
+        ir += 1
+    end
+
+    return true
+end
+Base.iszero(d::SymbolicDimensions) = iszero(getfield(d, :nzvals))
+
+# Defines `inv(::SymbolicDimensions)` and `^(::SymbolicDimensions, ::Number)`
+function map_dimensions(op::Function, d::SymbolicDimensions)
+    return SymbolicDimensions(getfield(d, :nzdims), map(op, getfield(d, :nzvals)))
```
MilesCranmer, line 158: Should `nzdims` be explicitly copied here? (Since it’s mutable storage for an otherwise immutable struct)

devmotion: Good catch. Yes, probably that's safer.

devmotion:
```suggestion
    return SymbolicDimensions(copy(getfield(d, :nzdims)), map(op, getfield(d, :nzvals)))
```

src/symbolic_dimensions.jl
MilesCranmer: Cool! Okay I don’t think there’s a need for this change then. Unless we want to indicate it’s a private field at the source code level.

src/symbolic_dimensions.jl
MilesCranmer: I think we could solve this in another PR, and just get this one out of the way first? Since this would be a breaking change anyways, we could get the SparseArrays removal into v0.7 first.

MilesCranmer (commented):
I think everything is looking good. The one thing remaining is adding some unittests to ensure coverage over all the modifications. In particular it looks like all the branches of `==` do not get tested by the current tests (green/red column to the left of code):

Screenshot 2023-10-12 at 13 36 08

which would be good to ensure we test extensively enough, since it is a custom sparse array type. Also, some of the utility functions are not currently tested:

Screenshot 2023-10-12 at 13 36 59

which would be good to add as well.

src/symbolic_dimensions.jl
devmotion: Yes, I agree. Such changes are not independent of the SparseArrays PR.

devmotion:
I added more tests that should improve coverage.

MilesCranmer:
One more TODO item (I think the last one, we're almost done) is to benchmark QuantityArray broadcasting for SymbolicDimensions. I am curious to see if the compiler inlining on the `==` changes at all with the custom sparse array implementation.

src/symbolic_dimensions.jl
```diff
@@ -95,15 +104,122 @@ end
...
+            push!(I, nzdims_l[il])
+            push!(V, s)
+        end
+        il += 1
+    end
+
+    while ir <= nr
+        s = op(zero_L, nzvals_r[ir])
+        if !iszero(s)
+            push!(I, op(nzdims_r[ir]))
```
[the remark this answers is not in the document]
MilesCranmer: (In the future I think we should add https://github.com/Seelengrab/PropCheck.jl to the test suite for catching these sorts of edge cases-within-edge cases.)

MilesCranmer:
Awesome. I have no idea how, but this PR makes the broadcasting on SymbolicDimensions ~18x as fast as it was before!

Before:

```julia
julia> f(v) = v^2 * 1.5;

julia> @btime $f.(qa) setup=(xa = randn(10000) .* us"km/s"; qa = QuantityArray(xa));
  17.795 ms (160019 allocations: 13.05 MiB)
```

after:

```julia
julia> @btime $f.(qa) setup=(xa = randn(10000) .* us"km/s"; qa = QuantityArray(xa));
  1.053 ms (40007 allocations: 2.06 MiB)
```

(Still much slower than with normal `Dimensions`, which has an advantage by being immutable, but I think this is pretty good)

---

Also, I went ahead and fixed the bug I noted [here](https://github.com/SymbolicML/DynamicQuantities.jl/pull/56#discussion_r1357103067).

MilesCranmer:
Hm, I found another bug:

```julia
> convert(Quantity{Float64,SymbolicDimensions}, u"kg")
1.0 
```

not sure why this happens.

src/symbolic_dimensions.jl
```diff
@@ -35,49 +35,59 @@ to one which uses `Dimensions` as its dimensions (i.e., base SI units)
...
-    )
-    SA.dropzeros!(data(dimension(output)))
-    return output
+function Base.convert(::Type{Quantity{T,SymbolicDimensions}}, q::Quantity{<:Any,<:Dimensions}) where {T}
+    return convert(Quantity{T,SymbolicDimensions{DEFAULT_DIM_BASE_TYPE}}, q)
+end
+function Base.convert(::Type{Quantity{T,SymbolicDimensions{R}}}, q::Quantity{<:Any,<:Dimensions}) where {T,R}
+    syms = (:m, :kg, :s, :A, :K, :cd, :mol)
+    vals = (ulength(q), umass(q), utime(q), ucurrent(q), utemperature(q), uluminosity(q), uamount(q))
+    I = INDEX_TYPE[ALL_MAPPING[s] for (s, v) in zip(syms, vals) if !iszero(v)]
```
MilesCranmer, line 77: I think this `if !iszero(v)` is what is introducing the second bug. I'll push a fix for this. For safety I think this `convert` should just call the normal constructor instead of writing out the `sortperm` stuff again.

devmotion: On purpose, I did not use the same constructor since this will lead to unnecessary zero dimensions. Using the same constructor will also change behaviour compared with the master branch.

So my suggestion is to revert the last commit (commits?) and instead fix and test the bug in the function.

MilesCranmer: Got it, thanks!

src/symbolic_dimensions.jl
```diff
@@ -95,15 +106,122 @@ end
...
+# Defines `inv(::SymbolicDimensions)` and `^(::SymbolicDimensions, ::Number)`
+function map_dimensions(op::Function, d::SymbolicDimensions)
+    return SymbolicDimensions(copy(getfield(d, :nzdims)), map(op, getfield(d, :nzvals)))
+end
 
+# Defines `*(::SymbolicDimensions, ::SymbolicDimensions)` and `/(::SymbolicDimensions, ::SymbolicDimensions)`
+function map_dimensions(op::O, l::SymbolicDimensions{L}, r::SymbolicDimensions{R}) where {O<:Function,L,R}
+    zero_L = zero(L)
+    zero_R = zero(R)
+    T = typeof(op(zero(L), zero(R)))
```
MilesCranmer, line 167: I'm a bit worried about type instabilities coming from this line if there is a failure by the compiler to propagate constants. e.g., for certain rational-like types they might have some complicated logic for `+` and the compiler could refuse to inline.

I don't see an immediate need for this, as `map_dimensions` is only used for `+` and `-` on `SymbolicDimensions` currently, so I propose to make this simply `promote_type(L, R)`. Any objections? `promote_type` is what is used for `map_dimensions` for `AbstractDimensions` anyways.

MilesCranmer: (Just pushed this change... let me know if you feel strongly against it)

devmotion: `promote_type` won't give the correct result in general. For instance, `promote_type(Bool, Bool) === Bool` but `typeof(+(zero(Bool), zero(Bool))) === Int`. Therefore such computations are also used in eg Statistics: https://github.com/JuliaStats/Statistics.jl/blob/bfc6326267f6fb922707ac0465e44218ff475d46/src/Statistics.jl#L216

src/symbolic_dimensions.jl
```diff
@@ -95,15 +106,122 @@ end
...
+    I = Vector{INDEX_TYPE}(undef, 0)
+    V = Vector{T}(undef, 0)
+    nzdims_l = getfield(l, :nzdims)
+    nzvals_l = getfield(l, :nzvals)
+    nzdims_r = getfield(r, :nzdims)
+    nzvals_r = getfield(r, :nzvals)
+    nl = length(nzdims_l)
+    nr = length(nzdims_r)
+    il = ir = 1
+    while il <= nl && ir <= nr
```
MilesCranmer, line 177: Is it worth it to add `@inbounds` here (and elsewhere)?

src/symbolic_dimensions.jl
devmotion: IMO one should not be worried about the compiler here. It will also not cause any type instabilities, in the worst case there's one additional operation (but for common types the compiler should figure this out at compile time).

MilesCranmer: Ah, damn, I see. Okay we can roll that back. Maybe we can have a function

```julia
@generated function promote_type_by_eval(f::F, ::Type{L}, ::Type{R}) where {F,L,R}
    T = F.instance(zero(L), zero(R))
    return :($T)
end
```

just in case the compiler doesn't feel like constant propagation for certain complicated rational types?

I guess we should also implement this for the general `map_dimensions`, if possible.

devmotion: No, generated functions should be just a last resort and there's absolutely no need for any concerns here.

MilesCranmer: (Or just implementing it for `+` and `-` to be safe)

devmotion: I still don't understand why you would want to add or change anything. The code (with `op`) is completely fine, performant, and guaranteed to create arrays of the correct type.

MilesCranmer: Ah sorry our messages crossed and I didn't see what you said. Yeah sounds good, I'll remove it.

src/symbolic_dimensions.jl
MilesCranmer: Leaving this and other tweaks for later.

MilesCranmer (approved):
Cheers for the awesome contribution @devmotion!! 🎉🎉

Issue #51, opened 2023-09-30
Apparent Issue with Precompilation for Dependent Packages


cadojo:
Edit: Actually, this issue happens even if I don't use `Revise.jl`! I think the `@eval` line somehow breaks precompilation?

I'm developing a new package, `AstrophysicalCalculations`, which depends on `DynamicQuantities`, and I'm using `Revise` to assist with that development. When I enter `using Revise, AstrophysicalCalculations`, I get the error shown below. I believe that the `@eval` line in `symbolic_dimensions.jl:215` makes using Revise impossible for dependent packages. I could be wrong / using this incorrectly, and this issue may be unresolvable, but I wanted to post this issue for awareness. Thank you for `DynamicQuantities.jl`!

The line where my package fails: [`src/src/StellarObservations.jl:111`](https://github.com/cadojo/AstrophysicalCalculations.jl/blob/af733323d98694c23d8126d4a33257326dd95729/src/src/StellarObservations.jl#L111).

<details>

```
ERROR: LoadError: Evaluation into the closed module `SymbolicUnitsParse` breaks incremental compilation because the side effects will not be permanent. This is likely due to some other module mutating `SymbolicUnitsParse` with `eval` during precompilation - don't do this.
Stacktrace:
  [1] eval
    @ ./boot.jl:370 [inlined]
  [2] (::DynamicQuantities.SymbolicUnitsParse.var"#1#2")()
    @ DynamicQuantities.SymbolicUnitsParse ~/.julia/packages/DynamicQuantities/AS62k/src/symbolic_dimensions.jl:162
  [3] lock(f::DynamicQuantities.SymbolicUnitsParse.var"#1#2", l::Base.Threads.SpinLock)
    @ Base ./lock.jl:229
  [4] _generate_unit_symbols
    @ ~/.julia/packages/DynamicQuantities/AS62k/src/symbolic_dimensions.jl:159 [inlined]
  [5] sym_uparse(raw_string::String)
    @ DynamicQuantities.SymbolicUnitsParse ~/.julia/packages/DynamicQuantities/AS62k/src/symbolic_dimensions.jl:186
  [6] var"@us_str"(__source__::LineNumberNode, __module__::Module, s::Any)
    @ DynamicQuantities ~/.julia/packages/DynamicQuantities/AS62k/src/symbolic_dimensions.jl:215
  [7] #macroexpand#63
    @ ./expr.jl:119 [inlined]
  [8] macroexpand
[...]

PR #58, opened 2023-10-13
Warn user about precompilation of symbolic units

MilesCranmer:
As noted in #51 by @cadojo, `SymbolicDimensions` parsing is not compatible with precompilation because the unit and constant variables are only created at runtime. (The reason for this is that it can take about a second to generate all the symbols. So, to keep startup time small for performant libraries, the symbolic dimensions are generated at runtime.)

This PR:

1. Prevents the creation of the unit symbols/constants if the user happens to trigger this, by exiting `_generate_unit_symbols` is precompilation is active.
2. Prints a helpful warning to the user:

> Creating SymbolicDimensions objects during precompilation is not allowed, as symbolic units are not created until the first runtime call. You should use regular `Dimensions` instead for computations, and generally use `SymbolicDimensions` for display purposes. If needed, you can manually create `SymbolicDimensions` by calling the constructor explicitly, such as `Quantity(1.5, SymbolicDimensions; km=1, s=-1)`, which is equivalent to `1.5us"km/s"`.

What do you think @cadojo?

cc @gaurav-arya

### DIFF
--- a/src/symbolic_dimensions.jl
+++ b/src/symbolic_dimensions.jl
@@ -240,6 +240,8 @@ module SymbolicUnitsParse
     import ...DEFAULT_VALUE_TYPE
     import ...DEFAULT_DIM_BASE_TYPE
 
+    _is_precompiling() = ccall(:jl_generating_output, Cint, ()) == 1
+
     # Lazily create unit symbols (since there are so many)
     module Constants
         import ..CONSTANT_SYMBOLS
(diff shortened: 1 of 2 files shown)


cadojo:
> What do you think @cadojo?

Looks good to me! Thanks for tagging me this patch. I hadn't considered the manual constructer either.

src/symbolic_dimensions.jl
```diff
@@ -314,7 +318,19 @@ module SymbolicUnitsParse
 
     const UNIT_SYMBOLS_EXIST = Ref{Bool}(false)
     const UNIT_SYMBOLS_LOCK = Threads.SpinLock()
-    function _generate_unit_symbols()
+    function _generate_unit_symbols(; testing=Val(false))
```
cadojo, line 321: Just an FYI — the changes in this PR completely address what I observed in #51! I checked out this branch with my `AstrophysicalCalculations` package, and found the expected warning. Then I changed my code to use the `Quantity(..., SymbolicDimensions; pc=1)` constructor, as instructed by the warning, and my package precompiled as expected.

MilesCranmer: Awesome!

The only reason I'm hesitating to merge is I'm not sure there are any cases where you may actually need to run through that function. I ran some testing and it seemed to hit that error even without necessarily precompiling a module with it... So maybe we could just leave the warning, but still allow it to execute as normal?

cadojo: >  I ran some testing and it seemed to hit that error even without necessarily precompiling a module with it...

That's interesting. I don't know enough about Julia's stages of compilation / precompilation. I think the warning is great; if I saw it before making #51, I would have used the `Quantity` constructor instead of filing an issue. Of course, no rush merging anything from my point of view; `AstrophysicalCalculations` is a "brush up on physics before graduate school" kind of package 😉 and the constructor is a great solution as-is.

MilesCranmer:
This is no longer needed because the feature has been implemented! 😄 #101 

But `u"..."` will still fail during precompilation until #106 is merged.

PR #61, opened 2023-10-14
Add `uconvert` method for `QuantityArray`

MilesCranmer:
cc @gaurav-arya

(For what it's worth, it seems like using `uconvert(us"km").(array)` works, but it relies on compiler inlining a lot of stuff to be as efficient. Also, `expand_units` has a dedicated method, so I think it makes sense to do it here as well.)


gaurav-arya (approved):
Looks good to me. One nit: the `@assert isone(...` is missing for the array method.

MilesCranmer:
Good catch! Added.

PR #66, opened 2023-10-16
Create LazyFloat64 for units

### LINKED ISSUE #65: Choosing type of quantity value based on type used in construction
Arguably, when we do `1.0f0u"m"`, the floating point backing type should remain `Float32` rather than being converted to `Float64`:
```julia
julia> (1.0f0u"m").value
1.0
```

In general, there's a question of what type we should choose for the quantity value. In theory, we could even keep integers as integers, i.e. `1u"m"` would remain `1 m`. But perhaps we would want to enforce floating point, in which case something like `float(T)` might make sense.

MilesCranmer:
Attempt 2 at #30. cc @gaurav-arya - let me know what you think. This should fixes #65.

The following behavior occurs:

| Input | Resulting Numeric Type |
| --- | --- |
| `1f0u"km/s"` | `Float32` type |
| `Float16(1.0)u"km/s"` | `Float16` type |
| `1.0u"km/s"` | `Float64` type |
| `big(1.0)u"km/s"` | `BigFloat` type |
| `1u"km/s"` | `Float64` type |

TODO remaining:
- [ ] Add LazyFloat64 for SymbolicDimensions versions of units
- [ ] Add tests for arrays


MilesCranmer:
> This seems too complicated to me. Isn't it possible to just convert apply `float` to parsed quantities as suggested in #65 but otherwise not modify the user input?

I am not sure how to do it any other way. Do you have an implementation in mind?

devmotion:
To me it seems the main problem is that the current implementation of the `u_str` macro always returns a `Quantity`. If no values are given, I think it should just return a `Dimensions`. This would also match the behaviour of Unitful:
```julia
julia> typeof(Unitful.u"m")
Unitful.FreeUnits{(m,), 𝐋, nothing}

julia> typeof(Unitful.u"1f0m")
Unitful.Quantity{Float32, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}

julia> typeof(1f0Unitful.u"m")
Unitful.Quantity{Float32, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}}

julia> typeof(Unitful.u"1f0")
Float32
```

I think this would be more intuitive than the current behaviour of DynamicQuantities:
```julia
julia> typeof(DynamicQuantities.u"m")
DynamicQuantities.Quantity{Float64, DynamicQuantities.Dimensions{DynamicQuantities.FixedRational{Int32, 25200}}}

julia> typeof(DynamicQuantities.u"1f0m")
DynamicQuantities.Quantity{Float64, DynamicQuantities.Dimensions{DynamicQuantities.FixedRational{Int32, 25200}}}

julia> typeof(1f0DynamicQuantities.u"m")
DynamicQuantities.Quantity{Float64, DynamicQuantities.Dimensions{DynamicQuantities.FixedRational{Int32, 25200}}}

julia> typeof(DynamicQuantities.u"1f0")
DynamicQuantities.Quantity{Float64, DynamicQuantities.Dimensions{DynamicQuantities.FixedRational{Int32, 25200}}}
```

MilesCranmer:
These conversions are to make the return value of `uparse` type stable. In Unitful you can even do `u"(m, km, cm)"`. However in DynamicQuantities this is not allowed since it would make `uparse` return `::Any`.

> I think it should just return a Dimensions

The operator `*` does not exist between `Number` and `Dimensions` which is on purpose. This is because it would necessitate choosing a default `AbstractQuantity`.

MilesCranmer:
Also keep in mind (I see your edit mentioned this) that all units and constants have numerical values in SI base units, so they are stored as `Quantity` rather than just `Dimensions`.

devmotion:
> These conversions are to make the return value of uparse type stable

It's a macro, so the string is replaced with whatever expression you want to replace it with. And then the resulting function/code/etc. is compiled by the Julia compiler. I don't see how the conversions help type stability.

As a quick demonstration I replaced the `uparse` function with just `return eval(Meta.parse(s))` (BTW I wonder if one could omit `eval` and just replace the `m` etc. Symbols with `$(Units.m)` etc., such that all evaluation is only performed after the macro has been applied). There are no type instability issues:
```julia
julia> using DynamicQuantities, Test

julia> julia> f(x) = x * u"3m/s"

julia> @inferred f(3)
9.0 m s⁻¹

julia> @inferred f(1.0)
3.0 m s⁻¹

julia> @code_warntype f(2)
MethodInstance for f(::Int64)
  from f(x) @ Main REPL[18]:1
Arguments
  #self#::Core.Const(f)
  x::Int64
Body::Quantity{Float64, Dimensions{DynamicQuantities.FixedRational{Int32, 25200}}}
1 ─ %1 = (x * 3.0 m s⁻¹)::Core.PartialStruct(Quantity{Float64, Dimensions{DynamicQuantities.FixedRational{Int32, 25200}}}, Any[Float64, Core.Const(m s⁻¹)])
└──      return %1

julia> @code_lowered f(1.0)
CodeInfo(
1 ─ %1 = x * 3.0 m s⁻¹
└──      return %1
)
```

> This is because it would necessitate choosing a default AbstractQuantity.

To be precise, it would only require to choose  a default quantity type for `Dimensions`. And in that case it might be fine to choose `Quantity` since it is the default quantity type anyway and is already based on `Dimensions`?

devmotion:
I still think that it would be preferable to change the macro and parsing in more fundamental ways. But in any case, I don't think adding a new numeric type is a good idea (I've seen too many things go wrong in different packages). The much simpler approach would be to make the "units" quantities with `Bool`ean values. `Bool`s are frequently used to avoid undesired promotions since they are the least invasive default number types in Julia:
```julia
julia> 1.0 * true
1.0

julia> 2 * true
2

julia> Float16(3) * true
Float16(3.0)

julia> 4f0 * true
4.0f0
```

MilesCranmer:
Unlike Bool I think we want to convert integers to floats, otherwise the user could write `1u"cm"` and get hit with a rounding error which seems unpleasant.

devmotion:
I'm not sure if that's a good idea. To me it seems rather surprising to not respect the types chosen by a user (which also lead to #65 and is inconsistent with Unitful).

If you really want to prevent this case, you could specialize `*(::Int, ::Quantity{Bool})` etc. (I'm a bit worried that specializing `*(::Integer, ::Quantity{Bool})` creates invalidations but this should be checked, of course).

MilesCranmer:
The user could still convert to a `Quantity{Int}` afterwards. It’s not so much it doesn’t respect their choice, it’s more: the units and constants are already floating point numerical values, and promote accordingly. Writing `1u"cm"` is multiplying an integer with a float.

gaurav-arya:
To clarify the key problem: the question is, when the user writes something like `u"fm"`, what should the resulting quantity look like? Since the quantity will be represented in SI, a `1e-15` has to be written *somewhere*, and thus a floating point type needs to be chosen. After understanding this, I see why the current u-str behaviour is what it is. In theory users could always use the proper constructor `Quantity` to specify the value type (I believe? Or does that also promote to `Float64`?) However, the u-str API is the most convenient way to create a quantity right now, and furthermore the promotion behaviour would be surprising to those who do not understand the internals of `DynamicQuantities.jl`. This raises the question: can we make the recommended / easiest to use API for constructing quantities more generic with respect to the value type?

It's a tricky problem and I don't have a perfect solution, but hopefully the thoughts below are helpful.

1. The simplest solution from an implementation perspective would be for the recommend API to require users to specify both the value and the units of the quantity at once. This would not require `AutoFloat`. The drawback is that the familiar u-str approach would no longer be OK, we'd have to tweak the API (or could it? Is there a way of keeping the same feel of the u-str approach, but somehow still requiring a value type that can be seen by `DynamicQuantities.jl` before constructing the first quantity). One API here (which takes inspiration from option 2.) would be to enforce expressions of the form `u"x * m"` in u-strs and syntactically extract `x` and use it as the value.

2. Alternatively, one could view the form `x * u"m"` as essentially a *fancy constructor* for quantities, made possible with the `AutoFloat` logic. In this solution (the one currently implemented by this PR), I would recommend strongly discouraging creating plain unit strings `u"m"` (even if `x` happens to be `1.0`), as the form `x * u"m"` is how units should be constructed (or even `u"x * m"`, see the last sentence). In this perspective, I think `AutoFloat` has a nice interpretation: it is fancy magic for getting this `Unitful`-like constructor syntax to work with this package, but if quantity are constructed as recommended then it should dissapear immediately for subsequent calculations. Furthermore, if the user writes `u"x * m"` with `x` *inside* the u-str, no `AutoFloat`s would appear at runtime whatsoever, only in the construction at compile time.

MilesCranmer:
Thanks for sharing all of your thoughts on this!

The one comment I will make is that I think we should try to avoid an API that requires things like `u"x * m"`. I think over-reliance on macros is not a great code pattern and introduces brittleness and new bugs (e.g., what if a user wants to have `x` as a variable?). Also we would still want to allow the user to do things like:

```julia
using DynamicQuantities.Units: m, kg
using DynamicQuantities.Constants: c

myval = 0.3 * kg/m^3 * c^2
```

which is just normal Julia code. Those `m`, `kg`, and `c` need to have some type after all, so the solution should be compatible with this as well.

Keep in mind this `AutoFloat` proposal is just a thin wrapper around `Float64` so I wouldn't expect it to have any runtime penalties as the compiler would inline stuff. But yeah I 100% agree that if we go this route we should try to force the user to never use `AutoFloat` in any computations; it's just there for storing the physical unit/constant's value – but should get converted immediately. The idea of throwing easy-to-understand errors for any other pattern is a good idea.

MilesCranmer:
Just to clarify something – the promotion rules here turn out to be the same as used in irrational numbers. e.g.,

```julia
julia> promote(π, 1) |> typeof
Tuple{Float64, Float64}

julia> promote(π, 1f0) |> typeof
Tuple{Float32, Float32}
```

PR #72, opened 2023-10-22
Switch to using ConstructionBase

MilesCranmer:
Rather than using our own function for `constructorof`, just use the one in ConstructionBase.jl, so (1) the code is more readable, and (2) the behavior is easier to overload.

@devmotion (I think you suggested this in maybe #54?)

### DIFF
--- a/src/arrays.jl
+++ b/src/arrays.jl
@@ -31,7 +31,7 @@ struct QuantityArray{T,N,D<:AbstractDimensions,Q<:AbstractUnionQuantity{T,D},V<:
     dimensions::D
 
     function QuantityArray(v::_V, d::_D, ::Type{_Q}) where {_T,_N,_D<:AbstractDimensions,_Q<:AbstractUnionQuantity,_V<:AbstractArray{_T,_N}}
-        Q_out = constructor_of(_Q){_T,_D}
+        Q_out = constructorof(_Q){_T,_D}
         return new{_T,_N,_D,Q_out,_V}(v, d)
     end
 end
@@ -74,7 +74,7 @@ function Base.convert(::Type{QA1}, A::QA2) where {QA1<:QuantityArray,QA2<:Quanti
     N = ndims(QA1)
 
     raw_array = Base.Fix1(convert, Q).(A)
-    output = QuantityArray(convert(constructor_of(V){Q,N}, raw_array))
+    output = QuantityArray(convert(constructorof(V){Q,N}, raw_array))
     # TODO: This will mess with static arrays
 
     return output::QA1
--- a/Project.toml
+++ b/Project.toml
@@ -5,6 +5,7 @@ version = "0.7.5"
 
 [deps]
 Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
+ConstructionBase = "187b0558-2788-49d3-abe0-74a17ed4e7c9"
 PackageExtensionCompat = "65ce6f38-6b18-4e1d-a461-8949797d7930"
 Tricks = "410a4b4d-49e4-4fbc-ab6d-cb71b17b3775"
 
@@ -22,6 +23,7 @@ DynamicQuantitiesUnitfulExt = "Unitful"
 
 [compat]
 Compat = "3.42, 4"
+ConstructionBase = "1"
 Measurements = "2"
 PackageExtensionCompat = "1.0.2"
 ScientificTypes = "3"
--- a/src/symbolic_dimensions.jl
+++ b/src/symbolic_dimensions.jl
@@ -1,3 +1,6 @@
+import ConstructionBase: constructorof
+import Tricks: static_fieldnames
+
 import .Units: UNIT_SYMBOLS, UNIT_MAPPING, UNIT_VALUES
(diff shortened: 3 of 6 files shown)


src/arrays.jl
```diff
@@ -31,7 +32,7 @@ struct QuantityArray{T,N,D<:AbstractDimensions,Q<:AbstractUnionQuantity{T,D},V<:
     dimensions::D
 
     function QuantityArray(v::_V, d::_D, ::Type{_Q}) where {_T,_N,_D<:AbstractDimensions,_Q<:AbstractUnionQuantity,_V<:AbstractArray{_T,_N}}
-        Q_out = constructor_of(_Q){_T,_D}
+        Q_out = constructorof(_Q){_T,_D}
```
devmotion, line 35: IMO this is wrong in general (both the old and new version). There's no guarantee that `consteuctorof` returns a parameterless type and even if in general  these type parameters might be wronf or not even exist.

src/arrays.jl
```diff
@@ -74,7 +75,7 @@ function Base.convert(::Type{QA1}, A::QA2) where {QA1<:QuantityArray,QA2<:Quanti
     N = ndims(QA1)
 
     raw_array = Base.Fix1(convert, Q).(A)
-    output = QuantityArray(convert(constructor_of(V){Q,N}, raw_array))
+    output = QuantityArray(convert(constructorof(V){Q,N}, raw_array))
```
devmotion, line 78: Same here.

src/arrays.jl
MilesCranmer: Not sure if this applies here since it is guaranteed that `Q <: AbstractUnionQuantity`? Also `T` and `D` are also required by the function.

src/arrays.jl
MilesCranmer: What should be used instead?

MilesCranmer: n.b., `array_type` is defined by:

https://github.com/SymbolicML/DynamicQuantities.jl/blob/151d3246071082be2edab606427e85e72a737a97/src/arrays.jl#L87

so it will throw an error if these type parameters are not available (like in a `UnionAll`)

MilesCranmer: Similar for `quantity_type`:

https://github.com/SymbolicML/DynamicQuantities.jl/blob/151d3246071082be2edab606427e85e72a737a97/src/arrays.jl#L90

MilesCranmer: I guess I could do something like:

```julia
    Q = quantity_type(QA1)
    D = dimension_type(QA1)
    V = array_type(QA1)

    output = QuantityArray(convert(V, ustrip(A)), convert(D, dimension(A)), Q)
```

and that would avoid the `constructorof`

MilesCranmer: Fixed.

src/arrays.jl
MilesCranmer: Maybe I should just rename it to be `base_quantity` or something, and get people to overload it? (Rather than `constructor`, which maybe is not appropriate)

devmotion: IMO these methods should just not be defined in general but only for concrete types such as `Quantity` - and in that case I think it would be preferable and easier to just call the desired constructor (`Quantity`, `Quantity{T}` or `Quantity{T,D}`) directly.

For other subtypes these type parameters might not even exist - a subtype of `AbstractQuantity{T,D}` can hardcode `T` and `D` to specific types. Moreover, even if `T` and `D` are still type parameters it is not guaranteed that `T` and `D` are also the first and second type parameter of the subtype.

MilesCranmer: I see what you mean but I think this might just be too complicated to be worth it. If the user wants to make an `AbstractQuantity{T,D}` and  not have `T` be the value type, then they would break several assumptions throughout the `QuantityArray` interface (which sets `T` based on the array you pass in).

Idk. I guess I'd rather have things break quickly and in an interpretable way via failing to build this constructor, rather than try to accommodate some very complex behavior for some user-defined struct that does not even exist.

So my vote is to just leave this as a TODO for later. Sound good?

devmotion: My suggestion would be to directly fix these issues: You just have to restrict this function to `_Q <: Union{Quantity, ...}` (whatever other concrete types exist in DynamicQuantities).

MilesCranmer: Do you mean: disallow any abstract quantity that is not `Quantity` or `GenericQuantity`?

Or, do you mean a trait-based approach like writing a custom function like

```julia
with_param_types(::Type{<:AbstractUnionQuantity}, _, _) = error("Please write a custom method")
with_param_types(::Type{<:Quantity}, ::Type{T}, ::Type{D}) = Quantity{T,D}
with_param_types(::Type{<:GenericQuantity}, ::Type{T}, ::Type{D}) = GenericQuantity{T,D}
```

This seems clean. Wdyt?

devmotion: I would only support `Quantity` and `GenericQuantity`. My general view is that code should be as simple as possible and traits and other generalizations should only be added if there is a clear use case for them.

MilesCranmer: Sg, added.

MilesCranmer:
I'd like to get this into v0.8 since it's a breaking change. So let me know soon if there are any other comments...

PR #74, opened 2023-10-27
Add RecursiveArrayTools.jl extension

MilesCranmer:
@ChrisRackauckas could you confirm this is the right behavior? I couldn't find many docs on this method unfortunately, but it seems required for downstream tools

```julia
function RecursiveArrayTools.recursive_unitless_bottom_eltype(::Type{<:AbstractQuantity{T}}) where {T}
    return T
end

function RecursiveArrayTools.recursive_unitless_eltype(::Type{<:AbstractQuantity{T}}) where {T}
    return T
end
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -11,12 +11,14 @@ Tricks = "410a4b4d-49e4-4fbc-ab6d-cb71b17b3775"
 [weakdeps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Measurements = "eff96d63-e80a-5855-80a2-b1b0885c5ab7"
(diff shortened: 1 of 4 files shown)


ChrisRackauckas:
I think you'll need a few changes to OrdinaryDiffEq.jl itself. Also, we'll need to think about how some things like scalar tolerances are handled if there's multiple different units, that needs some representation like a fillvector just for the units

MilesCranmer:
By multiple different units do you mean if the user were to use `SymbolicDimensions`, which let's you choose `cm` or `m` for storing length? The default `Dimensions` type enforces SI units everywhere; will that be an issue?

PR #75, opened 2023-10-27
Linear Algebra extension: `svd`, `inv`, `*`, `/`, `\` for `QuantityArray`

ggebbie:
Thanks for the cool and promising package!

On Julia Discourse there seemed to be some interest in extending this package for "multidimensional analysis" (following George W. Hart) which I mostly interpret to be linear algebra utilities for the `QuantityArray` type. I don't know whether this fits with the vision of this package or whether it is best completed in a separate package. Any comments/advice on this PR are appreciated.

Here I make a start using the existing `DynamicQuantitiesLinearAlgebraExt` to add functionality for: `inv`, `(\)`, `svd`, `SVD`, `Diagonal`, `eigen`, and `det` for `QuantityArray`s. 
All functions here use the "algebraic" interpretation of dimensional analysis.

Because `QuantityArray`s represent a dimensionally uniform matrix, the number of applicable linear algebra methods is fairly restricted. In the parlance of Multidimensional Analysis, an array would be any combination of values and dimensions/units in a tabular form. A matrix or multipliable matrix would be a restricted dimensional structure such that multiplication and other linear algebra functions are possible. In this light, the `QuantityArray` name for uniform matrices is somewhat confusing.

I think that other multipliable matrices (i.e., non-uniform) would best be represented by a different `struct` with the dimensional range and domain as two of the fields. This goes beyond an extension to this package and I plan to experiment on this idea in a separate package.


ext/DynamicQuantitiesLinearAlgebraExt.jl
```diff
@@ -1,8 +1,69 @@
...
-import LinearAlgebra: norm
-import DynamicQuantities: AbstractQuantity, ustrip, dimension, new_quantity
+import Base: (*) #, transpose
+import LinearAlgebra: norm, inv, (\), svd, Algorithm, default_svd_alg, SVD, Diagonal, Adjoint, Transpose, AbstractRotation, AbstractMatrix, eigen, eigsortby, Eigen, det
+import DynamicQuantities: AbstractQuantity, ustrip, dimension, new_quantity, AbstractDimensions, QuantityArray, Quantity
 
 norm(q::AbstractQuantity, p::Real=2) = new_quantity(typeof(q), norm(ustrip(q), p), dimension(q))
 
+# handle ambiguities
+*(q::QuantityArray,r::QuantityArray) = QuantityArray(ustrip(q)*ustrip(r),dimension(q)*dimension(r))
```
MilesCranmer, line 10: I think these should not go in the extension but in `src/arrays.jl`. (just the `Base` stuff, the rest can stay).

ext/DynamicQuantitiesLinearAlgebraExt.jl
```diff
@@ -1,8 +1,69 @@
...
+*(q::QuantityArray,r::AbstractMatrix) = QuantityArray(ustrip(q)*r,dimension(q))
+*(q::QuantityArray,r::AbstractVector)  = QuantityArray(ustrip(q)*r,dimension(q))
+*(r::AbstractMatrix,q::QuantityArray) = QuantityArray(r*ustrip(q),dimension(q))
+*(r::AbstractVector,q::QuantityArray) = QuantityArray(r*ustrip(q),dimension(q))
+
+# a choice to make this function type stable (to be reviewed/revisited)
+#transpose(q::QuantityArray) = QuantityArray(transpose(ustrip(q)),dimension(q))
+
+\(q::QuantityArray,r::QuantityArray) = QuantityArray(ustrip(q)\ustrip(r),dimension(r)/dimension(q))
+\(q::QuantityArray,r::Union{AbstractVector,AbstractMatrix}) = QuantityArray(ustrip(q)\r,dimension(q)^-1)
```
MilesCranmer, line 20: Could this be written as `inv` instead? (Just to match elsewhere)

ext/DynamicQuantitiesLinearAlgebraExt.jl
```diff
@@ -1,8 +1,69 @@
...
+# not implemented, AbstractMatrix \ QuantityArray
+
+inv(Q::QuantityArray) = QuantityArray(inv(ustrip(Q)),dimension(Q)^-1)
```
MilesCranmer, line 23: Likewise. Just can call `inv(dimension(..))`

ext/DynamicQuantitiesLinearAlgebraExt.jl
```diff
@@ -1,8 +1,69 @@
...
+*(q::QuantityArray,r::QuantityArray) = QuantityArray(ustrip(q)*ustrip(r),dimension(q)*dimension(r))
+*(q::QuantityArray,r::AbstractMatrix) = QuantityArray(ustrip(q)*r,dimension(q))
+*(q::QuantityArray,r::AbstractVector)  = QuantityArray(ustrip(q)*r,dimension(q))
+*(r::AbstractMatrix,q::QuantityArray) = QuantityArray(r*ustrip(q),dimension(q))
+*(r::AbstractVector,q::QuantityArray) = QuantityArray(r*ustrip(q),dimension(q))
+
+# a choice to make this function type stable (to be reviewed/revisited)
+#transpose(q::QuantityArray) = QuantityArray(transpose(ustrip(q)),dimension(q))
+
+\(q::QuantityArray,r::QuantityArray) = QuantityArray(ustrip(q)\ustrip(r),dimension(r)/dimension(q))
```
MilesCranmer, line 19: Can limit these to only matrices by setting the second parametric type to `2` instead of *all* QuantityArray.

MilesCranmer:
Nice work!! I added some initial comments.

ext/DynamicQuantitiesLinearAlgebraExt.jl
MilesCranmer: (If we want)

ggebbie: I attempted to restrict to 1D or 2D arrays, i.e., second parametric type to `1` or `2`. Please check code style/correctness and I can revise. Added another test to double-check.

ggebbie:
Failed Julia 1.6 runtest required some additional code for `svd` and `Diagonal`. I believe that the code should work for 1.9 and 1.6 (awaiting 1.6 GitHub Action to be sure). Involved adding some uglier code, however.

ext/DynamicQuantitiesLinearAlgebraExt.jl
```diff
@@ -1,8 +1,69 @@
...
+    Exists for uniform matrices which includes all `QuantityArray`s (pp. 124, Hart, 1995).
+
+    Returns SVD factorization of parametric type: `SVD{T, Quantity{T, Dimensions{DynamicQuantities.FixedRational{Int32, 25200}}}, Matrix{T}, QuantityArray{T, 1, Dimensions{DynamicQuantities.FixedRational{Int32, 25200}}, Quantity{T, Dimensions{DynamicQuantities.FixedRational{Int32, 25200}}}, Vector{T}}}`.
+
+    Factorization `F` can be deconstructed: `U,σ,V = F`. 
+
+    Functions working: , `size`, `adjoint`.
+    Functions partially working: `inv`
+    Functions not working: `svdvals`, `ldiv!`.
+"""
```
MilesCranmer, line 29: I don't think these are needed, since it will already show up for `?svd`? Likewise for `eigen`.

ext/DynamicQuantitiesLinearAlgebraExt.jl
```diff
@@ -1,8 +1,69 @@
...
-import LinearAlgebra: norm
-import DynamicQuantities: AbstractQuantity, ustrip, dimension, new_quantity
+import LinearAlgebra: norm, inv, (\), svd, Algorithm, default_svd_alg, SVD, Diagonal, Adjoint, Transpose, AbstractRotation, AbstractMatrix, eigen, eigsortby, Eigen, det
+import DynamicQuantities: AbstractQuantity, ustrip, dimension, new_quantity, AbstractDimensions, QuantityArray, Quantity
+
+const QuantityArrayVecOrMat{T} = Union{QuantityArray{T,2},QuantityArray{T,1}} where T
 
 norm(q::AbstractQuantity, p::Real=2) = new_quantity(typeof(q), norm(ustrip(q), p), dimension(q))
 
+\(q::QuantityArrayVecOrMat,r::QuantityArrayVecOrMat) = QuantityArray(ustrip(q)\ustrip(r),dimension(r)/dimension(q))
```
MilesCranmer, line 10: Small nitpick and I'm happy to do it but could you space each method definition out? i.e., 
```suggestion
\(q::QuantityArrayVecOrMat, r::QuantityArrayVecOrMat) = QuantityArray(ustrip(q) \ ustrip(r), dimension(r) / dimension(q))
```

ext/DynamicQuantitiesLinearAlgebraExt.jl
```diff
@@ -1,8 +1,69 @@
...
+\(q::QuantityArrayVecOrMat,r::Union{AbstractVector,AbstractMatrix}) = QuantityArray(ustrip(q)\r,inv(dimension(q)))
```
MilesCranmer, line 11: Nitpick: for binary operators I would be more meaningful to have `l` and `r` rather than `q` and `r` (especially when both inputs are quantities), rather than a mix. (Also applies elsewhere)

ext/DynamicQuantitiesLinearAlgebraExt.jl
```diff
@@ -1,8 +1,69 @@
...
+
+    Returns SVD factorization of parametric type: `SVD{T, Quantity{T, Dimensions{DynamicQuantities.FixedRational{Int32, 25200}}}, Matrix{T}, QuantityArray{T, 1, Dimensions{DynamicQuantities.FixedRational{Int32, 25200}}, Quantity{T, Dimensions{DynamicQuantities.FixedRational{Int32, 25200}}}, Vector{T}}}`.
+
+    Factorization `F` can be deconstructed: `U,σ,V = F`. 
+
+    Functions working: , `size`, `adjoint`.
+    Functions partially working: `inv`
+    Functions not working: `svdvals`, `ldiv!`.
+"""
+function svd(A::QuantityArray;full=false,alg::Algorithm = default_svd_alg(ustrip(A)))
```
MilesCranmer, line 30:
```suggestion
function svd(A::QuantityArray; full=false, alg::Algorithm=default_svd_alg(ustrip(A))) 
```

ext/DynamicQuantitiesLinearAlgebraExt.jl
```diff
@@ -1,8 +1,69 @@
 module DynamicQuantitiesLinearAlgebraExt
 
-import LinearAlgebra: norm
-import DynamicQuantities: AbstractQuantity, ustrip, dimension, new_quantity
+import LinearAlgebra: norm, inv, (\), svd, Algorithm, default_svd_alg, SVD, Diagonal, Adjoint, Transpose, AbstractRotation, AbstractMatrix, eigen, eigsortby, Eigen, det
+import DynamicQuantities: AbstractQuantity, ustrip, dimension, new_quantity, AbstractDimensions, QuantityArray, Quantity
```
MilesCranmer, line 4: I learned this the other day and haven't yet put it in for the rest of the code, but this should be 
```suggestion
import LinearAlgebra: norm, svd, Diagonal, eigen, det
using LinearAlgebra: Algorithm, default_svd_alg, SVD, Adjoint, Transpose, AbstractRotation, AbstractMatrix, eigsortby, Eigen
using DynamicQuantities: AbstractQuantity, ustrip, dimension, new_quantity, AbstractDimensions, QuantityArray, Quantity
```
because `import` should technically only be used if you are overloading a method.

Meanwhile the `inv` and `\` definitions should go in the main codebase since these are exported by `Base` rather than `LinearAlgebra` (they are only overloaded there, rather than given their original definitions).

Also, `AbstractMatrix` is already available in Julia, so does not need the import.

Also, some of these imports are not used – `Eigen`

ext/DynamicQuantitiesLinearAlgebraExt.jl
MilesCranmer: Oh wait maybe it is needed; is the behavior different at all?

ext/DynamicQuantitiesLinearAlgebraExt.jl
MilesCranmer: Looks good!

MilesCranmer:
Looks like stdlib has some bugs maybe?

```julia
julia> X = diagm(fill(1u"km/s", 3))
ERROR: Cannot create an additive identity for a `AbstractQuantity` type, as the dimensions are unknown. Please use `zero(::AbstractQuantity)` instead.
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] zero(::Type{Quantity{Float64, Dimensions{DynamicQuantities.FixedRational{Int32, 25200}}}})
   @ DynamicQuantities ~/Documents/DynamicQuantities.jl/src/utils.jl:128
 [3] diagm_container(size::Nothing, kv::Pair{Int64, QuantityArray{Float64, 1, Dimensions{…}, Quantity{…}, Vector{…}}})
   @ LinearAlgebra ~/.julia/juliaup/julia-1.10.0-beta3+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/LinearAlgebra/src/dense.jl:317
```

it should be using `zero(::T)` rather than `zero(::Type{T})` here, I think.

ext/DynamicQuantitiesLinearAlgebraExt.jl
ggebbie: Same core behavior but some of the side functionality is not implemented. I think it makes sense to remove this and the `eigen` docstring. It could be good to leave a comment regarding non-implemented functionality.

ggebbie:
> Edit: I guess we might need to add a special method for `diagm` as well, since it uses `zero(::Type{T})` rather than `zero(::T)`. So best to `ustrip -> diagm -> QuantityArray(result, dimension(input))`

This is the same issue that led me to writing `Diagonal`.  Right -- best to make a special `diagm` function.

MilesCranmer:
I guess some of these issues could also be fixed if we could figure out a way to implement https://github.com/SymbolicML/DynamicQuantities.jl/issues/76, right? Trading type instability with compatibility (though people could get type stability if they simply used `zero(::T)` instead... but at least it would avoid any errors).

ggebbie:
Getting closer and hope to not have missed anything. Certainly ok if you edit this PR. I tried to test on Julia 1.6 locally but Pkg complained about Compat... package not being in registry so I quit.

ggebbie:
Compatibility/composability with other packages is a big reason why I'm interested in this package. Unitful doesn't work with Turing, for example. It seems like there are issues with SciML and Unitful as well. Making some design choices in this package so that it is more usable in combination with other packages would be a good idea in my opinion.

MilesCranmer:
Hey @ggebbie,

I just made a few changes:

1. I incorporated the new abstract interface to your LinearAlgebra functions, so that any `AbstractQuantity` or `AbstractGenericQuantity` is permitted, rather than just `Quantity`. (Uses `quantity_type` to get the base quantity out of `QuantityArray`)
2. I removed the `QuantityArray{T,2} where T` statements to simplify things. I realized that those errors will just get caught anyways by the method operating on the raw array, so there's no need to limit it there.
3. Switched the dimension errors to use `DimensionError` rather than `ErrorException`.


Let me know if this sounds good. Also, I have a question for you: is the output dimension of `eigen` supposed to be `dimension(A)`? Shouldn't the output of `eigen` just be a regular scalar for the eigenvalue, and a regular vector for the eigenvectors? i.e., because both eigenvalues and eigenvectors are invariant to the scale of the matrix (and scale is sort of like physical dimension).

If this is true then `eigen` should be changed to reflect this.

Also, could you add unittests? I aim to have pretty robust coverage of all the code in the repo.

Thanks!
Miles

MilesCranmer:
@j-fu this is probably also of interest to you

ggebbie:
Looks like some good additions above, but it will take me about a week to take a closer look due to holidays. 

I might be misunderstanding your thought process with regards to `eigen`. I checked the code and it looks correct to me and is consistent with Hart's textbook. Consider the eigenvalue equation: Ax = lambda x.   If A is uniform (i.e., all entries have the same dimension/unit), then the action of A increments (i.e., multiplies) the dimension of x by dimension(A). For the right hand side of the equation to have a consistent dimension with the left hand side, dimension(lambda) must be equal to dimension(A). x could be any uniform vector, but the convention is to make it dimensionless, as it is the simplest eigenvector that satisfies the equation. So it makes sense to report the eigenvalues with units but the eigenvectors without.

In more complicated (non-uniform) matrices, both lambda and x may require units. And for other non-uniform matrices, there may be no eigenstructure at all. The requirement is that a matrix is squarable (i.e., A*A is dimensionally conformable) for eigenvalues and eigenvectors to exist.

MilesCranmer:
Sounds good! Thanks for clarifying

MilesCranmer:
Ping on this @ggebbie; what's the status and are there any blockers I can help with?

ggebbie:
Glad that this is still on the radar. I haven't found time to work on it. It looks like the action items are 1) improving test coverage and 2) implementing `@test_throws`. I can certainly handle 1) and will let you know if 2) is a blocker after I learn more about it.

BambOoxX:
Just for reference. https://discourse.julialang.org/t/ann-unitfultensors-jl-efficient-arrays-with-units/114028/4

MilesCranmer:
@ggebbie, do you have time to finish it off? It seems like you were almost done. Even if not all of the interface is ready, that would be much appreciated. Like `inv` and `\` would be great!

ext/DynamicQuantitiesLinearAlgebraExt.jl
```diff
@@ -1,11 +1,94 @@
...
+        Union{LA.Transpose{T,V}, LA.Adjoint{T,V}} where {T,V<:AbstractVector},
+        Union{LA.Transpose{T,V}, LA.Adjoint{T,V}} where {T,V<:LA.Bidiagonal},
+        LA.AbstractTriangular,
+        Union{LA.LowerTriangular,LA.UpperTriangular},
+        Union{LA.UnitLowerTriangular,LA.UnitUpperTriangular},
+        LA.Diagonal,
+        LA.Bidiagonal,
+        LA.SymTridiagonal,
+        Union{LA.Hermitian{T,S}, LA.Symmetric{T,S}} where {T,S},
+    ),
```
MilesCranmer, line 38: (Just added this after a lot of pain)

Surely there's a better way to deal with this sort of stuff for custom array types? Or is this what we have to do?

MilesCranmer: (Leaving this blob of disambiguities to figure out later. Doesn't seem to affect the load time at all which is good. Just seems ugly and hard to maintain.)

MilesCranmer (approved):
Looking good. I took out the `eigen` stuff as it needs more tests but that could be added later.

There are a couple of branches that aren't tested but they are disambiguities, so I am okay with 99% coverage for the moment. We should get back to 100% later though.

PR #77, opened 2023-11-02
[WIP] Create wildcard dimensions for zero(::Type)

### LINKED ISSUE #76: Compatibility with `zero(::Type{T})`
It seems like a reason DynamicQuantities.jl is incompatible with many libraries right now is due to their usage of `zero(::Type{T})`, which works for Unitful but not for DynamicQuantities.jl.

Therefore I wonder if there is a way of defining something like: 

```julia
zero(::Type{Q}) where {Q<:AbstractQuantity} = AutoQuantityZero{Q}()
```

which could avoid checking for any dimension errors in operations, and act as a _neutral zero_.

The problem with this is type instability, but perhaps that's better than simply throwing an error...? Then as packages adapt they would get type stability again. (And of course any package already using `zero(::T)` would maintain type stability).

Thoughts @gaurav-arya?

We could even think of adding a special boolean field to `Quantity` to indicate this... thus making everything type stable. But it's a bit worrisome...

MilesCranmer:
So in principle this will solve the compatibility issue with having `zero(::Type{T})` sprinkled throughout various libraries. Fixes #76. 

But I am worried this might introduce bugs that are silent rather than loud as they are now...

@gaurav-arya do you think you could take a look when you get a chance?

---

The PR solves this issue in a simple way. It creates a special `AbstractDimensions` type for wildcard dimensions:

```julia
struct WildcardDimensions{R} <: AbstractDimensions{R}
    WildcardDimensions() = new{Bool}()
end
```

Then, it makes `zero` for any quantity **type** simply set `WildcardDimensions` as the dimensions:

```julia
Base.zero(::Type{Q}) where {T,Q<:UnionAbstractQuantity{T}} = constructorof(Q)(zero(T), WildcardDimensions())
```

which makes all dimensionality checks pass automatically:

```julia
# Wildcard dimensions are always compatible:
Base.:(==)(::AbstractDimensions, ::WildcardDimensions) = true
Base.:(==)(::WildcardDimensions, ::AbstractDimensions) = true

# For propagation rules, we assume all relevant dimensions are zero
Base.getproperty(::WildcardDimensions, ::Symbol) = zero(Bool)
Base.iszero(::WildcardDimensions) = true
```

That's pretty much it. Then you can do things like:

```julia
julia> x = 1u"km/s"
1000.0 m s⁻¹

julia> x + zero(typeof(x))
1000.0 m s⁻¹

julia> x * zero(typeof(x))
0.0 m s⁻¹
```

whereas before this would simply fail.

I thought about adding this for `oneunit` as well, but there I'm a bit worried since you could potentially have `oneunit(::Type)` take the dimensions of the first quantity it comes into contact with, rather than the type of the input. Since all the units in this library are zero at zero (we don't have Celsius), `zero` seems safe to use with wildcards. But I don't think `oneunit` would be safe to use this way.


gaurav-arya:
The `x * zero(typeof(x))` example doesn't quite seem right to me: why should the wildcard become dimensionless in that case? Would it be better for the wildcard to only support operations like `+` where all inputs are expected to have the same dimension?

MilesCranmer:
Good point. Yeah on second thought I’m not even sure of the utility here... Just adding zeros doesn’t have practical value, but every other use case will have some sort of silent bug now...

src/wildcard.jl
```diff
@@ -0,0 +1,29 @@
...
+
+# For propagation rules, we assume all relevant dimensions are zero
+Base.getproperty(::WildcardDimensions, ::Symbol) = zero(Bool)
+Base.iszero(::WildcardDimensions) = true
+
+# Always promote to the other dimension type:
+Base.promote_rule(::Type{D}, ::Type{<:WildcardDimensions}) where {D<:AbstractDimensions} = D
+
+# Other utility rules:
+Base.show(io::IO, ::WildcardDimensions) = print(io, "[*]")
```
gaurav-arya, line 24: If `dimension(x)` is printed with `x` having `WildcardDimensions`, this looks like an array (totally unimportant since this is WIP, but just writing before I forget)

src/wildcard.jl
```diff
@@ -0,0 +1,29 @@
...
+# Wildcard dimensions are always compatible:
+Base.:(==)(::AbstractDimensions, ::WildcardDimensions) = true
+Base.:(==)(::WildcardDimensions, ::AbstractDimensions) = true
+
+# For any zero/oneunit specified on the type, we return a wildcard dimension instead:
+Base.zero(::Type{Q}) where {T,Q<:UnionAbstractQuantity{T}} = constructorof(Q)(zero(T), WildcardDimensions())
+
+# For propagation rules, we assume all relevant dimensions are zero
+Base.getproperty(::WildcardDimensions, ::Symbol) = zero(Bool)
+Base.iszero(::WildcardDimensions) = true
```
gaurav-arya, line 18: These lines seem too dangerous as discussed. Maybe `getproperty` should just be replaced with an error: (this is not very carefully considered, just a thought)
```suggestion
Base.getproperty(::WildcardDimensions, ::Symbol) = error("Cannot get dimensions of quantity with WildcardDimensions. This is probably because the quantity was created via `zero(T)` for a type `T`. Instead, use `zero(x)` where `x` is a quantity with the desired dimensions.")
```

gaurav-arya (commented):
Hm, I guess it depends on what downstream uses of `zero(T)` look like in real life. My intuition is actually that most cases would be covered by handling only ops that expect the same dimensions for all arguments, although I could be totally off. 

Whether the more permissive behaviour is worth it in that case, I am not sure -- but I think it's worth investigating in practice a bit further. To me, a big question is whether the union splitting of something like `Quantity{T, Union{WildcardDimensions, AbstractDimensions}}` is going to be somewhat efficient. If it kills performance, then I don't think the more permissive behaviour is worth it -- OTOH, if union splitting + maybe even some friendly branch prediction can substantially reduce the cost, then adding the more permissive behaviour for `+` and friends looks somewhat appealing. (Still with a clear error message for things like `*`, see a suggestion below.)

MilesCranmer:
I've thought about this a bit more. I think it likely be a nightmare to have working due to associativity being broken.

However I think one alternative is to allow a user to temporarily disable dimension error rules in a particular scope, so that the first call of a `Base.:+(::Quantity, ::Number)`would simply take the first dimensions it sees. Maybe like

```julia
using DynamicQuantities

@dimension_globbing begin
    u"km/s" + one(u"km/s")
end
```

which would result as `1001 m/s`.

PR #78, opened 2023-11-04
Add many other numeric methods

MilesCranmer:
This adds methods for the following functions. Most of these will throw an error if the quantity has nonzero dimensions. However some of the functions allow nonzero dimensions, and simply return an output with the same dimensions.

- `div`
- `sin`
- `cos`
- `tan`
- `sinh`
- `cosh`
- `tanh`
- `asin`
- `acos`
- `asinh`
- `acosh`
- `atanh`
- `sec`
- `csc`
- `cot`
- `asec`
- `acsc`
- `acot`
- `sech`
- `csch`
- `coth`
- `asech`
- `acsch`
- `acoth`
- `sinc`
- `cosc`
- `cosd`
- `cotd`
- `cscd`
- `secd`
- `sinpi`
- `cospi`
- `sind`
- `tand`
- `acosd`
- `acotd`
- `acscd`
- `asecd`
- `asind`
- `log`
- `log2`
- `log10`
- `log1p`
- `exp`
- `exp2`
- `exp10`
- `expm1`
- `frexp`
- `exponent`
- `atan`
- `atand`
- `adjoint`
- `unsigned`
- `identity`
- `transpose`
- `copysign`
- `flipsign`
- `mod`
- `ldexp`
- `round`
- `floor`
- `trunc`
- `ceil`
- `significand`
- `modf`
- `signbit` (in `utils.jl`)

It also moves methods over to `math.jl` for:

- `real`
- `imag`
- `conj`
- `nextfloat`
- `prevfloat`

---

TODO:

- [x] Add unittests for each

### DIFF
--- a/src/math.jl
+++ b/src/math.jl
@@ -2,12 +2,15 @@ for (type, base_type, _) in ABSTRACT_QUANTITY_TYPES
     @eval begin
         Base.:*(l::$type, r::$type) = new_quantity(typeof(l), ustrip(l) * ustrip(r), dimension(l) * dimension(r))
         Base.:/(l::$type, r::$type) = new_quantity(typeof(l), ustrip(l) / ustrip(r), dimension(l) / dimension(r))
+        Base.div(x::$type, y::$type, r::RoundingMode=RoundToZero) = new_quantity(typeof(x), div(ustrip(x), ustrip(y), r), dimension(x) / dimension(y))
 
         Base.:*(l::$type, r::$base_type) = new_quantity(typeof(l), ustrip(l) * r, dimension(l))
         Base.:/(l::$type, r::$base_type) = new_quantity(typeof(l), ustrip(l) / r, dimension(l))
+        Base.div(x::$type, y::$base_type, r::RoundingMode=RoundToZero) = new_quantity(typeof(x), div(ustrip(x), y, r), dimension(x))
 
         Base.:*(l::$base_type, r::$type) = new_quantity(typeof(r), l * ustrip(r), dimension(r))
         Base.:/(l::$base_type, r::$type) = new_quantity(typeof(r), l / ustrip(r), inv(dimension(r)))
+        Base.div(x::$base_type, y::$type, r::RoundingMode=RoundToZero) = new_quantity(typeof(y), div(x, ustrip(y), r), inv(dimension(y)))
 
         Base.:*(l::$type, r::AbstractDimensions) = new_quantity(typeof(l), ustrip(l), dimension(l) * r)
         Base.:/(l::$type, r::AbstractDimensions) = new_quantity(typeof(l), ustrip(l), dimension(l) / r)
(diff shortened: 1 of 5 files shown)


src/math.jl
```diff
@@ -97,6 +100,92 @@ Base.sqrt(q::UnionAbstractQuantity) = new_quantity(typeof(q), sqrt(ustrip(q)), s
...
+end
+for (type, base_type, _) in ABSTRACT_QUANTITY_TYPES, f in (:atan, :atand)
+    @eval begin
+        function Base.$f(x::$type)
+            iszero(dimension(x)) || throw(DimensionError(x))
+            return $f(ustrip(x))
+        end
+        function Base.$f(y::$type, x::$type)
+            dimension(y) == dimension(x) || throw(DimensionError(y, x))
+            y, x = promote(y, x)
```
MilesCranmer, line 129: Probably just need to add `:atan` and `:atand` to the loop below `# Combining different abstract types`

src/math.jl
```diff
@@ -97,6 +100,92 @@ Base.sqrt(q::UnionAbstractQuantity) = new_quantity(typeof(q), sqrt(ustrip(q)), s
...
+    @eval function Base.$f(q::UnionAbstractQuantity)
+        return new_quantity(typeof(q), $f(ustrip(q)), dimension(q))
+    end
+end
+for (type, base_type, _) in ABSTRACT_QUANTITY_TYPES, f in (:copysign, :flipsign, :mod)
+    # These treat the x as the magnitude, so we take the dimensions from there,
+    # and ignore any dimensions on y, since those will cancel out.
+    @eval begin
+        function Base.$f(x::$type, y::$type)
+            return new_quantity(typeof(x), $f(ustrip(x), ustrip(y)), dimension(x))
```
MilesCranmer, line 158: Likewise here

src/utils.jl
```diff
@@ -107,14 +107,13 @@ function Base.isless(l::Number, r::UnionAbstractQuantity)
 end
 
 # Simple flags:
-for f in (:iszero, :isfinite, :isinf, :isnan, :isreal)
+# :iseven, :ispow2, :isfinite, :isinf, :isodd, :isinteger, :isreal,
+# :isnan, :isempty, :iszero
+for f in (:iszero, :isfinite, :isinf, :isnan, :isreal, :signbit)
```
MilesCranmer, line 112: Might as well add all of these because it might be a dimensionless quantity

MilesCranmer:
FYI the errors seem to be coming from this issue: https://github.com/JuliaPhysics/Measurements.jl/issues/162 so not related to this PR at all.

src/utils.jl
```diff
@@ -107,14 +106,13 @@ function Base.isless(l::Number, r::UnionAbstractQuantity)
 end
 
 # Simple flags:
-for f in (:iszero, :isfinite, :isinf, :isnan, :isreal)
+# :iseven, :ispow2, :isfinite, :isinf, :isodd, :isinteger, :isreal,
```
gaurav-arya, line 109: Why commented?

gaurav-arya (approved):
Looks good to me! Only thing I wondered is whether something like `exp(x)` should return an instance of the base type or a dimensionless quantity. This goes back to the `one` debate too (ugh), e.g. `f(q)  = iszero(q) ? one(q) : exp(q)` wouldn't be type stable currently. But I don't see an obvious answer here (nor do I think it has to be resolved in this PR). It is indeed nice to go back to the simpler scalars when we can, which is a big plus of the current behaviour.

src/utils.jl
MilesCranmer: Fixed, thanks!

MilesCranmer:
I think it's okay because `one` is supposed to return the exact same type, but a lot of these numerical operators are not required nor expected to. For example, your function `f(q)` would also be type unstable for `Int64` input, as it would get converted to a float output for `exp`. But `one`'s purpose is to return the exact same time representing a 1.

MilesCranmer:
I guess the downside is that if someone did stuff like

```julia
x = 0.9u"km/s"
y = [x, x / x, exp(x / x * 0.3)]
```

it would be `Vector{Number}`. Which feels incorrect...

So maybe you are right and we should just always keep to the original quantity type...

MilesCranmer:
I think I am 50/50 on this... So I'd welcome any opinions one way or the other.

MilesCranmer:
> it would be `Vector{Number}`. Which is bad and feels like it approaches Unitful type instabilities again...

I guess we could also fix this by simply defining a promotion rule for all the primitive numeric types in https://docs.julialang.org/en/v1/manual/integers-and-floating-point-numbers/. (Recalling that the reason I avoided a generic `Number` promotion rule is that it might conflict with other abstract types that are also `<: Number`)

gaurav-arya:
I do feel like `[3.0, 1u"m"]` should promote in any case, yes.

gaurav-arya:
My overall opinion is that the scalar-output approach for `exp` and friends seems very appealing, and I think we should take it unless we/others find an insurmountable obstable, e.g. type stability in a reasonable use case with no other fix.

MilesCranmer:
Okay we could rebase this on #84 if you are on board with that change

MilesCranmer:
> My overall opinion is that the scalar-output approach for exp and friends seems very appealing, and I think we should take it unless we/others find an insurmountable obstable, e.g. type instability in a reasonable use case with no other fix.

Sounds good!

PR #81, opened 2023-11-09
Add some `Unitful` unit names that were missing

MichaelHatherly:
There are a couple of differences between the naming of units in `Unitful` and this package that makes it not quite a direct swap when changing between the two packages. This adds a couple of aliases for current units as well as a new `wk`.

The only one that we're really needing here is `wk` since we can deal with aliasing the others internally where needed, so I'm fine with dropping those if there are reasons that they weren't included already?

---

How would you like tests written for these kinds of additions? If they are needed?

### DIFF
--- a/src/units.jl
+++ b/src/units.jl
@@ -158,15 +158,21 @@ end
 # Common assorted units
 ## Time
 @register_unit min 60 * s
+@register_unit minute min
 @register_unit h 60 * min
 @register_unit hr h
 @register_unit day 24 * h
+@register_unit d day
+@register_unit wk 7 * day
 @register_unit yr 365.25 * day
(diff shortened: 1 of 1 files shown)


MilesCranmer:
Looks good! The one suggestion I have is to mention these units and aliases in this line:
```julia
    "Time in seconds. Available variants: `fs`, `ps`, `ns`, `μs` (/`us`), `ms`, `min`, `h` (/`hr`), `day`, `yr`, `kyr`, `Myr`, `Gyr`.",
```

MilesCranmer (approved):

MichaelHatherly:
Thanks Miles.

PR #82, opened 2023-11-09
Add `wk` unit for weeks

jkrumbiegel:
While certainly not the most used time unit, we have a need for `wk` in our project because Unitful also defines this by default and we want to ascertain compatibility. I hope you don't mind adding this one?

### DIFF
--- a/src/units.jl
+++ b/src/units.jl
@@ -68,7 +68,7 @@ end
     kg,
 )
 @doc(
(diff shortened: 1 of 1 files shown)


jkrumbiegel:
Ok by chance I did this right at the same time as https://github.com/SymbolicML/DynamicQuantities.jl/pull/81 so I'm going to close this one again

MilesCranmer:
Quite a coincidence to have PRs within 9 minutes of eachother adding the same thing! 🤯

PR #84, opened 2023-11-10
Fix numeric promotion rules

MilesCranmer:
@gaurav-arya this should address your comment in #78.

Now you can do 
```julia
[3.0, 1u"m"]
```
and it will be an array of `Quantity{Float64,...}`.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -43,6 +43,30 @@ end
 function Base.promote_rule(::Type{<:Quantity{T1,D1}}, ::Type{<:Quantity{T2,D2}}) where {T1,T2,D1,D2}
     return Quantity{promote_type(T1,T2),promote_type(D1,D2)}
 end
+
+# Define promotion rules for all basic numeric types, individually.
+# We don't want to define an opinionated promotion on <:Number,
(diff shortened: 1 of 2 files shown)


gaurav-arya (approved):
I'm not the biggest fan of having to manually list the numeric types here, as it's not extensible for other packages defining numeric-like types. 

I think it deserves a bit of thought whether there's a slicker way to do this that still doesn't run into the issue you mentioned with other number packages. But I can't think about this at least for the next day or two, so if you'd like to merge and improve this looks good to me:)

MilesCranmer:
Sounds good. Yeah I just have no idea how to figure out the ambiguity for things like
```julia
julia> using DynamicQuantities, Measurements

julia> [0.3±0.1, 3u"km/s"]
2-element Vector{Number}:
        0.3 ± 0.1
 3000.0 m s⁻¹
```
i.e., should this be a `Measurement{Quantity}` or a `Quantity{Measurement}`? If we define a promotion rule over all of `Number` we would effectively be choosing `Quantity{Measurement}` which seems like undesired behavior. Wdyt?

PR #85, opened 2023-11-13
Add AbstractRealQuantity

MilesCranmer:
This adds a new `AbstractRealQuantity <: Real` and a concrete type `RealQuantity`. Users can wrap quantities with `RealQuantity(0.3u"km/s")` to pass to functions that require real input. But the default would still be `Quantity <: Number` for reasons discussed [here](https://github.com/SymbolicML/DynamicQuantities.jl/pull/85#issuecomment-1820307355).

I also introduce a new function `promote_quantity` which one can overload as a sort of custom type hierarchy, so that `RealQuantity -> Quantity` if it interacts with a non-real, and `{RealQuantity,Quantity} -> GenericQuantity` if they interact with non-numeric.

~~However this is a bit of a mess right now due to promotion complexities, making `(0.3+2im)u"km/s"` promote to a `Complex{<:Quantity}` instead of a `Quantity`... Anybody interested in having a go?~~ Fixed!

### DIFF
--- a/src/arrays.jl
+++ b/src/arrays.jl
@@ -15,7 +15,7 @@ and so can be used in most places where a normal array would be used, including 
 # Constructors
 
 - `QuantityArray(v::AbstractArray, d::AbstractDimensions)`: Create a `QuantityArray` with value `v` and dimensions `d`,
-  using `Quantity` if the eltype of `v` is numeric, and `GenericQuantity` otherwise.
+  using `RealQuantity` if the eltype of `v` is real, `Quantity` if it is numeric, and `GenericQuantity` otherwise.
 - `QuantityArray(v::AbstractArray{<:Number}, q::AbstractQuantity)`: Create a `QuantityArray` with value `v` and dimensions inferred
    with `dimension(q)`. This is so that you can easily create an array with the units module, like so:
    ```julia
@@ -52,7 +52,6 @@ struct QuantityArray{T,N,D<:AbstractDimensions,Q<:UnionAbstractQuantity{T,D},V<:
     end
 end
 
-# Construct with a Quantity (easier, as you can use the units):
 QuantityArray(v::AbstractArray; kws...) = QuantityArray(v, DEFAULT_DIM_TYPE(; kws...))
 for (type, base_type, default_type) in ABSTRACT_QUANTITY_TYPES
     @eval begin
--- a/src/arrays.jl
+++ b/src/arrays.jl
@@ -15,7 +15,7 @@ and so can be used in most places where a normal array would be used, including 
 # Constructors
 
 - `QuantityArray(v::AbstractArray, d::AbstractDimensions)`: Create a `QuantityArray` with value `v` and dimensions `d`,
-  using `RealQuantity` if the eltype of `v` is real, `Quantity` if it is numeric, and `GenericQuantity` otherwise.
+  using `Quantity` if it is numeric, and `GenericQuantity` otherwise.
 - `QuantityArray(v::AbstractArray{<:Number}, q::AbstractQuantity)`: Create a `QuantityArray` with value `v` and dimensions inferred
    with `dimension(q)`. This is so that you can easily create an array with the units module, like so:
    ```julia
@@ -54,9 +54,11 @@ end
 
 QuantityArray(v::AbstractArray; kws...) = QuantityArray(v, DEFAULT_DIM_TYPE(; kws...))
 for (type, base_type, default_type) in ABSTRACT_QUANTITY_TYPES
-    @eval begin
-        QuantityArray(v::AbstractArray{<:$base_type}, q::$type) = QuantityArray(v .* ustrip(q), dimension(q), typeof(q))
-        QuantityArray(v::AbstractArray{<:$base_type}, d::AbstractDimensions) = QuantityArray(v, d, $default_type)
+    @eval QuantityArray(v::AbstractArray{<:$base_type}, q::$type) = QuantityArray(v .* ustrip(q), dimension(q), typeof(q))
+
+    # Only define defaults for Quantity and GenericQuantity. Other types, the user needs to declare explicitly.
+    if type in (AbstractQuantity, AbstractGenericQuantity)
+        @eval QuantityArray(v::AbstractArray{<:$base_type}, d::AbstractDimensions) = QuantityArray(v, d, $default_type)
     end
 end
 QuantityArray(v::QA) where {Q<:UnionAbstractQuantity,QA<:AbstractArray{Q}} =
--- a/src/arrays.jl
+++ b/src/arrays.jl
@@ -15,7 +15,7 @@ and so can be used in most places where a normal array would be used, including 
 # Constructors
 
 - `QuantityArray(v::AbstractArray, d::AbstractDimensions)`: Create a `QuantityArray` with value `v` and dimensions `d`,
-  using `Quantity` if it is numeric, and `GenericQuantity` otherwise.
+  using `Quantity` if the eltype of `v` is numeric, and `GenericQuantity` otherwise.
 - `QuantityArray(v::AbstractArray{<:Number}, q::AbstractQuantity)`: Create a `QuantityArray` with value `v` and dimensions inferred
    with `dimension(q)`. This is so that you can easily create an array with the units module, like so:
    ```julia
--- a/src/disambiguities.jl
+++ b/src/disambiguities.jl
@@ -47,6 +47,18 @@ end
 # Assorted calls found by Aqua: ################################################
 ################################################################################
 
+function Complex(q::AbstractRealQuantity)
+    @assert iszero(dimension(q)) "$(typeof(q)): $(q) has dimensions! Use `ustrip` instead."
+    return Complex(ustrip(q))
+end
+function Complex{T}(q::AbstractRealQuantity) where {T<:Real}
+    @assert iszero(dimension(q)) "$(typeof(q)): $(q) has dimensions! Use `ustrip` instead."
+    return Complex{T}(ustrip(q))
+end
+function Bool(q::AbstractRealQuantity)
+    @assert iszero(dimension(q)) "$(typeof(q)): $(q) has dimensions! Use `ustrip` instead."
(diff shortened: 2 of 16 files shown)


MilesCranmer:
Actually I think we shouldn't make all default quantities `Real`. We can just have these as quantities that someone can use if they want that particular dispatch.

MilesCranmer:
@gaurav-arya I'm wondering if we should just make `Quantity <: Real`, and `NumberQuantity <: Number`. Having the default numeric type be `RealQuantity` rather than `Quantity` could be a bit awkward...

src/types.jl
```diff
@@ -189,10 +224,15 @@ for (type, base_type, _) in ABSTRACT_QUANTITY_TYPES
...
 
-new_dimensions(::Type{D}, dims...) where {D<:AbstractDimensions} = constructorof(D)(dims...)
-new_quantity(::Type{Q}, l, r) where {Q<:UnionAbstractQuantity} = constructorof(Q)(l, r)
+@inline function new_dimensions(::Type{D}, dims...) where {D<:AbstractDimensions}
+    return constructorof(D)(dims...)
+end
+@inline function new_quantity(::Type{Q}, val, dims) where {Q<:UnionAbstractQuantity}
+    Qout = promote_quantity(Q, typeof(val))
+    return constructorof(Qout)(val, dims)
+end
```
MilesCranmer, line 235: Now `new_quantity` will automatically promote from `RealQuantity -> Quantity` when needed. (Just like how the value type can also change.)

This hierarchy is defined by `promote_quantity(::Type{<:UnionAbstractQuantity}, ::Type{<:Any})` which is given for the default concrete types, but which the user can obviously overload.

gaurav-arya:
> Actually I think we shouldn't make all default quantities Real. We can just have these as quantities that someone can use if they want that particular dispatch.

Why not choose the most specific type given the value type? E.g. `1.0 * u"m"` would be `<: Real` but `(1.0 + 1.0 * im) * u"m"` would be `<: Number` for now, etc. Is there a disadvantage, i.e. a reason the user would have preferred `Quantity{<:Real}` as opposed to `RealQuantity{<:Real}`? 

In this scenario, I wonder if we even need a single `DEFAULT_QUANTITY_TYPE`? Instead, we have a default mapping from value type -> quantity type?

MilesCranmer:
> Why not choose the most specific type given the value type? E.g. `1.0 * u"m"` would be `<: Real` but `(1.0 + 1.0 * im) * u"m"` would be `<: Number` for now, etc. Is there a disadvantage, i.e. a reason the user would have preferred `Quantity{<:Real}` as opposed to `RealQuantity{<:Real}`?

Yeah good point. The current behavior in this PR is actually `(1.0 * u"m")::Real` and `((1.0 + 1.0 * im) * u"m")::Number` already.

> In this scenario, I wonder if we even need a single `DEFAULT_QUANTITY_TYPE`? Instead, we have a default mapping from value type -> quantity type?

Maybe do you mean like #66? Although I don't think we necessarily need an `AbstractFloatQuantity` yet, unless a specific need arises, since it will increase load time. But `AbstractRealQuantity` seems necessary for many packages.

MilesCranmer:
Cool, coverage is now back to 100%. Feel free to take a look when you want. I'll keep `RealQuantity` as the default, but am still open to changing the name to be `Quantity` and `NumberQuantity` rather than `RealQuantity` and `Quantity`, respectively.

src/utils.jl
```diff
@@ -118,47 +159,85 @@ Base.keys(q::UnionAbstractQuantity) = keys(ustrip(q))
...
-end
-function Base.isapprox(l::UnionAbstractQuantity, r::Number; kws...)
-    iszero(dimension(l)) || throw(DimensionError(l, r))
-    return isapprox(ustrip(l), r; kws...)
-end
-Base.iszero(d::AbstractDimensions) = all_dimensions(iszero, d)
-function Base.:(==)(l::UnionAbstractQuantity, r::UnionAbstractQuantity)
-    l, r = promote_except_value(l, r)
-    ustrip(l) == ustrip(r) && dimension(l) == dimension(r)
+for op in (:(<=), :(<), :(>=), :(>), :isless),
```
MilesCranmer, line 162: For the record: I would have thought these operators would be automatically defined by `isless`, since that's what the docstring says. But it turns out that some packages will specifically check whether a method is defined on these (Turing.jl being an example).

MilesCranmer:
Also:

> Why not choose the most specific type given the value type? E.g. `1.0 * u"m"` would be `<: Real` but `(1.0 + 1.0 * im) * u"m"` would be `<: Number` for now, etc. Is there a disadvantage, i.e. a reason the user would have preferred `Quantity{<:Real}` as opposed to `RealQuantity{<:Real}`?

Is this even semantically correct? One could indeed argue that 0.5 meters is a type of number... But is it a _real_? 

Let me give an example. Would you agree with the statement

$(0.5 \text{ km} \text{ s}^{-1}) \in \mathbb{R}$

? I think it kind of seems incorrect... In which case it might not be the best option to set the default type as `::Real`.

Whereas `Number` is much more abstract and I could totally say that a physical quantity is a type of number.

This semantic ambiguity leads me to want to keep `Quantity <: Number` as the default, but allow `RealQuantity` as a wrapper for users who know what they are doing. Wdyt?

src/arrays.jl
```diff
@@ -52,12 +52,13 @@ struct QuantityArray{T,N,D<:AbstractDimensions,Q<:UnionAbstractQuantity{T,D},V<:
...
 QuantityArray(v::AbstractArray; kws...) = QuantityArray(v, DEFAULT_DIM_TYPE(; kws...))
 for (type, base_type, default_type) in ABSTRACT_QUANTITY_TYPES
-    @eval begin
-        QuantityArray(v::AbstractArray{<:$base_type}, q::$type) = QuantityArray(v .* ustrip(q), dimension(q), typeof(q))
-        QuantityArray(v::AbstractArray{<:$base_type}, d::AbstractDimensions) = QuantityArray(v, d, $default_type)
+    @eval QuantityArray(v::AbstractArray{<:$base_type}, q::$type) = QuantityArray(v .* ustrip(q), dimension(q), typeof(q))
+
+    # Only define defaults for Quantity and GenericQuantity. Other types, the user needs to declare explicitly.
+    if type in (AbstractQuantity, AbstractGenericQuantity)
+        @eval QuantityArray(v::AbstractArray{<:$base_type}, d::AbstractDimensions) = QuantityArray(v, d, $default_type)
```
gaurav-arya, line 61: Is there are a reason to only define this for the concrete types here (with `$default_type`), rather than the abstract types?

src/disambiguities.jl
```diff
@@ -1,14 +1,100 @@
...
+    return promote_type(Rational{eltype(F)}, T)
+end
+
+################################################################################
+# Assorted calls found by Aqua: ################################################
+################################################################################
+
+for type in (Signed, Float64, Float32, Rational), op in (:flipsign, :copysign)
+    @eval function Base.$(op)(x::$type, y::AbstractRealQuantity)
+        return $(op)(x, ustrip(y))
```
gaurav-arya, line 52:
```suggestion
        return Base.$(op)(x, ustrip(y))
```
for consistency?

src/disambiguities.jl
```diff
@@ -1,14 +1,100 @@
...
+end
+function Base.:/(l::AbstractRealQuantity, r::Complex)
+    new_quantity(typeof(l), ustrip(l) / r, dimension(l))
+end
+function Base.:*(l::Complex{Bool}, r::AbstractRealQuantity)
+    return new_quantity(typeof(r), l * ustrip(r), dimension(r))
+end
+function Base.:*(l::AbstractRealQuantity, r::Complex{Bool})
+    return new_quantity(typeof(l), ustrip(l) * r, dimension(l))
+end
```
gaurav-arya, line 72: This block of methods might be more readable as a metaprogramming loop? E.g. over the three pairs of (method, type)?

src/disambiguities.jl
gaurav-arya: Nothing to address in this PR, but while I'm here: I think I pefer the style while `op` iterates over the function objects directly (e.g. `op in (Base.flipsign, Base.copysign)`). Dispatch is still possible without the symbol, e.g. see https://github.com/gaurav-arya/StochasticAD.jl/blob/3638b3b36450451e07f890c033fb253199ca83b5/src/general_rules.jl#L200-L201.

But the iteration over symbol approach like here indeed is very common. It just seems a bit less elegant, e.g. what if you wanted `op` to iterate over multiple methods owned by different modular?

src/math.jl
```diff
@@ -1,4 +1,5 @@
 for (type, base_type, _) in ABSTRACT_QUANTITY_TYPES
+    div_base_type = type == AbstractGenericQuantity ? Number : base_type
```
gaurav-arya, line 2:
```suggestion
    div_base_type = type === AbstractGenericQuantity ? Number : base_type
```
Tiny nit: I think convention is to use `===` for comparing types. (Maybe because it's a little safer since `==` is technically overloadable even on types?)

gaurav-arya:
```suggestion
    div_base_type = type <: Number ? base_type : Number
```
Moreover, could this check be more elegantly written like this, or something similar?

gaurav-arya (commented):
Left some small comments! FYI this has been a bit of a "drive by" review, I'm a little swamped with my semester. But the overall idea I'm totally in support of.

Regarding aggressively defaulting to `<: Real`: point taken that it might require more thought, can easily be a separate PR (which we could arguably say is non-breaking, while going the other way around and reverting the aggressive default would deifnitely be breaking). One counterpoint is that `ForwardDiff.Dual <: Real` -- I would be interested in seeing if there's any practical use case where a quantity breaks the expected semantics of `Real`.

Edit: one more thing -- conditional on the default type definition remaining, I do agree with the current PR decision to keep it as `Number` rather than `Real`.

src/disambiguities.jl
MilesCranmer: Thanks. I'm open to it, I guess we could just do this in another PR?

MilesCranmer:
Thanks!

PR #91, opened 2023-11-23
Add file that generates logo

MilesCranmer:
@jkrumbiegel I made you a coauthor on the commit – hope that works for you. Thanks again for the nice logo!!

I also made some final tweaks in illustrator (smoothed corners and drop shadow):

dyn_q_logo

We can definitely keep updating it though.

### DIFF
--- a/docs/logo.jl
+++ b/docs/logo.jl
@@ -1,0 +1,56 @@
+using Luxor
+using Luxor.Colors
+
+image_width = 500
+
+@svg begin
+    circle_radius = 0.85 * image_width / 2
+    scale(circle_radius, circle_radius)
+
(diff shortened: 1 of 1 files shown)


MilesCranmer:
Hm, seems like the best way to actually put this on the README is just drag and drop it to the markdown. So will merge first and then drag the logo over.

jkrumbiegel:
Oh cool, yeah sorry I was super busy with work and the Makie 0.20 release.. I would maybe still replace the `ks` field with `h` or `hr` because `ks` is so unusual, but it doesn't really matter either way :)

MilesCranmer:
No worries! Yeah good point... will get to that eventually...

PR #98, opened 2023-12-03
More unitful compatibility

MilesCranmer:


src/DynamicQuantities.jl
```diff
@@ -28,6 +28,15 @@ import .Units
 import .Constants
 import .UnitsParse: uparse, @u_str
 
+using .Units: UNIT_SYMBOLS
+_units_import_expr = :(using .Units: m, g)
```
MilesCranmer, line 32: Put inside let

PR #99, opened 2023-12-03
More units

Mo-Gul:
I'd argue that `mbar` is also a quite common unit. At least much more than `dL`. And because I also think that `cL` is at least as common as `dL` that should be added as well.
Feel free to cherry-pick.

### DIFF
--- a/src/units.jl
+++ b/src/units.jl
@@ -187,10 +187,10 @@ end
 ## Pressure
 @register_unit bar 100 * kPa
 
-@add_prefixes bar ()
+@add_prefixes bar (m,)
 
 @doc(
-    "Pressure in bars.",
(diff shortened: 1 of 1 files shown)


src/units.jl
```diff
@@ -177,20 +177,20 @@ end
...
-    "Volume in liters. Available variants: `mL`, `dL`.",
+    "Volume in liters. Available variants: `mL`, `cL`, `dL`.",
     L,
 )
 
 ## Pressure
 @register_unit bar 100 * kPa
 
-@add_prefixes bar ()
+@add_prefixes bar (m)
```
MilesCranmer, line 190:
```suggestion
@add_prefixes bar (m,)
```

PR #101, opened 2023-12-03
Refactor symbolic units module, make all symbols available at import

### LINKED ISSUE #51: Apparent Issue with Precompilation for Dependent Packages

MilesCranmer:
Previously, units were created the first time the user used `sym_uparse`, because otherwise it required creating ~150 arrays at import inside an `@eval`. This also means that you could not precompile unit calculations done with symbolic units.

With this change, we now eagerly create symbolic units as immutable objects. For example, `h` would be created with

```julia
Quantity(1.0, SymbolicDimensionsSingleton(:h))
```

and this would represented in memory as the lightweight `(1.0, 85)`, where `85` is the index of `hours` in the constant tuple of units and quantities.
 
- This creates the abstract type `AbstractSymbolicDimensions <: AbstractDimensions` and defines several of the symbolic dimensions methods on it. 
- This uses a new type `SymbolicDimensionsSingleton` that stores a single integer corresponding to the index of the unit/constant in a constant tuple.
- This also exports modules `SymbolicUnits` and `SymbolicConstants`, so you can do things like:

```julia

julia> using DynamicQuantities.SymbolicUnits: km, h

julia> km |> typeof
Quantity{Float64, SymbolicDimensionsSingleton{DynamicQuantities.FixedRational{Int32, 25200}}}

julia> km/h |> typeof
Quantity{Float64, SymbolicDimensions{DynamicQuantities.FixedRational{Int32, 25200}}}

julia> km/h |> uexpand
0.2777777777777778 m s⁻¹

julia> km/h |> uexpand |> uconvert(SymbolicConstants.pc / SymbolicUnits.Myr)
0.00028408671251269307 Myr⁻¹ pc
```

@devmotion would you be up for reviewing this?

---

cc @gaurav-arya 

Fixes #51 by cadojo (and prevents the need for #58)

### DIFF
--- a/src/symbolic_dimensions.jl
+++ b/src/symbolic_dimensions.jl
@@ -53,8 +53,6 @@ be used for constructing symbolic units and constants without needing to allocat
 """
 struct SymbolicDimensionsSingleton{R} <: AbstractSymbolicDimensions{R}
     dim::INDEX_TYPE
-
-    SymbolicDimensionsSingleton(dim::INDEX_TYPE, ::Type{_R}) where {_R} = new{_R}(dim)
 end
 
 # Access:
@@ -93,7 +91,7 @@ end
 function SymbolicDimensionsSingleton{R}(s::Symbol) where {R}
     i = get(ALL_MAPPING, s, INDEX_TYPE(0))
     iszero(i) && error("$s is not available as a symbol in `SymbolicDimensionsSingleton`. Symbols available: $(ALL_SYMBOLS).")
-    return SymbolicDimensionsSingleton(i, R)
+    return SymbolicDimensionsSingleton{R}(i)
 end
 
 # Traits:
@@ -199,7 +197,7 @@ a function equivalent to `q -> uconvert(qout, q)`.
 uconvert(qout::UnionAbstractQuantity{<:Any,<:AbstractSymbolicDimensions}) = Base.Fix1(uconvert, qout)
 
 Base.copy(d::SymbolicDimensions) = SymbolicDimensions(copy(nzdims(d)), copy(nzvals(d)))
-Base.copy(d::SymbolicDimensionsSingleton) = SymbolicDimensionsSingleton(getfield(d, :dim), eltype(d))
+Base.copy(d::SymbolicDimensionsSingleton) = constructorof(d)(getfield(d, :dim))
 
 function Base.:(==)(l::AbstractSymbolicDimensions, r::AbstractSymbolicDimensions)
     nzdims_l = nzdims(l)
--- a/src/symbolic_dimensions.jl
+++ b/src/symbolic_dimensions.jl
@@ -59,11 +59,11 @@ end
 
 # Access:
 function Base.getproperty(d::SymbolicDimensions{R}, s::Symbol) where {R}
-    _nzdims = nzdims(d)
+    nzdims = DynamicQuantities.nzdims(d)
     i = get(ALL_MAPPING, s, INDEX_TYPE(0))
     iszero(i) && error("$s is not available as a symbol in `SymbolicDimensions`. Symbols available: $(ALL_SYMBOLS).")
-    ii = searchsortedfirst(_nzdims, i)
-    if ii <= length(_nzdims) && _nzdims[ii] == i
+    ii = searchsortedfirst(nzdims, i)
+    if ii <= length(nzdims) && nzdims[ii] == i
         return nzvals(d)[ii]
     else
         return zero(R)
@@ -105,18 +105,18 @@ constructorof(::Type{<:SymbolicDimensionsSingleton{R}}) where {R} = SymbolicDime
 with_type_parameters(::Type{<:SymbolicDimensions}, ::Type{R}) where {R} = SymbolicDimensions{R}
 with_type_parameters(::Type{<:SymbolicDimensionsSingleton}, ::Type{R}) where {R} = SymbolicDimensionsSingleton{R}
 nzdims(d::SymbolicDimensions) = getfield(d, :nzdims)
-nzdims(d::SymbolicDimensionsSingleton) = [getfield(d, :dim)]
+nzdims(d::SymbolicDimensionsSingleton) = (getfield(d, :dim),)
 nzvals(d::SymbolicDimensions) = getfield(d, :nzvals)
-nzvals(::SymbolicDimensionsSingleton{R}) where {R} = [one(R)]
+nzvals(::SymbolicDimensionsSingleton{R}) where {R} = (one(R),)
 Base.eltype(::AbstractSymbolicDimensions{R}) where {R} = R
 Base.eltype(::Type{<:AbstractSymbolicDimensions{R}}) where {R} = R
 
 # Conversion:
 function SymbolicDimensions(d::SymbolicDimensionsSingleton{R}) where {R}
-    return SymbolicDimensions{R}(nzdims(d), nzvals(d))
+    return SymbolicDimensions{R}([nzdims(d)...], [nzvals(d)...])
 end
 function SymbolicDimensions{R}(d::SymbolicDimensionsSingleton) where {R}
-    return SymbolicDimensions{R}(nzdims(d), nzvals(d))
+    return SymbolicDimensions{R}([nzdims(d)...], [nzvals(d)...])
 end
 Base.convert(::Type{SymbolicDimensions}, d::SymbolicDimensionsSingleton) = SymbolicDimensions(d)
 Base.convert(::Type{SymbolicDimensions{R}}, d::SymbolicDimensionsSingleton) where {R} = SymbolicDimensions{R}(d)
--- a/src/symbolic_dimensions.jl
+++ b/src/symbolic_dimensions.jl
@@ -4,21 +4,7 @@ import .Constants: CONSTANT_SYMBOLS, CONSTANT_MAPPING, CONSTANT_VALUES
 
 const SYMBOL_CONFLICTS = intersect(UNIT_SYMBOLS, CONSTANT_SYMBOLS)
 
-function disambiguate_symbol(s)
-    if s in SYMBOL_CONFLICTS
-        return Symbol(string(s) * "_constant")
-    else
-        return s
-    end
-end
-function reambiguate_symbol(s)
-    str_s = string(s)
-    if endswith(str_s, "_constant")
-        return Symbol(str_s[1:end-9])
-    else
-        return s
-    end
-end
+disambiguate_symbol(s) = s in SYMBOL_CONFLICTS ? Symbol(string(s) * "_constant") : s
 
 const INDEX_TYPE = UInt8
 # Prefer units over constants:
--- a/src/symbolic_dimensions.jl
+++ b/src/symbolic_dimensions.jl
@@ -4,7 +4,7 @@ import .Constants: CONSTANT_SYMBOLS, CONSTANT_MAPPING, CONSTANT_VALUES
 
 const SYMBOL_CONFLICTS = intersect(UNIT_SYMBOLS, CONSTANT_SYMBOLS)
 
-disambiguate_symbol(s) = s in SYMBOL_CONFLICTS ? Symbol(string(s) * "_constant") : s
+disambiguate_symbol(s) = s in SYMBOL_CONFLICTS ? Symbol(s, :_constant) : s
 
 const INDEX_TYPE = UInt8
 # Prefer units over constants:
@@ -15,7 +15,7 @@ const ALL_SYMBOLS = (
     disambiguate_symbol.(CONSTANT_SYMBOLS)...
 )
 const ALL_VALUES = (UNIT_VALUES..., CONSTANT_VALUES...)
-const ALL_MAPPING = NamedTuple([s => INDEX_TYPE(i) for (i, s) in enumerate(ALL_SYMBOLS)])
+const ALL_MAPPING = NamedTuple{ALL_SYMBOLS}(INDEX_TYPE(1):INDEX_TYPE(length(ALL_SYMBOLS)))
 
 """
     AbstractSymbolicDimensions{R} <: AbstractDimensions{R}
--- a/src/symbolic_dimensions.jl
+++ b/src/symbolic_dimensions.jl
@@ -1,7 +1,24 @@
 import .Units: UNIT_SYMBOLS, UNIT_MAPPING, UNIT_VALUES
 import .Constants: CONSTANT_SYMBOLS, CONSTANT_MAPPING, CONSTANT_VALUES
 
+
 const SYMBOL_CONFLICTS = intersect(UNIT_SYMBOLS, CONSTANT_SYMBOLS)
+
+function disambiguate_symbol(s)
+    if s in SYMBOL_CONFLICTS
+        return Symbol(string(s) * "_constant")
+    else
+        return s
+    end
+end
+function reambiguate_symbol(s)
+    str_s = string(s)
+    if endswith(str_s, "_constant")
+        return Symbol(str_s[1:end-9])
+    else
+        return s
+    end
+end
 
 const INDEX_TYPE = UInt8
 # Prefer units over constants:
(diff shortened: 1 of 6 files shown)


gaurav-arya:
It'll take me a bit of time since I'm not quite done with school yet, but I'll give this a review:)

src/symbolic_dimensions.jl
```diff
@@ -1,26 +1,32 @@
 import .Units: UNIT_SYMBOLS, UNIT_MAPPING, UNIT_VALUES
 import .Constants: CONSTANT_SYMBOLS, CONSTANT_MAPPING, CONSTANT_VALUES
 
+
 const SYMBOL_CONFLICTS = intersect(UNIT_SYMBOLS, CONSTANT_SYMBOLS)
 
+disambiguate_symbol(s) = s in SYMBOL_CONFLICTS ? Symbol(string(s) * "_constant") : s
```
devmotion, line 7: I think the following is sufficient:
```suggestion
disambiguate_symbol(s) = s in SYMBOL_CONFLICTS ? Symbol(s, :_constant) : s
```

devmotion: Maybe it would be better to avoid these conflicts in the first place though, e.g. by keeping separate containers or using a `_constant` suffix for all constants.

src/symbolic_dimensions.jl
```diff
@@ -1,26 +1,32 @@
...
-const ALL_VALUES = vcat(
-    UNIT_VALUES...,
-    (
-        v
-        for (k, v) in zip(CONSTANT_SYMBOLS, CONSTANT_VALUES)
-        if k ∉ SYMBOL_CONFLICTS
-    )...
+    disambiguate_symbol.(CONSTANT_SYMBOLS)...
 )
+const ALL_VALUES = (UNIT_VALUES..., CONSTANT_VALUES...)
```
devmotion, line 17: BTW is there a particular reason for using a tuple for `ALL_SYMBOL` instead of an array? And is `ALL_VALUES` actually needed? Every entry should be available as `getproperty(Units, symbol::Symbol)` and `getproperty(Constants, symbol::Symbol)`?

src/symbolic_dimensions.jl
```diff
@@ -1,26 +1,32 @@
...
 const ALL_MAPPING = NamedTuple([s => INDEX_TYPE(i) for (i, s) in enumerate(ALL_SYMBOLS)])
```
devmotion, line 18: Maybe just
```suggestion
const ALL_MAPPING = NamedTuple{ALL_SYMBOLS}(INDEX_TYPE(1):INDEX_TYPE(length(ALL_SYMBOLS)))
```

src/symbolic_dimensions.jl
```diff
@@ -34,31 +40,47 @@ You can convert a quantity using `SymbolicDimensions` as its dimensions
...
+"""
+    SymbolicDimensionsSingleton{R} <: AbstractSymbolicDimensions{R}
+
+This special symbolic dimensions types stores a single unit or constant, and can
+be used for constructing symbolic units and constants without needing to allocate mutable storage.
+"""
+struct SymbolicDimensionsSingleton{R} <: AbstractSymbolicDimensions{R}
+    dim::INDEX_TYPE
+
+    SymbolicDimensionsSingleton(dim::INDEX_TYPE, ::Type{_R}) where {_R} = new{_R}(dim)
```
devmotion, line 57: I still think these type arguments are very non-idiomatic. The more standard approach would be just
```suggestion
    SymbolicDimensionsSingleton{R}(dim::INDEX_TYPE) where {R} = new{R}(dim)
```
(which is already the default)

src/symbolic_dimensions.jl
```diff
@@ -34,31 +40,47 @@ You can convert a quantity using `SymbolicDimensions` as its dimensions
...
+end
+
+# Access:
 function Base.getproperty(d::SymbolicDimensions{R}, s::Symbol) where {R}
-    nzdims = getfield(d, :nzdims)
+    _nzdims = nzdims(d)
```
devmotion, line 62: You could minimize the diff by using
```suggestion
    nzdims = DynamicQuantities.nzdims(d)
```

src/symbolic_dimensions.jl
```diff
@@ -68,7 +90,36 @@ function SymbolicDimensions{R}(; kws...) where {R}
...
+# Traits:
+dimension_names(::Type{<:AbstractSymbolicDimensions}) = ALL_SYMBOLS
+Base.propertynames(::AbstractSymbolicDimensions) = ALL_SYMBOLS
+Base.getindex(d::AbstractSymbolicDimensions, k::Symbol) = getproperty(d, k)
+constructorof(::Type{<:SymbolicDimensions}) = SymbolicDimensions
+constructorof(::Type{<:SymbolicDimensionsSingleton{R}}) where {R} = SymbolicDimensionsSingleton{R}
+with_type_parameters(::Type{<:SymbolicDimensions}, ::Type{R}) where {R} = SymbolicDimensions{R}
+with_type_parameters(::Type{<:SymbolicDimensionsSingleton}, ::Type{R}) where {R} = SymbolicDimensionsSingleton{R}
+nzdims(d::SymbolicDimensions) = getfield(d, :nzdims)
+nzdims(d::SymbolicDimensionsSingleton) = [getfield(d, :dim)]
```
devmotion, line 108: Could this be a tuple for `SymbolicDimensionsSingleton`? These allocations seem undesirable.
```suggestion
nzdims(d::SymbolicDimensionsSingleton) = (getfield(d, :dim),)
```

devmotion: Maybe it would be easier to just define all methods for the `...Singleton` type directly, without going through something like `nzdims`.

src/symbolic_dimensions.jl
```diff
@@ -86,12 +137,12 @@ for (type, _, _) in ABSTRACT_QUANTITY_TYPES
...
-        function Base.convert(::Type{Q}, q::UnionAbstractQuantity{<:Any,<:SymbolicDimensions}) where {T,D<:Dimensions,Q<:$type{T,D}}
+        function Base.convert(::Type{Q}, q::UnionAbstractQuantity{<:Any,<:AbstractSymbolicDimensions}) where {T,D<:Dimensions,Q<:$type{T,D}}
             result = constructorof(Q)(convert(T, ustrip(q)), D())
             d = dimension(q)
-            for (idx, value) in zip(getfield(d, :nzdims), getfield(d, :nzvals))
+            for (idx, value) in zip(nzdims(d), nzvals(d))
                 if !iszero(value)
-                    result = result * convert(with_type_parameters(Q, T, D), ALL_VALUES[idx]) ^ value
+                    result *= convert(with_type_parameters(Q, T, D), ALL_VALUES[idx] ^ value)
                 end
```
devmotion, line 146: This could be simplified for the `...Singleton` type.

src/symbolic_dimensions.jl
```diff
@@ -34,31 +40,47 @@ You can convert a quantity using `SymbolicDimensions` as its dimensions
...
     nzdims::Vector{INDEX_TYPE}
     nzvals::Vector{R}
 end
 
-@inline dimension_names(::Type{<:SymbolicDimensions}) = ALL_SYMBOLS
+"""
+    SymbolicDimensionsSingleton{R} <: AbstractSymbolicDimensions{R}
+
+This special symbolic dimensions types stores a single unit or constant, and can
+be used for constructing symbolic units and constants without needing to allocate mutable storage.
```
devmotion, line 52: Given also the comments below about allocations in `nzdims` etc. I wonder if it would be easier to allow non-mutable containers (`Tuple`s?) in `SymbolicDimensions`.

src/symbolic_dimensions.jl
MilesCranmer: Good tip, I didn't know that syntax!

> Maybe it would be better to avoid these conflicts in the first place though, e.g. by keeping separate containers or using a `_constant` suffix for all constants.

Oh, the `_constant` here is just for printing. The `h` symbol still works as you might expect, in a separate module:

```julia
julia> using DynamicQuantities.SymbolicUnits: h as h1

julia> using DynamicQuantities.SymbolicConstants: h as h2

julia> h1
1.0 h

julia> h2
1.0 h_constant
```

Then in string form you would write `us"Constants.h"`

src/symbolic_dimensions.jl
MilesCranmer: I admit I don't know where the benefits of tuple-ifying constants end. Where would you draw the line? I'm assuming small tuples help with extra allocations. Are there downsides to using large tuples, if they are declared constants?

MilesCranmer: > Every entry should be available as `getproperty(Units, symbol::Symbol)` and `getproperty(Constants, symbol::Symbol)`?

Nice!! I'll do that.

MilesCranmer: Wait, nevermind re: this^. The combined tuples here simplify things as I don't have to check whether the symbol is in Units or Constants, I just look up the index.

src/symbolic_dimensions.jl
MilesCranmer: I'm afraid this would be a lot more complexity than it would save as you would have to account for additional type parameters in `AbstractDimensions` types which I haven't tested yet. But it's only a single edge case for right now. Maybe if the need arises we could take a look.

MilesCranmer: Actually, wait... This could actually be very cool, because (and maybe this is why you suggested it?), this would technically be a way to have full static-typing equivalence to Unitful.

src/symbolic_dimensions.jl
MilesCranmer: Just checking – the compiler can probably simplify this itself now that I made the `nzvals` and `nzdims` tuples, right?

src/symbolic_dimensions.jl
MilesCranmer: Added! Yeah I'm not sure why I had written that initially...

MilesCranmer:
Thanks very much for the suggestions @devmotion. Everything implemented!

src/symbolic_dimensions.jl
```diff
@@ -263,72 +317,60 @@ function map_dimensions(op::O, l::SymbolicDimensions{L}, r::SymbolicDimensions{R
...
-        UNIT_SYMBOLS_EXIST[] || lock(UNIT_SYMBOLS_LOCK) do
-            UNIT_SYMBOLS_EXIST[] && return nothing
-            for unit in UNIT_SYMBOLS
-                @eval const $unit = constructorof(DEFAULT_SYMBOLIC_QUANTITY_TYPE)(DEFAULT_VALUE_TYPE(1.0), SymbolicDimensions{DEFAULT_DIM_BASE_TYPE}; $(unit)=1)
-            end
-            UNIT_SYMBOLS_EXIST[] = true
-        end
-        return nothing
+    for unit in UNIT_SYMBOLS
+        @eval const $unit = constructorof(DEFAULT_SYMBOLIC_QUANTITY_TYPE)(
```
MilesCranmer, line 367: Not sure the benefits, but I could write this as a single tuple with `@eval const $(UNIT_SYMBOLS) = `. Wdyt @devmotion?

MilesCranmer:
Let me know if there are any other comments! Otherwise I will try to merge this soon.

PR #106, opened 2024-01-04
Improve `@u_str` implementation to fix downstream precompilation

### LINKED ISSUE #105: u"Ω" errors in precompilation
MWE: write the literal`u"Ω"` in a package that precompiles.

```julia
caused by: LoadError: Evaluation into the closed module `UnitsParse` breaks incremental compilation because the side effects will not be permanent. This is likely due to some other module mutating `UnitsParse` with `eval` during precompilation - don't do this.
Stacktrace:
  [1] eval
    @ Core ./boot.jl:385 [inlined]
  [2] eval
    @ DynamicQuantities.UnitsParse ~/.julia/packages/DynamicQuantities/5QflN/src/uparse.jl:1 [inlined]
  [3] uparse(s::String)
    @ DynamicQuantities.UnitsParse ~/.julia/packages/DynamicQuantities/5QflN/src/uparse.jl:37
  [4] var"@u_str"(__source__::LineNumberNode, __module__::Module, s::Any)
    @ DynamicQuantities.UnitsParse ~/.julia/packages/DynamicQuantities/5QflN/src/uparse.jl:57
...
```

MilesCranmer:
Fixes #105 which reported precompilation issues.

- Manual parser for `u_str` to avoid calling `eval` (which breaks precompilation)
  - Also implements for `us_str`
- Creates the `NoDims` type so that `dimension(1.0)` doesn't throw a method error.
- Allows you to use `uconvert` in more contexts, like `SymbolicDimensions` -> `SymbolicDimensions`

### DIFF
--- a/src/uparse.jl
+++ b/src/uparse.jl
@@ -66,14 +66,11 @@ function map_to_scope(ex::Expr)
     if ex.head == :call
         ex.args[2:end] = map(map_to_scope, ex.args[2:end])
         return ex
-    elseif ex.head == :tuple
-        ex.args[:] = map(map_to_scope, ex.args)
-        return ex
     elseif ex.head == :. && ex.args[1] == :Constants
         @assert ex.args[2] isa QuoteNode
         return lookup_constant(ex.args[2].value)
     else
-        throw(ArgumentError("Unexpected expression: $ex. Only `:call`, `:tuple`, and `:.` (for `Constants`) are expected."))
+        throw(ArgumentError("Unexpected expression: $ex. Only `:call` and `:.` (for `Constants`) are expected."))
     end
 end
 function map_to_scope(sym::Symbol)
--- a/src/uparse.jl
+++ b/src/uparse.jl
@@ -54,7 +54,7 @@ the quantity corresponding to the speed of light multiplied by Hertz,
 squared.
 """
 macro u_str(s)
-    return esc(uparse(s))
+    return esc(Meta.parse(s))
 end
 
 end
--- a/src/uparse.jl
+++ b/src/uparse.jl
@@ -4,7 +4,8 @@ import ..constructorof
 import ..DEFAULT_QUANTITY_TYPE
 import ..DEFAULT_DIM_TYPE
 import ..DEFAULT_VALUE_TYPE
-import ..Units: UNIT_SYMBOLS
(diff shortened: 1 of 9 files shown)


MilesCranmer:
@YingboMa what do you think? This completely avoids any `eval` when evaluating `u_str`.

I also removed the limitation of type requirement. So you can do things like `u"(m, s)"`.

@gaurav-arya are you okay with this? I think we might have to remove that type requirement to fix precompilation working in downstream packages, since evaluating types requires `eval`.

src/uparse.jl
```diff
@@ -54,7 +55,43 @@ the quantity corresponding to the speed of light multiplied by Hertz,
...
+function map_to_scope(ex)
+    return ex
+end
+function lookup_unit(ex::Symbol)
+    i = findfirst(==(ex), UNIT_SYMBOLS)::Int
+    return UNIT_VALUES[i]
+end
+function lookup_constant(ex::Symbol)
+    i = findfirst(==(ex), CONSTANT_SYMBOLS)::Int
+    return CONSTANT_VALUES[i]
```
MilesCranmer, line 94: This is the fix for #105 @YingboMa

MilesCranmer:
Since this is quite a significant change in the parsing I'm going to sit on this PR for a bit in case I think of anything wrong with it.

@gaurav-arya if you could take a look also that would be amazing

MilesCranmer:
For backwards compatibility we could have `as_quantity` injected into the output – then one could still use `u"1"` to create quantities?

It's not 1:1 with Unitful then but at least it preserves the type stability. Like `q = u"1"; q *= 1.5u"m"` for example.

MilesCranmer:
Okay I convinced myself the other way. I think `u_str` should always return the same type no matter what – it's just the safest thing to do. Even though macros are expanded during compilation, I can easily see the user encountering type instabilities because of this.

This is unlike Unitful, but is like the current implementation of DynamicQuantities. So with this change it's no longer breaking.

MilesCranmer:
Woo! 100% test coverage!

Alright we're good now. I'm merging as it's no longer a breaking change. @gaurav-arya if you did have a comment on the behavior of `u_str` we can always implement the breaking change on a new version later.

@YingboMa please confirm that precompilation works on v0.11.2 and feel free to add to the stdlib now.

PR #107, opened 2024-01-09
Allow External Unit Registration

ven-k:
- By interpolating `Units.UNIT_SYMBOLS` while registering units and updating vectors-and-maps-of-units, new units can be registered even from outside the `Units` module
- Internally, units are registered lazily with `_lazy_register_unit` aka
`UNIT_MAPPING` is updated after all units are defined.

Fixes:
```julia
julia> DynamicQuantities.Units.@register_unit Wb "m^2*kg*s^-2*A^-1"
ERROR: UndefVarError: `_UNIT_SYMBOLS` not defined
Stacktrace:
 [1] top-level scope
   @ C:\Users\user\.julia\packages\DynamicQuantities\5QflN\src\units.jl:25
```

### DIFF
--- a/src/register_units.jl
+++ b/src/register_units.jl
@@ -20,6 +20,30 @@ end
 
 Register a new unit under the given symbol to have
 a particular value.
+
+# Example
+
+```julia
+julia> @register_unit MyVolt 1.5u"V"
+```
+
+This will register a new unit `MyVolt` with a value of `1.5u"V"`.
+You can then use this unit in your calculations:
+
+```julia
+julia> x = 20us"MyVolt^2"
+20.0 MyVolt²
+
+julia> y = 2.5us"A"
+2.5 A
+
+julia> x * y^2 |> uconvert(us"W^2")
+281.25 W²
+
+julia> x * y^2 |> uconvert(us"W^2") |> sqrt |> uexpand
+16.77050983124842 m² kg s⁻³
+```
+
 """
 macro register_unit(symbol, value)
     return esc(_register_unit(symbol, value))
--- a/src/register_units.jl
+++ b/src/register_units.jl
@@ -52,19 +52,21 @@ end
 function _register_unit(name::Symbol, value)
     name_symbol = Meta.quot(name)
     index = get(ALL_MAPPING, name, INDEX_TYPE(0))
-    if iszero(index)
-        reg_expr = _lazy_register_unit(name, value)
-        push!(reg_expr.args, quote
-            $update_all_values($name_symbol, $value)
-            nothing
-        end)
-        return reg_expr
-    else
+    if !iszero(index)
         unit = ALL_VALUES[index]
         # When a utility function to expand `value` to its final form becomes
         # available, enable the following check. This will avoid throwing an error
         # if user is trying to register an existing unit with matching values.
         # unit.value != value && throw("Unit $name is already defined as $unit")
-        throw("Unit `$name` is already defined as `$unit`")
+        error("Unit `$name` is already defined as `$unit`")
     end
+    reg_expr = _lazy_register_unit(name, value)
+    push!(
+        reg_expr.args,
+        quote
+            $update_all_values($name_symbol, $value)
+            nothing
+        end
+    )
+    return reg_expr
 end
--- a/src/register_units.jl
+++ b/src/register_units.jl
@@ -3,14 +3,17 @@ import .SymbolicUnits:
     SymbolicDimensionsSingleton, SYMBOLIC_UNIT_VALUES, update_symbolic_unit_values!
 
 # Update the unit collections
-function update_unit_mapping(name, value, unit_mapping::Dict{Symbol,Int} = UNIT_MAPPING)
-    unit_mapping[name] = length(unit_mapping) + 1
-end
+const UNIT_UPDATE_LOCK = Threads.SpinLock()
 
 function update_all_values(name_symbol, unit)
-    push!(ALL_SYMBOLS, name_symbol)
-    push!(ALL_VALUES, unit)
-    ALL_MAPPING[name_symbol] = INDEX_TYPE(length(ALL_MAPPING) + 1)
+    lock(UNIT_UPDATE_LOCK) do
+        push!(ALL_SYMBOLS, name_symbol)
+        push!(ALL_VALUES, unit)
+        i = lastindex(ALL_VALUES)
+        ALL_MAPPING[name_symbol] = i
+        UNIT_MAPPING[name_symbol] = i
+        update_symbolic_unit_values!(name_symbol)
+    end
 end
 
 # Register
@@ -20,16 +23,20 @@ end
 
 function _register_unit(name::Symbol, value)
     name_symbol = Meta.quot(name)
-    reg_expr = _lazy_register_unit(name, value)
-    push!(reg_expr.args,
-        quote
-            $update_unit_mapping($name_symbol, $value)
+    index = get(ALL_MAPPING, name, INDEX_TYPE(0))
+    if iszero(index)
+        reg_expr = _lazy_register_unit(name, value)
+        push!(reg_expr.args, quote
             $update_all_values($name_symbol, $value)
-            $update_symbolic_unit_values!($name_symbol)
-            # suppress the print of `SYMBOLIC_UNIT_VALUES`
             nothing
-        end
-    )
-    return reg_expr
+        end)
+        return reg_expr
+    else
+        unit = ALL_VALUES[index]
+        # When a utility function to expand `value` to its final form becomes
+        # available, enable the following check. This will avoid throwing an error
+        # if user is trying to register an existing unit with matching values.
+        # unit.value != value && throw("Unit $name is already defined as $unit")
+        throw("Unit `$name` is already defined as `$unit`")
+    end
 end
-
--- a/src/register_units.jl
+++ b/src/register_units.jl
@@ -1,6 +1,5 @@
 import .Units: UNIT_MAPPING, UNIT_SYMBOLS, UNIT_VALUES, _lazy_register_unit
-import .SymbolicUnits:
-    SymbolicDimensionsSingleton, SYMBOLIC_UNIT_VALUES, update_symbolic_unit_values!
+import .SymbolicUnits: update_external_symbolic_unit_value
 
 # Update the unit collections
 const UNIT_UPDATE_LOCK = Threads.SpinLock()
@@ -12,7 +11,7 @@ function update_all_values(name_symbol, unit)
         i = lastindex(ALL_VALUES)
         ALL_MAPPING[name_symbol] = i
         UNIT_MAPPING[name_symbol] = i
-        update_symbolic_unit_values!(name_symbol)
+        update_external_symbolic_unit_value(name_symbol)
     end
 end
 
--- a/src/register_units.jl
+++ b/src/register_units.jl
@@ -15,9 +15,14 @@ function update_all_values(name_symbol, unit)
     end
 end
 
-# Register
-macro register_unit(name, value)
-    return esc(_register_unit(name, value))
+"""
+    @register_unit symbol value
+
+Register a new unit under the given symbol to have
+a particular value.
+"""
+macro register_unit(symbol, value)
+    return esc(_register_unit(symbol, value))
 end
 
 function _register_unit(name::Symbol, value)
--- a/src/register_units.jl
+++ b/src/register_units.jl
@@ -1,4 +1,3 @@
-
 import .Units: UNIT_MAPPING, UNIT_SYMBOLS, UNIT_VALUES, _lazy_register_unit
 import .SymbolicUnits:
     SymbolicDimensionsSingleton, SYMBOLIC_UNIT_VALUES, update_symbolic_unit_values!
@@ -13,7 +12,6 @@ function update_all_values(name_symbol, unit)
     push!(ALL_VALUES, unit)
     ALL_MAPPING[name_symbol] = INDEX_TYPE(length(ALL_MAPPING) + 1)
 end
-
 
 # Register
 macro register_unit(name, value)
--- a/src/register_units.jl
+++ b/src/register_units.jl
@@ -1,0 +1,37 @@
+
+import .Units: UNIT_MAPPING, UNIT_SYMBOLS, UNIT_VALUES, _lazy_register_unit
+import .SymbolicUnits:
+    SymbolicDimensionsSingleton, SYMBOLIC_UNIT_VALUES, update_symbolic_unit_values!
+
+# Update the unit collections
+function update_unit_mapping(name, value, unit_mapping::Dict{Symbol,Int} = UNIT_MAPPING)
+    unit_mapping[name] = length(unit_mapping) + 1
+end
+
+function update_all_values(name_symbol, unit)
+    push!(ALL_SYMBOLS, name_symbol)
+    push!(ALL_VALUES, unit)
+    ALL_MAPPING[name_symbol] = INDEX_TYPE(length(ALL_MAPPING) + 1)
+end
+
+
+# Register
+macro register_unit(name, value)
+    return esc(_register_unit(name, value))
+end
+
+function _register_unit(name::Symbol, value)
+    name_symbol = Meta.quot(name)
+    reg_expr = _lazy_register_unit(name, value)
+    push!(reg_expr.args,
+        quote
+            $update_unit_mapping($name_symbol, $value)
+            $update_all_values($name_symbol, $value)
+            $update_symbolic_unit_values!($name_symbol)
+            # suppress the print of `SYMBOLIC_UNIT_VALUES`
+            nothing
+        end
+    )
+    return reg_expr
+end
+
--- a/src/symbolic_dimensions.jl
+++ b/src/symbolic_dimensions.jl
@@ -1,10 +1,11 @@
 import ..WriteOnceReadMany
-import ..INDEX_TYPE
 import .Units: UNIT_SYMBOLS, UNIT_MAPPING, UNIT_VALUES
 import .Constants: CONSTANT_SYMBOLS, CONSTANT_MAPPING, CONSTANT_VALUES
 
+
 disambiguate_constant_symbol(s) = s in UNIT_SYMBOLS ? Symbol(s, :_constant) : s
 
+const INDEX_TYPE = UInt16
 # Prefer units over constants:
 # For example, this means we can't have a symbolic Planck's constant,
 # as it is just "hours" (h), which is more common.
--- a/src/symbolic_dimensions.jl
+++ b/src/symbolic_dimensions.jl
@@ -1,21 +1,20 @@
 import .Units: UNIT_SYMBOLS, UNIT_MAPPING, UNIT_VALUES
 import .Constants: CONSTANT_SYMBOLS, CONSTANT_MAPPING, CONSTANT_VALUES
 
-
 const SYMBOL_CONFLICTS = intersect(UNIT_SYMBOLS, CONSTANT_SYMBOLS)
 
 disambiguate_symbol(s) = s in SYMBOL_CONFLICTS ? Symbol(s, :_constant) : s
 
-const INDEX_TYPE = UInt8
 # Prefer units over constants:
 # For example, this means we can't have a symbolic Planck's constant,
 # as it is just "hours" (h), which is more common.
-const ALL_SYMBOLS = (
-    UNIT_SYMBOLS...,
-    disambiguate_symbol.(CONSTANT_SYMBOLS)...
-)
-const ALL_VALUES = (UNIT_VALUES..., CONSTANT_VALUES...)
-const ALL_MAPPING = NamedTuple{ALL_SYMBOLS}(INDEX_TYPE(1):INDEX_TYPE(length(ALL_SYMBOLS)))
+const INDEX_TYPE = UInt16
+# Prefer units over constants:
+# For example, this means we can't have a symbolic Planck's constant,
+# as it is just "hours" (h), which is more common.
+const ALL_SYMBOLS = [UNIT_SYMBOLS..., disambiguate_symbol.(CONSTANT_SYMBOLS)...]
+const ALL_VALUES = [UNIT_VALUES..., CONSTANT_VALUES...]
+const ALL_MAPPING = Dict(ALL_SYMBOLS .=> (INDEX_TYPE(1):INDEX_TYPE(length(ALL_SYMBOLS))))
 
 """
     AbstractSymbolicDimensions{R} <: AbstractDimensions{R}
@@ -91,7 +90,7 @@ end
 function SymbolicDimensionsSingleton{R}(s::Symbol) where {R}
     i = get(ALL_MAPPING, s, INDEX_TYPE(0))
     iszero(i) && error("$s is not available as a symbol in `SymbolicDimensionsSingleton`. Symbols available: $(ALL_SYMBOLS).")
-    return SymbolicDimensionsSingleton{R}(i)
+    SymbolicDimensionsSingleton{R}(i)
 end
 
 # Traits:
@@ -169,7 +168,7 @@ uexpand(q::QuantityArray) = uexpand.(q)
     uconvert(qout::UnionAbstractQuantity{<:Any, <:AbstractSymbolicDimensions}, q::UnionAbstractQuantity{<:Any, <:Dimensions})
 
 Convert a quantity `q` with base SI units to the symbolic units of `qout`, for `q` and `qout` with compatible units.
-Mathematically, the result has value `q / uexpand(qout)` and units `dimension(qout)`. 
+Mathematically, the result has value `q / uexpand(qout)` and units `dimension(qout)`.
 """
 function uconvert(qout::UnionAbstractQuantity{<:Any, <:SymbolicDimensions}, q::UnionAbstractQuantity{<:Any, <:Dimensions})
     @assert isone(ustrip(qout)) "You passed a quantity with a non-unit value to uconvert."
@@ -224,7 +223,7 @@ end
 """
     uconvert(qout::UnionAbstractQuantity{<:Any, <:AbstractSymbolicDimensions})
 
-Create a function that converts an input quantity `q` with base SI units to the symbolic units of `qout`, i.e 
+Create a function that converts an input quantity `q` with base SI units to the symbolic units of `qout`, i.e
 a function equivalent to `q -> uconvert(qout, q)`.
 """
 uconvert(qout::UnionAbstractQuantity{<:Any,<:AbstractSymbolicDimensions}) = Base.Fix1(uconvert, qout)
@@ -368,6 +367,7 @@ to enable pretty-printing of units.
 """
 module SymbolicUnits
 
+
     import ..UNIT_SYMBOLS
     import ..CONSTANT_SYMBOLS
     import ..SymbolicDimensionsSingleton
@@ -383,9 +383,9 @@ module SymbolicUnits
         import ...SymbolicDimensionsSingleton
         import ...constructorof
         import ...disambiguate_symbol
-        import ....DEFAULT_SYMBOLIC_QUANTITY_TYPE
-        import ....DEFAULT_VALUE_TYPE
-        import ....DEFAULT_DIM_BASE_TYPE
+        import ...DEFAULT_SYMBOLIC_QUANTITY_TYPE
+        import ...DEFAULT_VALUE_TYPE
+        import ...DEFAULT_DIM_BASE_TYPE
 
         const _SYMBOLIC_CONSTANT_VALUES = DEFAULT_SYMBOLIC_QUANTITY_TYPE[]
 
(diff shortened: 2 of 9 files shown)


MilesCranmer:
Great idea! Can you rebase to #106? As there are changes which might affect the behavior.

test/unittests.jl
```diff
@@ -1697,3 +1698,8 @@ end
     y = Quantity(2.0im, mass=1)
     @test_throws DimensionError x^y
 end
+
+@testset "Register Units" begin
+    @register_unit MyV u"V"
+    @test MyV === u"V"
```
MilesCranmer, line 1704: Should test for `u"MyV"`

Also include a test for symbolic units, i.e., `us"MyV"`

MilesCranmer: To get this working you might need to let `UNIT_SYMBOLS` be a regular vector rather than a tuple FYI.

src/units.jl
```diff
@@ -21,9 +21,9 @@ end
 function _register_unit(name::Symbol, value)
     s = string(name)
     return quote
-        const $name = $value
```
MilesCranmer, line 24: I think the symbols should still be constants though?

MilesCranmer: Otherwise `using DynamicQuantities.Units: m`  – the `m` would be type unstable as its not a constant.

ven-k: I reverted to original.
Instead, I'm registering units outside testset. See [here](https://github.com/SymbolicML/DynamicQuantities.jl/pull/107/files#diff-97287d1f47874f9eb568bc53640b753ce7f6cd5fd70a18180a4beb9dd8a7939bR1852-R1854).

MilesCranmer:
Thanks! Quick initial comment is that this will need to update `symbolic_dimensions.jl`'s `ALL_VALUES` to be vectors as well, so that new units are also included in symbolic units (like `us"km/h"`).

test/unittests.jl
```diff
@@ -1,16 +1,18 @@
...
 using DynamicQuantities: promote_quantity_on_quantity, promote_quantity_on_value
+using DynamicQuantities.Units: @register_unit, UNIT_VALUES, UNIT_MAPPING
 using DynamicQuantities: map_dimensions
 using Ratios: SimpleRatio
 using SaferIntegers: SafeInt16
 using StaticArrays: SArray, MArray
 using LinearAlgebra: norm
 using Test
-
+#=
```
ven-k, line 15: `#=...=#` Should be removed here and at line 1847

src/units.jl
```diff
@@ -6,24 +6,35 @@ import ..DEFAULT_QUANTITY_TYPE
...
+function _register_unit(name::Symbol, value; lazy = false)
+    name_symbol = Meta.quot(name)
     return quote
+        haskey($UNIT_MAPPING, $name_symbol) && throw("Unit $name_symbol already exists.")
         const $name = $value
-        push!(_UNIT_SYMBOLS, Symbol($s))
-        push!(_UNIT_VALUES, $name)
+        push!($UNIT_SYMBOLS, $name_symbol)
+        push!($UNIT_VALUES, $name)
+        !$lazy && $update_unit_mapping($name_symbol, $value)
```
MilesCranmer, line 37: Since all the calls are `_lazy` here, when does this function actually get called? (I assume it does, but I don't see where)

ven-k: Whenever externally a unit is registered, this executes and updates the UNIT_MAPPING; I'll extend it to update the `ALL_UNITS`, `ALL_MAPPING` and `ALL_SYMBOLS` too.

MilesCranmer:
One other quick thought is that we will probably need to change `INDEX_TYPE` to `UInt16` in `symbolic_dimensions.jl`. I could see that if we are to allow users to add units and constants, the 8-bit indexing will probably be closer to overflow. However this also probably means things will be a bit slower.

MilesCranmer:
Let me know when you're ready and I can give more comments or help out

ChrisRackauckas:
@ven-k ?

src/units.jl
ven-k: Resolved with [`a4335c5` (#107)](https://github.com/SymbolicML/DynamicQuantities.jl/pull/107/commits/a4335c5e5792d947932766b62eb4f4e836437d29)

ven-k:
Now works:
```julia
julia> using ModelingToolkit, DynamicQuantities

julia> @register_unit Wb u"kg*m^2*s^−2*A^−1"

julia> @variables k [unit = u"Wb"]
1-element Vector{Num}:
 k
```

src/symbolic_dimensions.jl
```diff
@@ -1,21 +1,20 @@
...
-)
-const ALL_VALUES = (UNIT_VALUES..., CONSTANT_VALUES...)
-const ALL_MAPPING = NamedTuple{ALL_SYMBOLS}(INDEX_TYPE(1):INDEX_TYPE(length(ALL_SYMBOLS)))
+const INDEX_TYPE = UInt16
+# Prefer units over constants:
+# For example, this means we can't have a symbolic Planck's constant,
+# as it is just "hours" (h), which is more common.
+const ALL_SYMBOLS = [UNIT_SYMBOLS..., disambiguate_symbol.(CONSTANT_SYMBOLS)...]
+const ALL_VALUES = [UNIT_VALUES..., CONSTANT_VALUES...]
+const ALL_MAPPING = Dict(ALL_SYMBOLS .=> (INDEX_TYPE(1):INDEX_TYPE(length(ALL_SYMBOLS))))
```
devmotion, line 17: The broadcasting is not needed:
```suggestion
const ALL_MAPPING = Dict(s => INDEX_TYPE(i) for (i, s) in enumerate(ALL_SYMBOLS))
```

src/units.jl
```diff
@@ -6,27 +6,30 @@ import ..DEFAULT_QUANTITY_TYPE
 
 @assert DEFAULT_VALUE_TYPE == Float64 "`units.jl` must be updated to support a different default value type."
 
-const _UNIT_SYMBOLS = Symbol[]
-const _UNIT_VALUES = DEFAULT_QUANTITY_TYPE[]
+const UNIT_SYMBOLS = Symbol[]
+const UNIT_VALUES = DEFAULT_QUANTITY_TYPE[]
+const UNIT_MAPPING = Dict{Symbol,Int}()
```
devmotion, line 11: It can be defined below?
```suggestion
```

src/units.jl
```diff
@@ -203,9 +206,7 @@ end
 # Do not wish to define physical constants, as the number of symbols might lead to ambiguity.
 # The user should define these instead.
 
-"""A tuple of all possible unit symbols."""
-const UNIT_SYMBOLS = Tuple(_UNIT_SYMBOLS)
-const UNIT_VALUES = Tuple(_UNIT_VALUES)
-const UNIT_MAPPING = NamedTuple([s => i for (i, s) in enumerate(UNIT_SYMBOLS)])
+# Update `UNIT_MAPPING` with all internally defined unit symbols.
+merge!(UNIT_MAPPING, Dict(UNIT_SYMBOLS .=> 1:lastindex(UNIT_SYMBOLS)))
```
devmotion, line 210:
```suggestion
const UNIT_MAPPING = Dict(s => i for (i, s) in enumerate(UNIT_SYMBOLS))
```

src/symbolic_dimensions.jl
ven-k: When I benchmarked, broadcasting was much faster than enumerating. So I switched.

This is what I ran:
```julia
julia> k = [Symbol(randstring(10)) for i in 1:100]
julia>INDEX_TYPE = UInt16
julia> @benchmark ALL_MAPPING = Dict($k .=> (INDEX_TYPE(1):INDEX_TYPE(length($k))))
BenchmarkTools.Trial: 10000 samples with 9 evaluations.
 Range (min … max):  1.878 μs … 411.511 μs  ┊ GC (min … max): 0.00% … 98.51%
 Time  (median):     2.100 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   3.013 μs ±   9.624 μs  ┊ GC (mean ± σ):  9.70% ±  3.10%

  ▇█▄▂▃▃▄▃▂▂▃▁                              ▂▂                ▂
  ██████████████▇▇▅▆▅▆▄▄▄▄▄▄▃▃▄▃▁▁▃▁▁▁▃▁▄▆▇████▇▅▇▆▆▆▄▅▅▅▄▅▃▅ █
  1.88 μs      Histogram: log(frequency) by time      11.8 μs <

 Memory estimate: 5.08 KiB, allocs estimate: 10.

julia> @benchmark  ALL_MAPPING = Dict(s => INDEX_TYPE(i) for (i, s) in enumerate($k))
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  30.000 μs …  3.332 ms  ┊ GC (min … max): 0.00% … 97.70%
 Time  (median):     31.000 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   33.411 μs ± 33.523 μs  ┊ GC (mean ± σ):  0.97% ±  0.98%

  ▃█▇▇▅▄▃▃▂▂▂                                     ▁▂▁▁▁       ▂
  ██████████████▇▇▆▆▆▅▆▆▅▅▅▄▆▆▆▆▅▅▆▆▆▇▆▇█▇▇█▇▇▇██▇█████▇██▆▆▆ █
  30 μs        Histogram: log(frequency) by time      51.5 μs <

 Memory estimate: 7.95 KiB, allocs estimate: 116.
```

devmotion: The problem is that you benchmark directly with the global `INDEX_TYPE`. This leads to incorrect results (an indication are the allocations in the second case). For instance, I get
```julia
julia> using BenchmarkTools

julia> k = [Symbol(randstring(10)) for i in 1:100];

julia> INDEX_TYPE = UInt16
UInt16

julia> function f(k, ::Type{T}) where {T}
           Dict(k .=> (T(1):T(length(k))))
       end

julia> function g(k, ::Type{T}) where {T}
           Dict(s => T(i) for (i, s) in enumerate(k))
       end
g (generic function with 1 method)

julia> function h(k, ::Type{T}) where {T}
           Dict(map((i, s) -> s => T(i), 1:length(k), k))
       end
h (generic function with 1 method)

julia> f(k, INDEX_TYPE) == g(k, INDEX_TYPE) == h(k, INDEX_TYPE)
true

julia> @benchmark f($k, $INDEX_TYPE)
BenchmarkTools.Trial: 10000 samples with 14 evaluations.
 Range (min … max):  961.286 ns … 99.491 μs  ┊ GC (min … max): 0.00% … 97.89%
 Time  (median):       1.116 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):     1.243 μs ±  3.204 μs  ┊ GC (mean ± σ):  9.84% ±  3.78%

      ▆▄▇▂     ▂▂▃█▄█▂▄
  ▁▂▆██████▇▅▇▆████████▇▅▆▄▄▃▂▂▂▂▂▂▂▂▂▂▁▁▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▃
  961 ns          Histogram: frequency by time         1.57 μs <

 Memory estimate: 5.03 KiB, allocs estimate: 8.

julia> @benchmark g($k, $INDEX_TYPE)
BenchmarkTools.Trial: 10000 samples with 69 evaluations.
 Range (min … max):  837.565 ns … 112.315 μs  ┊ GC (min … max):  0.00% … 98.62%
 Time  (median):       1.010 μs               ┊ GC (median):     0.00%
 Time  (mean ± σ):     1.194 μs ±   2.656 μs  ┊ GC (mean ± σ):  14.97% ±  6.82%

    ▂▁             ▄▃██▆▆▄▆▃▂▁
  ▂███▅▂▂▂▂▃▃▂▃▃▃▆████████████▇█▇▇▆▅▅▄▅▄▃▃▃▃▂▂▃▃▂▂▂▂▂▂▂▂▂▁▁▁▁▁▁ ▄
  838 ns           Histogram: frequency by time          1.3 μs <

 Memory estimate: 3.27 KiB, allocs estimate: 7.

julia> @benchmark h($k, $INDEX_TYPE)
BenchmarkTools.Trial: 10000 samples with 10 evaluations.
 Range (min … max):  991.700 ns … 169.917 μs  ┊ GC (min … max):  0.00% … 98.46%
 Time  (median):       1.171 μs               ┊ GC (median):     0.00%
 Time  (mean ± σ):     1.340 μs ±   4.559 μs  ┊ GC (mean ± σ):  10.60% ±  3.11%

     ▂    ▃▃▆█▅█▃▄▁
  ▂▄▇███▇▇█████████▆▆▄▃▃▃▃▃▃▂▃▃▂▃▂▃▃▃▃▃▃▃▂▂▂▂▂▂▂▂▂▁▂▂▂▂▂▂▂▂▂▂▂▂ ▄
  992 ns           Histogram: frequency by time          1.9 μs <

 Memory estimate: 5.03 KiB, allocs estimate: 8.
```

MilesCranmer:
One other comment: trying to register the same unit twice should have no effect. That way you could register units within a method scope without fear of creating duplicates. Also it would probably make things safer for Revise.jl

src/units.jl
MilesCranmer: Doesn’t even need to be a set, you can just check existence within the `push!` and throw an error if the same symbol was pushed twice

src/register_units.jl
```diff
@@ -0,0 +1,35 @@
...
+# Update the unit collections
+function update_unit_mapping(name, value, unit_mapping::Dict{Symbol,Int} = UNIT_MAPPING)
+    unit_mapping[name] = length(unit_mapping) + 1
+end
+
+function update_all_values(name_symbol, unit)
+    push!(ALL_SYMBOLS, name_symbol)
+    push!(ALL_VALUES, unit)
+    ALL_MAPPING[name_symbol] = INDEX_TYPE(length(ALL_MAPPING) + 1)
+end
```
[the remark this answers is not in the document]
ven-k: Resolved with https://github.com/SymbolicML/DynamicQuantities.jl/pull/107/commits/bfc699ffbe5f0e0d882f168de2b8bd92fec9941f

src/register_units.jl
```diff
@@ -0,0 +1,35 @@
+import .Units: UNIT_MAPPING, UNIT_SYMBOLS, UNIT_VALUES, _lazy_register_unit
+import .SymbolicUnits:
+    SymbolicDimensionsSingleton, SYMBOLIC_UNIT_VALUES, update_symbolic_unit_values!
+
+# Update the unit collections
+function update_unit_mapping(name, value, unit_mapping::Dict{Symbol,Int} = UNIT_MAPPING)
+    unit_mapping[name] = length(unit_mapping) + 1
+end
+
```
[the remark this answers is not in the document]
ven-k: Resolved https://github.com/SymbolicML/DynamicQuantities.jl/pull/107/commits/bfc699ffbe5f0e0d882f168de2b8bd92fec9941f

src/register_units.jl
```diff
@@ -0,0 +1,35 @@
...
+function update_all_values(name_symbol, unit)
```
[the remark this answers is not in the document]
ven-k: Resolved https://github.com/SymbolicML/DynamicQuantities.jl/pull/107/commits/bfc699ffbe5f0e0d882f168de2b8bd92fec9941f

src/register_units.jl
```diff
@@ -0,0 +1,35 @@
...
+macro register_unit(name, value)
+    return esc(_register_unit(name, value))
+end
+
+function _register_unit(name::Symbol, value)
+    name_symbol = Meta.quot(name)
+    reg_expr = _lazy_register_unit(name, value)
+    push!(reg_expr.args,
+        quote
+            $update_unit_mapping($name_symbol, $value)
```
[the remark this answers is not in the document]
ven-k: Resolved https://github.com/SymbolicML/DynamicQuantities.jl/pull/107/commits/bfc699ffbe5f0e0d882f168de2b8bd92fec9941f

src/register_units.jl
```diff
@@ -0,0 +1,35 @@
+import .Units: UNIT_MAPPING, UNIT_SYMBOLS, UNIT_VALUES, _lazy_register_unit
+import .SymbolicUnits:
+    SymbolicDimensionsSingleton, SYMBOLIC_UNIT_VALUES, update_symbolic_unit_values!
+
+# Update the unit collections
+function update_unit_mapping(name, value, unit_mapping::Dict{Symbol,Int} = UNIT_MAPPING)
+    unit_mapping[name] = length(unit_mapping) + 1
+end
```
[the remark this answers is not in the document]
ven-k: Resolved https://github.com/SymbolicML/DynamicQuantities.jl/pull/107/commits/bfc699ffbe5f0e0d882f168de2b8bd92fec9941f

src/units.jl
ven-k: Resolved https://github.com/SymbolicML/DynamicQuantities.jl/pull/107/commits/bfc699ffbe5f0e0d882f168de2b8bd92fec9941f

src/units.jl
ven-k: I've added `DynamicQuantitiesCollection` with restricted set of operations.

src/symbolic_dimensions.jl
ven-k: Resovled https://github.com/SymbolicML/DynamicQuantities.jl/pull/107/commits/bfc699ffbe5f0e0d882f168de2b8bd92fec9941f

ven-k:
> trying to register the same unit twice should have no effect

Just comparing `UNIT_VALUE[index].value == value`, [here](https://github.com/SymbolicML/DynamicQuantities.jl/pull/107/files#diff-1c43fbb05972d51cae7f3220da8abb11e83c694296581f940b5f6e49153facd7R36), (aka user input value) will not work as latter is an Expr. For running this check, is there an internal helper which can expand it to its intended form?

MilesCranmer:
Thanks for implementing all of those changes. Sorry to go back on what @devmotion said but do you think we could avoid the `OrderedSet` use? Since `ALL_VALUES` is a vector I think `ALL_SYMBOLS` should be as well – otherwise I get worried there could be an indexing bug introduced somewhere. If not now, maybe later. But I think this is actually already buggy because `ALL_SYMBOLS` could be left unchanged (since its a set) but `ALL_VALUES` could have another element added to it – since the function doesn't return early.

Basically I think those constants should just be [write once read many](https://en.wikipedia.org/wiki/Write_once_read_many) arrays. When you `push!` to the `ALL_SYMBOLS` within the register units part, you could simply have a check to see if the symbol is already in it – if so, return early.

I think the struct could just be something like

```
struct WriteOnceReadManyVector{T}
    _raw_data::Vector{T}
end
```

and just define `Base.push!` and only the methods needed but nothing else (so the array can't be removed from). You do do the same for `WriteOnceReadManyDict` for `ALL_MAPPING` also I think.

MilesCranmer:
Now that I think about it, trying to register a unit should trigger an error if the symbol is already used AND the value assigned to that symbol is different than the one being registered.

But if the symbol is used already and the value is the same, it could simply be ignored and the unit registration should return early (before pushing anything).

MilesCranmer:
One option is to simply have a single array storing a tuple of (1) the symbol, (2) the value in `Dimensions`, and (3) the value in `SymbolicDimensions`. Then we don’t have to worry about issues from expanding one array and not the other.

src/utils.jl
```diff
@@ -428,3 +428,33 @@ Get the amount dimension of a quantity (e.g., mol^(uamount)).
...
+for f in (:enumerate, :length, :lastindex)
+    @eval begin
+        Base.$f(u::DynamicQuantitiesCollection) = $f(u.collection)
+    end
+end
+
+Base.getindex(u::DynamicQuantitiesCollection, i::Union{Int,  Symbol}) = getindex(u.collection, i)
+Base.getindex(u::DynamicQuantitiesCollection, i::UInt16) = getindex(u.collection, convert(Int, i))
+
+Base.setindex!(u::DynamicQuantities.DynamicQuantitiesCollection{Dict{Symbol, T}}, i::Int64, s::Symbol) where T <: Integer = setindex!(u.collection, i, s)
```
MilesCranmer, line 445: Should throw an error if the key already exists

src/utils.jl
```diff
@@ -428,3 +428,33 @@ Get the amount dimension of a quantity (e.g., mol^(uamount)).
...
+Base.get(u::DynamicQuantitiesCollection{Dict{Symbol, INDEX_TYPE}}, a, b) = get(u.collection, a, b)
+
+Base.iterate(u::DynamicQuantitiesCollection) = iterate(u.collection)
+Base.iterate(u::DynamicQuantitiesCollection, i::Int) = iterate(u.collection, i)
+
+Base.intersect(u::DynamicQuantitiesCollection, v::AbstractSet) = intersect(u.collection, v)
+Base.intersect(v::AbstractSet, u::DynamicQuantitiesCollection) = intersect(v, u.collection)
```
MilesCranmer, line 452: I don’t think these are used anywhere FYI

src/utils.jl
```diff
@@ -428,3 +428,33 @@ Get the amount dimension of a quantity (e.g., mol^(uamount)).
 """
 uamount(q::UnionAbstractQuantity) = uamount(dimension(q))
 uamount(d::AbstractDimensions) = d.amount
+
+"""
+Utility functions for `DynamicQuantitiesCollection`
+"""
+
+for f in (:enumerate, :length, :lastindex)
```
MilesCranmer, line 436: I think defining maybe `eachindex` (?) would be better and would also give you `lastindex` for free

src/utils.jl
```diff
@@ -428,3 +428,33 @@ Get the amount dimension of a quantity (e.g., mol^(uamount)).
...
+    @eval begin
+        Base.$f(u::DynamicQuantitiesCollection) = $f(u.collection)
+    end
+end
+
+Base.getindex(u::DynamicQuantitiesCollection, i::Union{Int,  Symbol}) = getindex(u.collection, i)
+Base.getindex(u::DynamicQuantitiesCollection, i::UInt16) = getindex(u.collection, convert(Int, i))
```
MilesCranmer, line 443: Maybe just use `Union{Integer,Symbol}` for a single method? It should automatically get converted by the underlying collection, no?

src/utils.jl
MilesCranmer: Wait sorry you are right. But these are needed: https://docs.julialang.org/en/v1/manual/interfaces/#Indexing

src/utils.jl
ven-k: Used here: https://github.com/SymbolicML/DynamicQuantities.jl/blob/main/src/symbolic_dimensions.jl#L5.

Maybe even Constant-collection can use the same struct; but I don't want to overload this PR.

ven-k:
> One option is to simply have a single array storing a tuple of (1) the symbol, (2) the value in Dimensions, and (3) the value in SymbolicDimensionsSingleton

The `SymbolicDimensionsSingleton` is defined in `SymbolicUnits`, which is executed after Units is done updating `UNIT_*`. Updating one such array would be neat; however would need some major changes in code placement.

src/utils.jl
MilesCranmer: Oh I see, sorry. I’m not sure why the testing coverage didn’t catch that.

MilesCranmer:
Good point. Okay let’s table that for now and consider in a future PR.

MilesCranmer:
Also I guess

```julia
struct WriteOnceReadMany{A}
    _raw_data::A
end
```
Might make more sense, as you could put dicts and vectors in the same type.

ven-k:
The `DynamicQuantitiesCollection` is similar to `WriteOnceReadMany`; except that one of its instance is an OrderedSet. I'll rename the former and drop the latter.

src/utils.jl
ven-k: Resolved https://github.com/SymbolicML/DynamicQuantities.jl/pull/107/commits/4ef0e83098e34666f46a866152ae183e98536168#diff-47c27891e951c8cd946b850dc2df31082624afdf57446c21cb6992f5f4b74aa2R442

src/utils.jl
ven-k: I've added a check here https://github.com/SymbolicML/DynamicQuantities.jl/pull/107/commits/54ee0f8d27f1bd89dfa237329449067a421004c5#diff-47c27891e951c8cd946b850dc2df31082624afdf57446c21cb6992f5f4b74aa2R459

MilesCranmer:
Thanks! I think we are almost done :)

A few more comments:

1. Could we do a test for whether this works for pre-compilation? For example, if a downstream package wants to register units like:

```julia
module MyLibrary

using DynamicQuantities: @register_unit, @u_str

@register_unit Wb u"m^2*kg*s^-2*A^-1"

end
```

and then, elsewhere, you use this like:

```julia
using DynamicQuantities
using MyLibrary

x = u"Wb"
y = us"Wb"
```

So it would be good to know whether pre-compilation works in this sort of context (not sure how to test this).

2.

I'm not sure whether we could actually unit test this, but I'm just curious how this would all work in the context of Revise.jl. If you are working on a package, would Revise.jl correctly register/un-register units as you add/remove them to a package?

ven-k:
I've added tests for _precompilation of register_unit in a closed module_.
And updating https://github.com/SciML/ModelingToolkitStandardLibrary.jl/pull/214 with new units[^1], works too.

[^1]:https://github.com/SciML/ModelingToolkitStandardLibrary.jl/pull/214/commits/17dc7cb4bf756db95268bf1c7454c4e6c4207f98

MilesCranmer (commented):
Looking good. Quick comment is I wonder if you could move the type definition of `WriteOnceReadMany` as well as its methods into a separate file, rather than mix them with `Quantity`? e.g., you will see that `FixedRational`, another internal type, gets its own file.

src/register_units.jl
```diff
@@ -0,0 +1,41 @@
...
+    index = get(ALL_MAPPING, name, INDEX_TYPE(0))
+    if iszero(index)
+        reg_expr = _lazy_register_unit(name, value)
+        push!(reg_expr.args, quote
+            $update_all_values($name_symbol, $value)
+            nothing
+        end)
+        return reg_expr
+    else
+        unit = ALL_VALUES[index]
```
MilesCranmer, line 34: This branch currently doesn’t show up in test coverage. Could it be included in tests just to verify it does what we expect?

src/register_units.jl
```diff
@@ -0,0 +1,41 @@
...
+        # When a utility function to expand `value` to its final form becomes
+        # available, enable the following check. This will avoid throwing an error
+        # if user is trying to register an existing unit with matching values.
+        # unit.value != value && throw("Unit $name is already defined as $unit")
```
MilesCranmer, line 38: I think this is old code? (Should be deleted)

ven-k: Actually no.
It's a note for future when `value` can be expanded as final form value without actually assigning to the `name`.
This is to just warn for `@register_unit m u"m"` instead of throwing an error.

ven-k:
I expected `types.jl` to house all types; now I've moved `WriteOnceReadMany` to its own file with this https://github.com/SymbolicML/DynamicQuantities.jl/pull/107/commits/5c00b031ee32ae1820d094404a7277ac13a0b9b0.

MilesCranmer:
Thanks!

(Currently types.jl is supposed to be “exported types” while internal types live in separate files. However I haven’t been completely consistent on this and a future refactor would certainly improve readability.)

MilesCranmer:
Cool. We're back at 100% test coverage. I did a deep review + some edits and everything is looking good now! Making some final tweaks and then will merge.

Thanks so much for the great contribution!!

MilesCranmer:
One thing I'm confused about is that the `@register_unit` does not seem to propagate to other modules. e.g., if I register a unit in `ExternalUnitRegistration.jl`, and load it with `using ExternalUnitRegistration`, I can't use it in my current module in `u"Wb"`... Why is that?

MilesCranmer (approved):
Everything looks good to me. I'm just super confused about the following behaviour:

```julia
julia> push!(LOAD_PATH, joinpath(@__DIR__, "test", "precompile_test"))
4-element Vector{String}:
 "@"
 "@v#.#"
 "@stdlib"
 "/Users/mcranmer/PermaDocuments/DynamicQuantities.jl/test/precompile_test"

julia> using ExternalUnitRegistration

julia> ExternalUnitRegistration.ALL_SYMBOLS
DynamicQuantities.WriteOnceReadMany{Vector{Symbol}}([:m, :g, :s, :A, :K, :cd, :mol, :fm, :pm, :nm  …  :L_sun, :L_bol0, :sigma_T, :au, :pc, :ly, :atm, :kpc, :Mpc, :Gpc])
```

Where is `:Wb`? Did the symbol list somehow get reset?

---

Maybe this is because I'm using Revise.jl?

ven-k:
That's an expected behavior (even without Revise).

So currently the externally defined units can be used with `u"unitname"` within the scope of the module (and its sub-modules by importing). Outside they should be treated as a var. However it can be used, divided, multiplied as any other unit (but without u"").

Ex: `@test Wb/u"m^2*kg*s^-2*A^-1" == 1.0` here[^1]

[^1]:https://github.com/SymbolicML/DynamicQuantities.jl/pull/107/files#diff-97287d1f47874f9eb568bc53640b753ce7f6cd5fd70a18180a4beb9dd8a7939bR1895

MilesCranmer:
Got it! I like that better actually, feels safer.

Issue #111, opened 2024-01-20
Base.isapprox fails for Vector{Quantity}


mikeingold:
Extending this from a comment in #76, I reduced the issue to a MWE. The issue seems to occur specifically when running `Base.isapprox` on a `Vector{Quantity}`.

```julia
using DynamicQuantities

a = 1.0u"m"
b = 2.0u"m"

a ≈ b
# Works ok: returns False

[a] ≈ [b]
#=
ERROR: Cannot create an additive identity for a `UnionAbstractQuantity` type, as the dimensions are unknown. Please use `zero(::UnionAbstractQuantity)` instead.
Stacktrace:
 [1] error(s::String)
   @ Base .\error.jl:35
 [2] zero(::Type{Quantity{Float64, Dimensions{DynamicQuantities.FixedRational{Int32, 25200}}}})
   @ DynamicQuantities C:\Users\mikei\.julia\packages\DynamicQuantities\jkfTz\src\utils.jl:280
 [3] real(T::Type)
   @ Base .\complex.jl:120
 [4] rtoldefault(x::Type{Quantity{Float64, Dimensions{…}}}, y::Type{Quantity{Float64, Dimensions{…}}}, atol::Int64)
   @ Base .\floatfuncs.jl:347
 [5] isapprox(x::Vector{Quantity{Float64, Dimensions{…}}}, y::Vector{Quantity{Float64, Dimensions{…}}})
   @ LinearAlgebra C:\Users\mikei\.julia\juliaup\julia-1.10.0+0.x64.w64.mingw32\share\julia\stdlib\v1.10\LinearAlgebra\src\generic.jl:1785
 [6] top-level scope
   @ REPL[8]:1
=#
```

mikeingold:
Looks like I mis-stated in #76: the method `isapprox(::AbstractArray, ::AbstractArray)` is implemented under **LinearAlgebra.jl** [here](https://github.com/JuliaLang/LinearAlgebra.jl/blob/8108bc4ff2dbd5b2143d6279efad640952f97f8e/src/generic.jl#L1768), currently:

```julia
function isapprox(x::AbstractArray, y::AbstractArray;
    atol::Real=0,
    rtol::Real=Base.rtoldefault(promote_leaf_eltypes(x),promote_leaf_eltypes(y),atol),
    nans::Bool=false, norm::Function=norm)
    d = norm(x - y)
    if isfinite(d)
[...]

PR #114, opened 2024-01-26
Improved Base.isapprox support

mikeingold:
This PR is intended to address concerns discussed in #111, specifically:
- Implements support for dimensionful types in the `atol` keyword argument for `Base.isapprox`
- Adds support for `isapprox(::AbstractVector{Q}, ::AbstractVector{Q})` where `Q` is a Quantity type (in the LinearAlgebra extension)

This would technically be considered a breaking change due to the behavior change for `atol` keyword arguments. Previously, this argument took a `::Real` and treated it as implicitly having the same units as the quantities being compared. As of this update, applicable units must be specified manually. This will make comparisons more explicit and less error prone. Additionally, this will enable comparisons like `isapprox(1.234u"m", 1234.0u"cm"; atol=1.0u"mm")`.

### DIFF
--- a/ext/DynamicQuantitiesLinearAlgebraExt.jl
+++ b/ext/DynamicQuantitiesLinearAlgebraExt.jl
@@ -5,4 +5,20 @@ import DynamicQuantities: UnionAbstractQuantity, ustrip, dimension, new_quantity
 
 norm(q::UnionAbstractQuantity, p::Real=2) = new_quantity(typeof(q), norm(ustrip(q), p), dimension(q))
 
+# Define isapprox for vectors of Quantity types
+function Base.isapprox(
+    u::AbstractArray{Q}, v::AbstractArray{Q};
+    atol=new_quantity(Q, zero(T), dimension(first(u))),
+    rtol=Base.rtoldefault(T),
+    nans::Bool=false, norm::Function=norm
+) where {T, D, Q<:UnionAbstractQuantity{T,D}}
+    d = norm(u - v)
+    if isfinite(d)
+        return d <= max(atol, rtol*max(norm(u), norm(v)))
+    else
+        # Fall back to a component-wise approximate comparison
+        return all(ab -> isapprox(uv[1], uv[2]; rtol=rtol, atol=atol, nans=nans), zip(u, v))
+    end
+end
+
 end
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -257,12 +257,11 @@ end
 
 # Define isapprox for vectors
 function Base.isapprox(
-    u::AbstractArray{<:UnionAbstractQuantity{T,D}},
-	v::AbstractArray{<:UnionAbstractQuantity{T,D}};
-    atol=new_quantity(T, zero(T), dimension(first(u))),
+    u::AbstractArray{Q}, v::AbstractArray{Q};
+    atol=new_quantity(Q, zero(T), dimension(first(u))),
     rtol=Base.rtoldefault(ustrip(first(u))),
     nans::Bool=false, norm::Function=norm
-) where {T,D}
+) where {T, D, Q<:UnionAbstractQuantity{T,D}}
     d = norm(u - v)
     if isfinite(d)
         return d <= max(atol, rtol*max(norm(u), norm(v)))
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -255,6 +255,23 @@ for (type, true_base_type, _) in ABSTRACT_QUANTITY_TYPES
     end
 end
 
+# Define isapprox for vectors
+function Base.isapprox(
+    u::AbstractArray{<:UnionAbstractQuantity{T,D}},
+	v::AbstractArray{<:UnionAbstractQuantity{T,D}};
+    atol=new_quantity(T, zero(T), dimension(first(u))),
+    rtol=Base.rtoldefault(ustrip(first(u))),
+    nans::Bool=false, norm::Function=norm
+) where {T,D}
+    d = norm(u - v)
+    if isfinite(d)
+        return d <= max(atol, rtol*max(norm(u), norm(v)))
+    else
+        # Fall back to a component-wise approximate comparison
+        return all(ab -> isapprox(uv[1], uv[2]; rtol=rtol, atol=atol, nans=nans), zip(u, v))
+    end
+end
+
 # Simple flags:
 for f in (
     :isone, :iszero, :isfinite, :isinf, :isnan, :isreal, :signbit,
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -169,14 +169,15 @@ for (type, _, _) in ABSTRACT_QUANTITY_TYPES
 end
 Base.keys(q::UnionAbstractQuantity) = keys(ustrip(q))
 
-# If atol specified in kwargs, check its dimensions and then strip units
+# If atol specified in kwargs, validate its dimensions and then strip units
 function _strip_kws(dimcheck, kws)
     if :atol in keys(kws)
         dimension(dimcheck) == dimension(kws[:atol]) || throw(DimensionError(dimcheck, kws[:atol]))
-        kws_dict = Dict{Symbol,Any}(kws)
-        kws_dict[:atol] = ustrip(kws_dict[:atol])
+        kws = Dict{Symbol,Any}(kws)
+        kws[:atol] = ustrip(kws[:atol])
     end
-    return kws_dict
+
+    return kws
 end
 
 # Numeric checks
@@ -233,9 +234,7 @@ for (type, true_base_type, _) in ABSTRACT_QUANTITY_TYPES
         function Base.isapprox(l::$type, r::$type; kws...)
             l, r = promote_except_value(l, r)
             dimension(l) == dimension(r) || throw(DimensionError(l, r))
-            new_kws = _strip_kws(l, kws)
-            @info new_kws
-            return isapprox(ustrip(l), ustrip(r); new_kws...)
+            return isapprox(ustrip(l), ustrip(r); _strip_kws(l, kws)...)
         end
         function Base.isapprox(l::$base_type, r::$type; kws...)
             iszero(dimension(r)) || throw(DimensionError(l, r))
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -173,11 +173,13 @@ Base.keys(q::UnionAbstractQuantity) = keys(ustrip(q))
 function _strip_kws(dimcheck, kws)
     if :atol in keys(kws)
         dimension(dimcheck) == dimension(kws[:atol]) || throw(DimensionError(dimcheck, kws[:atol]))
-        kws = replace(kws) do kv
+        new_kws = replace(kws) do kv
             first(kv) == :atol ? first(kv)=>ustrip(last(kv)) : kv
         end
     end
-    return kws
+    @info kws
+    @info new_kws
+    return new_kws
 end
 
 # Numeric checks
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -177,8 +177,6 @@ function _strip_kws(dimcheck, kws)
             first(kv) == :atol ? first(kv)=>ustrip(last(kv)) : kv
         end
     end
-    @info kws
-    @info new_kws
     return new_kws
 end
 
@@ -236,7 +234,9 @@ for (type, true_base_type, _) in ABSTRACT_QUANTITY_TYPES
         function Base.isapprox(l::$type, r::$type; kws...)
             l, r = promote_except_value(l, r)
             dimension(l) == dimension(r) || throw(DimensionError(l, r))
-            return isapprox(ustrip(l), ustrip(r); _strip_kws(l, kws)...)
+            new_kws = _strip_kws(l, kws)
+            @info new_kws
+            return isapprox(ustrip(l), ustrip(r); new_kws...)
         end
         function Base.isapprox(l::$base_type, r::$type; kws...)
             iszero(dimension(r)) || throw(DimensionError(l, r))
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -173,11 +173,10 @@ Base.keys(q::UnionAbstractQuantity) = keys(ustrip(q))
 function _strip_kws(dimcheck, kws)
     if :atol in keys(kws)
         dimension(dimcheck) == dimension(kws[:atol]) || throw(DimensionError(dimcheck, kws[:atol]))
-        new_kws = replace(kws) do kv
-            first(kv) == :atol ? first(kv)=>ustrip(last(kv)) : kv
-        end
+        kws_dict = Dict{Symbol,Any}(kws)
+        kws_dict[:atol] = ustrip(kws_dict[:atol])
     end
-    return new_kws
+    return kws_dict
 end
 
 # Numeric checks
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -259,7 +259,7 @@ end
 function Base.isapprox(
     u::AbstractArray{Q}, v::AbstractArray{Q};
     atol=new_quantity(Q, zero(T), dimension(first(u))),
-    rtol=Base.rtoldefault(ustrip(first(u))),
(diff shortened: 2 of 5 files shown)


src/utils.jl
```diff
@@ -169,6 +169,16 @@ for (type, _, _) in ABSTRACT_QUANTITY_TYPES
 end
 Base.keys(q::UnionAbstractQuantity) = keys(ustrip(q))
 
+# If atol specified in kwargs, validate its dimensions and then strip units
+function _strip_kws(dimcheck, kws)
```
MilesCranmer, line 173: Could we do a more descriptive name like `_validate_isapprox_kws`?

src/utils.jl
```diff
@@ -169,6 +169,16 @@ for (type, _, _) in ABSTRACT_QUANTITY_TYPES
...
+    if :atol in keys(kws)
+        dimension(dimcheck) == dimension(kws[:atol]) || throw(DimensionError(dimcheck, kws[:atol]))
+        kws = Dict{Symbol,Any}(kws)
```
MilesCranmer, line 176: Wont this create extra allocations? i think it should be a named tuple like `kws = (; kws…, atol=ustrip(kws[:atol]))`

src/utils.jl
```diff
@@ -169,6 +169,16 @@ for (type, _, _) in ABSTRACT_QUANTITY_TYPES
 end
 Base.keys(q::UnionAbstractQuantity) = keys(ustrip(q))
 
+# If atol specified in kwargs, validate its dimensions and then strip units
+function _strip_kws(dimcheck, kws)
+    if :atol in keys(kws)
```
MilesCranmer, line 174: Should be faster to do `haskey(kws, :atol)`

src/utils.jl
MilesCranmer: Also should add an `@inline` to help force it

test/test_unitful.jl
```diff
@@ -14,13 +14,15 @@ end
...
     x_unitful = T(0.2)u"m*mol^2*A^(-1//2)*cd^(2//5)"
+    tol_unitful = T(1e-6)u"m*mol^2*A^(-1//2)*cd^(2//5)"
 
     @test risapprox(convert(Unitful.Quantity, x), x_unitful; atol=1e-6)
     @test typeof(convert(DynamicQuantities.Quantity, convert(Unitful.Quantity, x))) <: DynamicQuantities.Quantity{T,DynamicQuantities.DEFAULT_DIM_TYPE}
-    @test isapprox(convert(DynamicQuantities.Quantity, convert(Unitful.Quantity, x)), x; atol=1e-6)
+    @test isapprox(convert(DynamicQuantities.Quantity, convert(Unitful.Quantity, x)), x; atol=tol_dq)
 
-    @test isapprox(convert(DynamicQuantities.Quantity{T,D}, x_unitful), x; atol=1e-6)
+    @test isapprox(convert(DynamicQuantities.Quantity{T,D}, x_unitful), x; atol=tol_dq)
```
MilesCranmer, line 25: Can you add some tests of the errors, e.g., `@test_throws`?

ext/DynamicQuantitiesLinearAlgebraExt.jl
```diff
@@ -5,4 +5,20 @@ import DynamicQuantities: UnionAbstractQuantity, ustrip, dimension, new_quantity
 
 norm(q::UnionAbstractQuantity, p::Real=2) = new_quantity(typeof(q), norm(ustrip(q), p), dimension(q))
 
+# Define isapprox for vectors of Quantity types
+function Base.isapprox(
+    u::AbstractArray{Q}, v::AbstractArray{Q};
```
MilesCranmer, line 10: Will promotion rules automatically run beforehand? Otherwise should add another method for `Q1` and `Q2`. Also for `QuantityArray` and `AbstractArray`

ext/DynamicQuantitiesLinearAlgebraExt.jl
```diff
@@ -5,4 +5,20 @@ import DynamicQuantities: UnionAbstractQuantity, ustrip, dimension, new_quantity
...
+    atol=new_quantity(Q, zero(T), dimension(first(u))),
+    rtol=Base.rtoldefault(T),
+    nans::Bool=false, norm::Function=norm
+) where {T, D, Q<:UnionAbstractQuantity{T,D}}
+    d = norm(u - v)
+    if isfinite(d)
+        return d <= max(atol, rtol*max(norm(u), norm(v)))
+    else
+        # Fall back to a component-wise approximate comparison
+        return all(ab -> isapprox(uv[1], uv[2]; rtol=rtol, atol=atol, nans=nans), zip(u, v))
```
MilesCranmer, line 20: Is this what normal isapprox does? Maybe add a test to assert the behavior is the same?

src/utils.jl
mikeingold: I can take a look at that idea. I ran into two issues here that led to this version:
- The `kws` produced when args are passed is immutable so it couldn’t be changed in-place
- If e.g. atol is the only keyword arg passed, then `kws` is a set of Pair’s `Symbol => Quantity`. I tried running a replace operation here instead but found that the ustrip result was by default promoted to a dimensionless Quantity, which generates an error.

MilesCranmer: NamedTuples let you change at construction time. Whatever is given last will take priority in case of duplicate keys:

```julia
julia> kws = (;atol=1.0)
(atol = 1.0,)

julia> (; kws..., atol=2.0)
(atol = 2.0,)
```

And there are zero allocations here.

mikeingold: Oh, nice! Wish I had known about this as it would’ve saved me a good bit of troubleshooting. As long as the Pair types change to prevent the stripped result from promoting back into a Quantity it should work great.

src/utils.jl
mikeingold: Just commit an update with these comments addressed.

src/utils.jl
mikeingold: Just commit an update with these comments addressed.

src/utils.jl
mikeingold: Updated.

ext/DynamicQuantitiesLinearAlgebraExt.jl
mikeingold: The standard `Base.isapprox` method for vectors is defined by LinearAlgebra [here](https://github.com/JuliaLang/LinearAlgebra.jl/blob/8108bc4ff2dbd5b2143d6279efad640952f97f8e/src/generic.jl#L1768). This method for dimensionful types is pretty much a direct translation of that code but fixes `atol` and `rtol` handling.

mikeingold:
Sorry, I’ve been swamped the past few weeks. I’m going to try to make a final push on this PR in the coming days.

test/test_unitful.jl
mikeingold: Added

mikeingold:
I just sync'd my fork with upstream main and merged all of the updates into this branch -- no conflicts.

ext/DynamicQuantitiesLinearAlgebraExt.jl
mikeingold: I've updated this section to include versions specialized for `QuantityArray` and `AbstractArray` allowing for type-inequality between the arguments. This commit also added unit tests to exercise both of these method definitions, and all tests currently pass.

mikeingold:
I think this PR is in pretty good shape. Let me know if you have any more feedback that I can address.

Reminder: this PR would technically be considered breaking due to the change in `atol` behavior (requires dimensionful quantity to eliminate ambiguity and a potential foot-gun).

MilesCranmer:
Thanks, will take a look now!

MilesCranmer:
It seems like loading LinearAlgebra will change the behavior of isapprox on arrays:

```julia

julia> x = [u"m/s" for _ in 1:3]
3-element Vector{Quantity{Float64, Dimensions{DynamicQuantities.FixedRational{Int32, 25200}}}}:
 1.0 m s⁻¹
 1.0 m s⁻¹
 1.0 m s⁻¹

julia> y = [u"m/s" for _ in 1:3]
3-element Vector{Quantity{Float64, Dimensions{DynamicQuantities.FixedRational{Int32, 25200}}}}:
 1.0 m s⁻¹
 1.0 m s⁻¹
 1.0 m s⁻¹

julia> isapprox(x, y)
ERROR: Cannot create an additive identity for a `UnionAbstractQuantity` type, as the dimensions are unknown. Please use `zero(::UnionAbstractQuantity)` instead.
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] zero(::Type{Quantity{Float64, Dimensions{DynamicQuantities.FixedRational{Int32, 25200}}}})
   @ DynamicQuantities ~/PermaDocuments/SymbolicRegressionMonorepo/DynamicQuantities.jl/src/utils.jl:289
 [3] real(T::Type)
   @ Base ./complex.jl:120
 [4] rtoldefault(x::Type{Quantity{Float64, Dimensions{DynamicQuantities.FixedRational{Int32, 25200}}}}, y::Type{Quantity{Float64, Dimensions{DynamicQuantities.FixedRational{Int32, 25200}}}}, atol::Int64)
   @ Base ./floatfuncs.jl:263
 [5] isapprox(x::Vector{Quantity{Float64, Dimensions{DynamicQuantities.FixedRational{Int32, 25200}}}}, y::Vector{Quantity{Float64, Dimensions{DynamicQuantities.FixedRational{Int32, 25200}}}})
   @ LinearAlgebra ~/Documents/julia/usr/share/julia/stdlib/v1.11/LinearAlgebra/src/generic.jl:1802
 [6] top-level scope
   @ REPL[4]:1

julia> using LinearAlgebra

julia> isapprox(x, y)
true
```

I think for this we should just define is isapprox method within the normal codebase, but throw an error if there is any code involving LinearAlgebra, telling the user to load it. I'll try to push a fix for this.

src/arrays.jl
```diff
@@ -320,6 +320,46 @@ end
...
+# Define isapprox for vectors of Quantity's
+function Base.isapprox(
+    u::AbstractArray{Q1},
+    v::AbstractArray{Q2};
+    atol=new_quantity(Q1, zero(T1), dimension(first(u))),
+    rtol=Base.rtoldefault(T1),
+    nans::Bool=false,
+    norm::F=_norm
+) where {T1,D1,Q1<:UnionAbstractQuantity{T1,D1},T2,D2,Q2<:UnionAbstractQuantity{T2,D2},F<:Function}
+    d = norm(u - v)
```
MilesCranmer, line 335: Presumably we also want to fall back to component-wise comparison when dimensions are not equal across the array?

src/arrays.jl
```diff
@@ -320,6 +320,46 @@ end
...
+    if isfinite(d)
+        return d <= max(atol, rtol*max(norm(u), norm(v)))
+    else
+        # Fall back to a component-wise approximate comparison
+        return all(ab -> isapprox(uv[1], uv[2]; rtol=rtol, atol=atol, nans=nans), zip(u, v))
```
MilesCranmer, line 340: The `zip` is dangerous because if you zip two arrays which have different lengths, it won't throw an error, it will just go until one of the arrays runs out of elements. I think `eachindex(u, v)` is safer.

MilesCranmer:
It looks like `rtoldefault` is actually a 3-argument function:

```julia
function rtoldefault(x::Union{T,Type{T}}, y::Union{S,Type{S}}, atol::Real) where {T<:Number,S<:Number}
    rtol = max(rtoldefault(real(T)),rtoldefault(real(S)))
    return atol > 0 ? zero(rtol) : rtol
end
```

Is this behavior accounted for in the current code?

src/arrays.jl
MilesCranmer: Pushed code to fix this.

src/arrays.jl
MilesCranmer: Pushed code for this.

test/unittests.jl
```diff
@@ -288,6 +288,12 @@ end
         uX = X .* Quantity(2, length=2.5, luminosity=0.5)
         @test sum(X) == 0.5 * ustrip(sum(uX))
 
+        @test isapprox(uX, uX, atol=Quantity{T,D}(1e-6, length=2.5, luminosity=0.5))
+        @test_throws DimensionError isapprox(uX, uX, atol=1e-6)
+        @test_throws ErrorException DynamicQuantities._norm(1.0)
+        VERSION >= v"1.9" &&
+            @test_throws "Please load the `LinearAlgebra.jl` package." DynamicQuantities._norm(1.0)
```
MilesCranmer, line 295: Need test with mixed unit array, like

```julia
julia> x = [1.0u"m", 1.2u"s"]
```

which should be approximately equal to

```julia
julia> y = [1.00000001u"m", 1.2u"s"]
```

MilesCranmer: (Although here it should not be possible to provide an `atol`; it should fall back to `rtol`)

Or maybe this doesn't even make sense to check for...?

MilesCranmer:
Okay I made some changes. Let me know what you think. Ideally we could also have tests for all the branches of code, such as atol/rtol set explicitly/left as default (with expected `isapprox = true` and `= false`) just so we know it works exactly as expected.

src/arrays.jl
mikeingold: The way that I laid out the specialized `isapprox` methods in this branch was intended to mirror how things are defined upstream.

- Julia Base only defines `isapprox` as a 1:1 scalar comparison and doesn't attempt to handle vector comparisons, but users are free to perform component-wise comparisons via broadcasting.
- LinearAlgebra.jl introduces a method for vector comparison, where the delta isn't component-wise, rather it assesses similarity based on a threshold for the `norm` of the "tip-to-tip" vector difference. This works well for vectors composed of simple scalar values and vector spaces with homogeneous units/dimensions, but isn't well-defined for vectors containing heterogeneous units/dimensions (what's the norm?).

Based on these upstream patterns and this mixed-units issue, I feel like not defining a component-wise fallback is probably the "correct" solution. That being said, the pragmatic side of me recognizes that users probably aren't thinking this deeply about what it means for two vectors to be approximately equal, and so would be confused by a `DimensionError` when trying to compute an `isapprox` without broadcasting.

src/arrays.jl
mikeingold: This pattern is pretty much directly ripped from the method in LinearAlgebra.jl. This case is something I'd thought about, but you only reach this line after completing the previous `d = norm(u - v)` calculation, where non-equal lengths would definitely be caught.

I can take a look at making this change. If we have a better way, I think I'd also recommend we make the same argument upstream in a PR to LinearAlgebra.jl.

src/arrays.jl
MilesCranmer: > Based on these upstream patterns and this mixed-units issue, I feel like not defining a component-wise fallback is probably the "correct" solution.

I agree with this. Users can always do `.≈` if needed. We can also just tag the `DimensionError` with some extra debugging info for users. Generally I do think its better to error loudly and obviously (with good error messages).

test/unittests.jl
MilesCranmer: Will remove this if we want to avoid the per-component fallback (which is probably more expensive to check for anyways).

MilesCranmer:
Okay I will implement the changes to avoid the per-component fallback. Give me a bit and I will declare when I'm done.

test/unittests.jl
mikeingold: I wrote a comment on one of the reviews above describing my thoughts on mixed-unit vectors. This is one of the complications potentially need to address, that any `atol` argument is pretty much guaranteed to produce a `DimensionError`. Really the only way to handle mixed-units is probably to just internally broadcast the operation. We could also try to catch the condition and generate a descriptive error suggesting they use broadcasting or similar.

MilesCranmer:
Ok, implemented. Let me know what you think.

mikeingold:
Looks good to me. Nice work!

MilesCranmer (approved):

MilesCranmer:
Thanks again for pushing this through!

PR #115, opened 2024-02-12
Add `webers`, `henry`, and `siemens`

ven-k:
These are quite commonly used units and will be useful to house them in here.

I didn't see any testset for derived units. LMK if & how I should add tests for these four.

### DIFF
--- a/src/units.jl
+++ b/src/units.jl
@@ -100,6 +100,7 @@ end
 @_lazy_register_unit Ω V / A
 @_lazy_register_unit ohm Ω
 @_lazy_register_unit T N / (A * m)
+@_lazy_register_unit Wb m^2 * kg * s^-2 * A^-1
 
 @add_prefixes Hz (n, μ, u, m, k, M, G)
 @add_prefixes N ()
@@ -112,6 +113,7 @@ end
 @add_prefixes Ω (n, μ, u, m, k, M, G)
 @add_prefixes ohm (n, μ, u, m, k, M, G)
 @add_prefixes T ()
+@add_prefixes Wb (n, μ, u, m)
 
 # SI derived units
 @doc(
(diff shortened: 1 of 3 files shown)


MilesCranmer:
I'm happy with the others but radians are really tricky and are probably left for later.

MilesCranmer:
For example, what should `sind(1u"rad")` do? Right now with this PR it would evaluate as the sinusoid of 1 degree, even though it was clearly written as `1u"rad"`... So basically needs a lot of work to get right.

src/units.jl
```diff
@@ -110,7 +113,10 @@ end
 @add_prefixes F (f, p, n, μ, u, m)
 @add_prefixes Ω (n, μ, u, m, k, M, G)
 @add_prefixes ohm (n, μ, u, m, k, M, G)
+@add_prefixes S (n, μ, u, m, k, M, G)
```
MilesCranmer, line 116: What I’ve done so far (to prevent bloat) is only put in the units that people are actually using. What prefixes do you actually use in code? Is it just `S` or do you have examples of the use of prefixes?

ven-k: I'm keeping it on par with ohm as S = 1/ohm

MilesCranmer: Oh I see. That makes sense. Maybe for the other ones though, reduce the number of prefixes? For all of the units my strategy has basically been to wait until someone wants a given prefix and then add it.

ven-k: I've dropped all the prefixes of henry.
For webers, m, n, and u are commonly used; I've dropped the rest.

MilesCranmer: Cool, thanks!

MilesCranmer (approved):

PR #117, opened 2024-03-02
Add angular units

### LINKED ISSUE #28: Radian/Steradian
Radians and steradians are dimensionless, but really should be treated like units. So far they have been excluded from the library (along with their derived units, like lumens and lux), but it could be useful to include them somehow. Maybe once #24 merges, there could be an alternate `RadQuantity`, `RadDimensions` objects for propagating these units, with promotion properties from regular `Quantity` objects so the units submodule can be used as normal.

MilesCranmer:
Fixes #28 

@mikeingold since you have interest in using this, what do you think? 

Here I am using the SI definition that `1 rad = 1`. So to have "rad" actually show up in quantities you would need to use symbolic units, like the following:

```julia
julia> θ = 10.2us"deg"
10.2 deg

julia> r = 5us"km"
5.0 km

julia> θ * r |> uexpand
890.117918517108 m
```

which I think seems okay?

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,7 +258,7 @@ so you can simply access these as `SymbolicUnits.cm` and `SymbolicConstants.h`,
 respectively.
 
 
-### Custom Units
+#### Custom Units
 
 You can create custom units with the `@register_unit` macro:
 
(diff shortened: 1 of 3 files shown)


MilesCranmer:
Cool. Thanks! And yeah it looks like Unitful does this the same way so all is good.

PR #133, opened 2024-04-22
fix tests on 1.11

KristofferC:
A refactoring in the `round` function caused this method definition to be ambiguous.


MilesCranmer:
Thanks!

PR #153, opened 2024-10-25
Add mu, m, k prefixes for Newtons

archermarx:
In my field (spacecraft propulsion), we regularly express thrusts in terms of micro-, milli-, and kilo- Newtons. This PR adds those prefixes.


MilesCranmer (approved):
Thanks!

PR #152, opened 2024-09-06
adds Gauss as unit for magnetic flux densities

bodokaiser:
Related #151

Issue #151, opened 2024-09-05
@register_unit not global


bodokaiser:
Hi all,

I am in the process of upgrading the Coils.jl package from `Unitful` to `DynamicQuantities` and register Gauss as units:

```julia
using DynamicQuantities: @register_unit, @u_str

@register_unit Gauss 1e-4u"T"
```

However, when I run my package tests, I need to re-register the unit; otherwise, I get "LoadError: ArgumentError: Symbol Gauss not found in `Units` or `Constants`."

I think this is not what one would expect and its a bit annoying for users to have to re-register the units themselves.

MilesCranmer:
Do you want to add Gauss to the units? I didn’t add it to emphasise that DQ is SI based (so nobody mistakenly uses the CGS version of Maxwells equations) but we could explain this in the docs.

The `@register_unit` acting locally rather than globally is a feature rather than a bug. Though it can sometimes be convenient, the global behavior of Unitful in registering units is very dangerous and opens up all sorts of footguns and “spooky action at a distance”.

bodokaiser:
Hi Miles, we should allow people to register their own defined units globally or support the common units. If you prefer to add more units to the package, let's do that!

In the atom physics community, "Gauss" is the standard unit for magnetic fields, and with `Coils.jl`, we can use Julia to design coil geometries to create magnetic fields in atom physics experiments.

What do Maxwell's equations have to do with `DynamicQuantities`, or is that another package of yours that strongly uses `DynamicQuantities`?

MilesCranmer:
They key part of the wiki article is:

> Conversions between quantities in Gaussian and SI units are not direct unit conversions, because the quantities themselves are defined differently in each system. This means that the equations expressing physical laws of electromagnetism—such as [Maxwell's equations](https://en.wikipedia.org/api/rest_v1/page/mobile-html/Maxwell's_equations)—will change depending on the system of units employed. As an example, quantities that are [dimensionless](https://en.wikipedia.org/api/rest_v1/page/mobile-html/Dimensionless) in one system may have dimension in the other.

bodokaiser:
Thanks for sharing the details ☺️. Can I prepare a PR to add Gauss?

PR #154, opened 2024-11-01
Add some common units

Ickaser:
In the spirit of #153 and #152 , adding some more units with prefixes, since I ran into the same situation as #151 of wanting to register new units.

Specifically, adding atmosphere (atm), Torr, and cal as pressure and energy units respectively. The fact that people in my field commonly use these somewhat-annoying units is one of the main reasons I use packages like this and Unitful.

I would be willing to go through e.g. https://github.com/PainterQubits/Unitful.jl/blob/master/src/pkgdefaults.jl and add more of the default units listed there, to avoid needing a multitude of this kind of pull request.

### DIFF
--- a/src/units.jl
+++ b/src/units.jl
@@ -263,7 +263,7 @@ end
 ## Energy
 @_lazy_register_unit cal 4.184 * J
 
-@add_prefixes cal (k)
+@add_prefixes cal (k,)
 
 @doc(
     "Energy in calories. Available variants: `kcal`.",
--- a/src/units.jl
+++ b/src/units.jl
@@ -214,12 +214,20 @@ end
 
 ## Pressure
 @_lazy_register_unit bar 100 * kPa
+@_lazy_register_unit atm 101325 * Pa
+@_lazy_register_unit Torr 101325//760 * Pa
 
 @add_prefixes bar (m,)
+@add_prefixes atm ()
+@add_prefixes Torr (m,)
 
 @doc(
     "Pressure in bars. Available variants: `mbar`.",
     bar,
+    "Pressure in atmospheres.",
+    atm,
+    "Pressure in Torr. Available variants: `mTorr`.",
+    Torr,
 )
 
 ## Angles
@@ -251,6 +259,16 @@ end
 
 # Do not wish to define Gaussian units, as it changes
 # some formulas. Safer to force user to work exclusively in one unit system.
+
+## Energy
+@_lazy_register_unit cal 4.184 * J
+
+@add_prefixes cal (k)
+
+@doc(
+    "Energy in calories. Available variants: `kcal`.",
+    cal,
+)
 
 # Do not wish to define physical constants, as the number of symbols might lead to ambiguity.
 # The user should define these instead.
--- a/src/units.jl
+++ b/src/units.jl
@@ -215,7 +215,7 @@ end
 ## Pressure
 @_lazy_register_unit bar 100 * kPa
 @_lazy_register_unit Torr 101325//760 * Pa
-@_lazy_register_unit mmHg 101325//760 * Pa
+@_lazy_register_unit mmHg Torr
 
 @add_prefixes bar (m,)
 @add_prefixes Torr (m,)
@@ -224,8 +224,12 @@ end
 @doc(
     "Pressure in bars. Available variants: `mbar`.",
     bar,
+)
+@doc(
     "Pressure in Torr. Available variants: `mTorr`.",
     Torr,
+)
+@doc(
     "Pressure in mmHg.",
     mmHg,
 )
--- a/src/units.jl
+++ b/src/units.jl
@@ -214,20 +214,20 @@ end
 
 ## Pressure
 @_lazy_register_unit bar 100 * kPa
-@_lazy_register_unit atm 101325 * Pa
 @_lazy_register_unit Torr 101325//760 * Pa
+@_lazy_register_unit mmHg 101325//760 * Pa
 
 @add_prefixes bar (m,)
-@add_prefixes atm ()
 @add_prefixes Torr (m,)
+@add_prefixes mmHg ()
 
 @doc(
     "Pressure in bars. Available variants: `mbar`.",
     bar,
-    "Pressure in atmospheres.",
-    atm,
     "Pressure in Torr. Available variants: `mTorr`.",
     Torr,
+    "Pressure in mmHg.",
+    mmHg,
 )
 
 ## Angles
@@ -259,16 +259,6 @@ end
 
 # Do not wish to define Gaussian units, as it changes
 # some formulas. Safer to force user to work exclusively in one unit system.
-
-## Energy
-@_lazy_register_unit cal 4.184 * J
-
-@add_prefixes cal (k,)
-
-@doc(
-    "Energy in calories. Available variants: `kcal`.",
-    cal,
-)
 
 # Do not wish to define physical constants, as the number of symbols might lead to ambiguity.
 # The user should define these instead.


src/units.jl
```diff
@@ -252,6 +260,16 @@ end
 # Do not wish to define Gaussian units, as it changes
 # some formulas. Safer to force user to work exclusively in one unit system.
 
+## Energy
+@_lazy_register_unit cal 4.184 * J
```
MilesCranmer, line 264: How important are calories to you? Since the unit is ambiguous ("small calorie" vs "large calorie") I wonder if we should instead have the user use only Joules, which is more explicit.

MilesCranmer:
> I would be willing to go through e.g. https://github.com/PainterQubits/Unitful.jl/blob/master/src/pkgdefaults.jl and add more of the default units listed there, to avoid needing a multitude of this kind of pull request.

Basically the way I am adding units is only if people have a specific need for them. So please only request units and prefixes that you use in your own work. This is to prevent bloating the library with units that are never used. Hopefully that sounds reasonable.

Ickaser:
> > I would be willing to go through e.g. https://github.com/PainterQubits/Unitful.jl/blob/master/src/pkgdefaults.jl and add more of the default units listed there, to avoid needing a multitude of this kind of pull request.
> 
> Basically the way I am adding units is only if people have a specific need for them. So please only request units and prefixes that you use in your own work. This is to prevent bloating the library with units that are never used. Hopefully that sounds reasonable.

I understand that philosophy, and the rationale for not allowing global unit registering as you explained at #151 . But it does become a barrier to switching away from e.g. Unitful or pint, where I have basically never needed to register new units because these kinds of units are already defined; speaking as a chemical engineering student, one of the main reasons I want to use a package like this is that there are still quite a few cases where fields still have not fully made the switch to SI units. If all I had to deal with were plain SI units, I would just write everything in SI marks and not worry about conversions. Verifying correct dimensionality comes as a nice sanity check, but it's not what attracts me to the package.

There is also some amount of friction, where if I often need to deal with a unit (e.g. cal) I have to ask at what level I should be registering the unit. In my startup.jl? That seems weird, and doesn't port well to other people. At the start of every script? It's annoying to have that added to my boilerplate, etc.

Strictly speaking, I only need Torr, mTorr and cal, so I can take atmospheres out of this pull request if you prefer. But in general, I think a units package is more useful if you don't need to define your own units so often.

MilesCranmer:
Oops I just remembered, atm is already in the package but is listed in the constants module: https://github.com/SymbolicML/DynamicQuantities.jl/blob/659302918140400a7d13ca26e5d02b60bb941474/src/constants.jl#L161

MilesCranmer:
Do you use cal or kcal? I would feel more comfortable adding kcal and avoiding cal entirely.

MilesCranmer:
> doesn't port well to other people

I think for cases like this, this is a feature, not a bug. We don’t want to have a definition of cal that is unexpected — it could cause calculations to be wrong but in subtle ways. Sometimes safety needs to be emphasised in the design over convenience

MilesCranmer:
Another way of saying this: it's always better to have something fail loudly

Ickaser:
> Do you use cal or kcal? I would feel more comfortable adding kcal and avoiding cal entirely, due to the ambiguity.

Like I said, we have a heat transfer coefficient given in cal/s/K/cm^2 in even recent papers, with a history of those units in the field. Using kcal would be less ambiguous, but wouldn't achieve my goal here, unfortunately.

The US usage of "nutritional calorie" as 4184 J is, I feel, restricted to nutrition, and even reading the entire Wikipedia article it seems clear to me that the authors of the article had bought into the general scientific definition of 4.184 J. It's the kind of thing that got brought up in my high school chemistry class, so to me it feels like, technically, a possible footgun but not one that I have ever seen as a real problem. (In part, that's because the food definition of a "calorie" is not exactly correlated to what a biological system can use, so what you read on food packaging isn't ever really convertible to other units). 

All that said, if the risk of someone trying to do math with the values reported on a US food label is more than you are willing to accept, then I can try to find a way to reasonably register calories for my scripts (or just default back to Unitful).

MilesCranmer:
I think nutritional calories are for more commonly used than the gram calorie, as it's literally on every nutrition label in the US. It just seems like a massive footgun to me to say that this calorie, that every adult in the US intuitively understands, is different from the one used in this package. Even if other units packages do this, I think it's a mistake. I just don't see convenience for a subset of users as being a strong enough argument for introducing this in a package that targets a broad user base.

The `@register_unit` macro is built for this exact purpose though! Is there any fundamental reason you wish to avoid it? You can also just do

```julia
const cal = 4.184u"J"
```

I'm happy to add `Torr` though. Also note that `atm` is already in the package in the constants submodule - though perhaps it could be "aliased" to the units module somehow. But the current code in this PR will error since `atm` is already a symbol.

MilesCranmer:
While we are at it I suppose we should add mmHg as well.

Ickaser:
> I think nutritional calories are for more commonly used than the gram calorie, as it's literally on every nutrition label in the US. It just seems like a massive footgun to me to say that this calorie, that every adult in the US intuitively understands, is different from the one used in this package. Even if other units packages do this, I think it's a mistake. I just don't see convenience for a subset of users as being a strong enough argument for introducing this in a package that targets a broad user base.
> 
> The `@register_unit` macro is built for this exact purpose though! Is there any fundamental reason you wish to avoid it? You can also just do
> 
> ```julia
> const cal = 4.184u"J"
> ```
> 
> I'm happy to add `Torr` though. Also note that `atm` is already in the package in the constants submodule - though perhaps it could be "aliased" to the units module somehow. But the current code in this PR will error since `atm` is already a symbol.

I suppose, when it comes down to it, the main thing keeping me from just registering the unit myself and going forward is an aesthetic issue; I like that having all my units wrapped in `u""` makes it easy to pick them out and doesn't clutter the namespace too much. But if it's just for calories, I can register the unit within my base package (which gets imported by my other projects), export it, and just use the `cal` symbol instead of `u"cal"`. That's not too onerous.

(I'm guessing that the scoping mechanism built into `@u_str` couldn't be adjusted to recognize that a symbol `cal` has been registered in a package and exported, and pick it up? That sounds like a complicated procedure.)

I appreciate your willingness to talk me through this--after thinking it through, I can live with `cal` being its own symbol that lives outside of `u""`.

> Oops I just remembered, atm is already in the package but is listed in the constants module:
> 
> https://github.com/SymbolicML/DynamicQuantities.jl/blob/659302918140400a7d13ca26e5d02b60bb941474/src/constants.jl#L161

I'm used to thinking of `atm` as a unit of pressure, rather than a measured constant, since actual atmospheric pressure varies so much from that single value. But for the sake of moving this along I'm happy to leave it in the constants module if that's what you prefer.

Since I'm adding `mTorr` (which would be equivalent to `umHg`), if we add `mmHg`, should I add it as `mHg` with prefixes `u` and `m` allowed? I do hear elder statesmen in my field talk about "microns" meaning "umHg" sometimes, but `mTorr` is probably enough to cover that unit anyway.

src/units.jl
```diff
@@ -214,12 +214,20 @@ end
...
+@_lazy_register_unit mmHg 101325//760 * Pa
 
 @add_prefixes bar (m,)
+@add_prefixes Torr (m,)
+@add_prefixes mmHg ()
 
 @doc(
     "Pressure in bars. Available variants: `mbar`.",
     bar,
+    "Pressure in Torr. Available variants: `mTorr`.",
```
MilesCranmer, line 227: Can you make this a new `@doc` call? Just to keep the same style as elsewhere

src/units.jl
```diff
@@ -214,12 +214,20 @@ end
 
 ## Pressure
 @_lazy_register_unit bar 100 * kPa
+@_lazy_register_unit Torr 101325//760 * Pa
+@_lazy_register_unit mmHg 101325//760 * Pa
```
MilesCranmer, line 218:
```suggestion
@_lazy_register_unit mmHg Torr
```

MilesCranmer (approved):
Thanks for the contribution!

PR #155, opened 2024-11-01
Deprecate Julia 1.9

MilesCranmer:


MilesCranmer:
Performance looks fine in a local benchmark, so it's probably just the github action

PR #158, opened 2024-11-19
fix: applying units to range

### LINKED ISSUE #157: Can't apply units to a range
I think this should work:
```julia
julia> (1:0.25:4)u"inch"
ERROR: Cannot create an additive identity from `Type{<:Quantity}`, as the dimensions are unknown. Please use `zero(::Quantity)` instead.

julia> collect(1:0.25:4)u"inch"
13-element Vector{Quantity{Float64, Dimensions{FixedRational{Int32, 25200}}}}:
 0.025400000000000002 m
              0.03175 m
               0.0381 m
              0.04445 m
...
```
Came up in https://github.com/JuliaLang/julia/issues/56610, which is about `1u"inch":0.25u"inch":4u"inch"`. Constructing a range like that will (bugs aside) circumvent all the "do what I mean" floating-point magic of `0:0.1:1`. But making the range first & then applying units would mean the magic applies to the numbers as written.

MilesCranmer:
fixes #157

@mcabbott could you take a look?

### DIFF
--- a/src/math.jl
+++ b/src/math.jl
@@ -58,10 +58,9 @@ Base.:/(l::AbstractDimensions, r::AbstractDimensions) = map_dimensions(-, l, r)
 # Multiplying ranges with units
 for (type, _, _) in ABSTRACT_QUANTITY_TYPES
     @eval begin
-        Base.:*(r::UnitRange, q::$type) = (first(r) * q):(last(r) * q)
         Base.:*(r::StepRange, q::$type) = (first(r) * q):(Base.step(r) * q):(last(r) * q)
         Base.:*(r::StepRangeLen, q::$type) = range(first(r) * q; step=Base.step(r) * q, length=length(r))
-        Base.:*(q::$type, r::Union{UnitRange,StepRange,StepRangeLen}) = r * q
+        Base.:*(q::$type, r::Union{StepRange,StepRangeLen}) = r * q
     end
 end
 
--- a/src/math.jl
+++ b/src/math.jl
@@ -60,6 +60,27 @@ Base.TwicePrecision{T}(x::T) where {T<:AbstractQuantity} = Base.TwicePrecision{t
 # TODO: Note that to get RealQuantity working, we have to overload many other functions,
 #       which is why we skip it.
 
+# Avoid https://github.com/JuliaLang/julia/issues/56610
+for T1 in (AbstractQuantity{<:Real}, Real),
+    T2 in (AbstractQuantity{<:Real}, Real),
+    T3 in (AbstractQuantity{<:Real}, Real)
+
+    T1 === T2 === T3 === Real && continue
+
+    @eval function Base.:(:)(start::$T1, step::$T2, stop::$T3)
+        dimension(start) == dimension(step) || throw(DimensionError(start, step))
+        dimension(start) == dimension(stop) || throw(DimensionError(start, stop))
+        return range(start, stop, length=length(ustrip(start):ustrip(step):ustrip(stop)))
+    end
+
+    if T3 === Real && !(T1 === T2 === Real)
+        @eval function Base.:(:)(::$T1, ::$T2)
+            error("When creating a range over quantities, you must specify a step.")
+        end
+    end
+end
+
+
 # Defines +, -, and mod
 for (type, true_base_type, _) in ABSTRACT_QUANTITY_TYPES, op in (:+, :-, :mod)
     # Only define `mod` on `Number` types:
--- a/src/math.jl
+++ b/src/math.jl
@@ -74,8 +74,11 @@ for T1 in (AbstractQuantity{<:Real}, Real),
     end
 
     if T3 === Real && !(T1 === T2 === Real)
-        @eval function Base.:(:)(::$T1, ::$T2)
-            error("When creating a range over quantities, you must specify a step.")
+        @eval function Base.:(:)(start::$T1, stop::$T2)
+            if !iszero(dimension(start)) || !iszero(dimension(stop))
+                error("When creating a range over dimensionful quantities, you must specify a step.")
+            end
+            return start:1:stop
         end
     end
 end
--- a/src/math.jl
+++ b/src/math.jl
@@ -56,12 +56,9 @@ Base.:*(l::AbstractDimensions, r::AbstractDimensions) = map_dimensions(+, l, r)
 Base.:/(l::AbstractDimensions, r::AbstractDimensions) = map_dimensions(-, l, r)
 
 # Multiplying ranges with units
-for (type, _, _) in ABSTRACT_QUANTITY_TYPES
-    @eval begin
-        Base.:*(r::Union{StepRange,StepRangeLen}, q::$type) = range(first(r) * q; step=Base.step(r) * q, length=length(r))
-        Base.:*(q::$type, r::Union{StepRange,StepRangeLen}) = r * q
-    end
-end
+Base.TwicePrecision{T}(x::T) where {T<:AbstractQuantity} = Base.TwicePrecision{typeof(x)}(x, zero(x))
+# TODO: Note that to get RealQuantity working, we have to overload many other functions,
+#       which is why we skip it.
 
 # Defines +, -, and mod
 for (type, true_base_type, _) in ABSTRACT_QUANTITY_TYPES, op in (:+, :-, :mod)
--- a/src/math.jl
+++ b/src/math.jl
@@ -58,8 +58,7 @@ Base.:/(l::AbstractDimensions, r::AbstractDimensions) = map_dimensions(-, l, r)
 # Multiplying ranges with units
 for (type, _, _) in ABSTRACT_QUANTITY_TYPES
     @eval begin
-        Base.:*(r::StepRange, q::$type) = (first(r) * q):(Base.step(r) * q):(last(r) * q)
-        Base.:*(r::StepRangeLen, q::$type) = range(first(r) * q; step=Base.step(r) * q, length=length(r))
+        Base.:*(r::Union{StepRange,StepRangeLen}, q::$type) = range(first(r) * q; step=Base.step(r) * q, length=length(r))
         Base.:*(q::$type, r::Union{StepRange,StepRangeLen}) = r * q
     end
 end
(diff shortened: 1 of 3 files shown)


src/math.jl
```diff
@@ -55,6 +55,15 @@ end
 Base.:*(l::AbstractDimensions, r::AbstractDimensions) = map_dimensions(+, l, r)
 Base.:/(l::AbstractDimensions, r::AbstractDimensions) = map_dimensions(-, l, r)
 
+# Multiplying ranges with units
+for (type, _, _) in ABSTRACT_QUANTITY_TYPES
+    @eval begin
+        Base.:*(r::StepRange, q::$type) = (first(r) * q):(Base.step(r) * q):(last(r) * q)
```
mcabbott, line 61: This does not seem safe from floating point weirdness. 

I would call `range(start, stop, length)` for all cases.

src/math.jl
```diff
@@ -55,6 +55,15 @@ end
...
+        Base.:*(r::StepRangeLen, q::$type) = range(first(r) * q; step=Base.step(r) * q, length=length(r))
+        Base.:*(q::$type, r::Union{StepRange,StepRangeLen}) = r * q
```
mcabbott, line 63: Notice how all ways to do this land at the same `broadcasted` method. That's probably what I'd try to overload, rather than aiming for left & right & broadcasting separately:
```
julia> (1:0.25:4) * (u"inch")
ERROR: Cannot create an additive identity from `Type{<:Quantity}`, as the dimensions are unknown. Please use `zero(::Quantity)` instead.
Stacktrace:
...
  [5] *
    @ ./twiceprecision.jl:308 [inlined]
  [6] broadcasted
    @ ./broadcast.jl:1155 [inlined]
  [7] broadcasted
    @ ./broadcast.jl:1332 [inlined]
  [8] broadcast_preserving_zero_d
    @ ./broadcast.jl:861 [inlined]
  [9] *(A::StepRangeLen{Float64, Base.TwicePrecision{…}, Base.TwicePrecision{…}, Int64}, B::Quantity{Float64, Dimensions{…}})
    @ Base ./arraymath.jl:24

julia> (1:0.25:4) .* (u"inch")
ERROR: Cannot create an additive identity from `Type{<:Quantity}`, as the dimensions are unknown. Please use `zero(::Quantity)` instead.
...
 [5] *
   @ ./twiceprecision.jl:308 [inlined]
 [6] broadcasted
   @ ./broadcast.jl:1155 [inlined]
 [7] broadcasted(::typeof(*), ::StepRangeLen{…}, ::Quantity{…})  <--- this one
   @ Base.Broadcast ./broadcast.jl:1332
 [8] top-level scope
   @ REPL[20]:1
Some type information was truncated. Use `show(err)` to see complete types.

julia> (u"inch") * (1:0.25:4)
ERROR: Cannot create an additive identity from `Type{<:Quantity}`, as the dimensions are unknown. Please use `zero(::Quantity)` instead.
...
  [6] *
    @ ./twiceprecision.jl:316 [inlined]
  [7] broadcasted
    @ ./broadcast.jl:1148 [inlined]
  [8] broadcasted
    @ ./broadcast.jl:1332 [inlined]
  [9] broadcast_preserving_zero_d
    @ ./broadcast.jl:861 [inlined]
 [10] *(A::Quantity{Float64, Dimensions{…}}, B::StepRangeLen{Float64, Base.TwicePrecision{…}, Base.TwicePrecision{…}, Int64})
```

MilesCranmer: > Notice how all ways to do this land at the same broadcasted method. That's probably what I'd try to overload, rather than aiming for left & right & broadcasting separately:

My worry is that this dependence is version dependent, and the syntax may break in the future. Wdyt?

src/math.jl
MilesCranmer: Good call!

src/math.jl
mcabbott: You should certainly test the user-facing `λ*r`, `r*λ`, `r.*λ` etc. I think this path to make a new range via broadcasting has been the same forever.

You might also be able to fix it later in the pathway. Maybe you just need to define `Base.TwicePrecision(x::Quantity)` appropriately?

BTW I suspect it should be illegal to have a StepRange of quantities, as OrdinalRange demands:

> The steps should be always-exact multiples of oneunit, and T should be a
  "discrete" type, which cannot have values smaller than oneunit.

MilesCranmer: The `broadcasted` method it's landing on is

```julia
@inline function broadcasted(f::F, arg1, arg2, args...) where {F}
    arg1′ = broadcastable(arg1)
    arg2′ = broadcastable(arg2)
    args′ = map(broadcastable, args)
    broadcasted(combine_styles(arg1′, arg2′, args′...), f, arg1′, arg2′, args′...)
end
```

I feel like it would be hard to maintain if I overload this? (Also not sure how)

Am I looking at the wrong method?

MilesCranmer: Oh yeah TwicePrecision might be much easier. Looking at it now...

MilesCranmer:
@alhirzel everything should be good now:

```julia

julia> c = collect(1u"inch":0.25u"inch":4u"inch")
13-element Vector{Quantity{Float64, Dimensions{FixedRational{Int32, 25200}}}}:
 0.025400000000000002 m
              0.03175 m
  0.03810000000000001 m
              0.04445 m
  0.05080000000000001 m
 0.057150000000000006 m
               0.0635 m
  0.06985000000000001 m
               0.0762 m
              0.08255 m
  0.08890000000000002 m
              0.09525 m
  0.10160000000000001 m
```

I also made the handling of steps very strict which I think is safer. e.g., you need to pass a step:

``julia
julia> 1u"km":2u"km"
ERROR: When creating a range over quantities, you must specify a step.
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] (::Colon)(::Quantity{Float64, Dimensions{FixedRational{…}}}, ::Quantity{Float64, Dimensions{FixedRational{…}}})
   @ DynamicQuantities ~/PermaDocuments/SymbolicRegressionMonorepo/DynamicQuantities.jl/src/math.jl:78
 [3] top-level scope
   @ REPL[3]:1
Some type information was truncated. Use `show(err)` to see complete types.
```

And the units must be compatible:

```julia
julia> 1us"m":1us"inch":3us"m"
ERROR: DimensionError: 1.0 m and 1.0 inch have incompatible dimensions
Stacktrace:
 [1] (::Colon)(start::Quantity{…}, step::Quantity{…}, stop::Quantity{…})
   @ DynamicQuantities ~/PermaDocuments/SymbolicRegressionMonorepo/DynamicQuantities.jl/src/math.jl:71
 [2] top-level scope
   @ REPL[4]:1
Some type information was truncated. Use `show(err)` to see complete types.
```

However, it is ok to use units and non-units if they are dimensionless:

```julia
julia> 1u"1":0.5:10u"1"
19-element LinRange{Quantity{Float64, Dimensions{FixedRational{Int32, 25200}}}, Int64}:
 1.0 , 1.5 , 2.0 , 2.5 , 3.0 , 3.5 , 4.0 , 4.5 , 5.0 , 5.5 , 6.0 , 6.5 , 7.0 , 7.5 , 8.0 , 8.5 , 9.0 , 9.5 , 10.0 

julia> 1:0.5:10
1.0:0.5:10.0

julia> 1:0.5:10 |> length
19
```

MilesCranmer:
Cool!

DQ tries to avoid automatic unit conversion for symbolic dimensions in the spirit of doing things explicitly/safely. Also, the symbolic dimensions should really just be used for displaying things to the user, while regular dimensions (where you can write out this range) are meant to be more efficient.

So you could do:
```julia
intervals = 1u"m":1u"cm":2u"m"

println(intervals .|> us"cm")
```
which will display everything in centimeters

MilesCranmer:
Thanks for explaining this. Yeah it is indeed tricky. Part of the reason for the existence of symbolic units is precisely so that you will get errors when trying to add two quantities which are dimensionally equivalent, but have different units. e.g., perhaps you are working with angle measured from units `m/km` (meters per kilometer), and you accidentally add it to a scalar value (radians are dimensionless according to SI rules). Though that is dimensionally correct, it could be semantically _incorrect_! It is really context dependent when you want this to be automatic. And I feel like the cost of a slight reduction in usability is worth this amount of added safety. (Though I understand that the exact weighting is highly subjective!)

So SymbolicDimensions requires you to be explicit about such conversions to avoid such mistakes, though you can always use Dimensions and then convert back later. (If I recall correctly, some people working on JuMP.jl requested this feature so they could warn users about unit mismatch.)

**However!** I would be open to having a second type of SymbolicDimensions that _does_ automatically convert, if you would be interested in contributing this. The relevant file is here: https://github.com/SymbolicML/DynamicQuantities.jl/blob/main/src/symbolic_dimensions.jl. I think you would basically need to create a new abstract type:

```julia
abstract type AbstractCompositeSymbolicDimensions{R} <: AbstractSymbolicDimensions{R} end
```

and rewrite `SymbolicDimensions` to inherit from that instead of `AbstractSymbolicDimensions`. Also rewrite nearly every method that takes `SymbolicDimensions` as input to instead take `AbstractCompositeSymbolicDimensions`.

Then create a new symbolic dimensions type:

```julia
struct AutoSymbolicDimensions{R} <: AbstractCompositeSymbolicDimensions{R}
    # same as SymbolicDimensions
end
```

So it's like

```julia
AbstractDimensions{R}
│
└── AbstractSymbolicDimensions{R}
    │
    ├── SymbolicDimensionsSingleton{R}
    │
    └── AbstractCompositeSymbolicDimensions{R}
        │
        ├── SymbolicDimensions{R}
        │
        └── AutoSymbolicDimensions{R}
```

and you could define a trait, like:

```julia
is_autoconverting(::Type{<:SymbolicDimensions}) = false
is_autoconverting(::Type{<:AutoSymbolicDimensions}) = true
```

Then check that trait in the symbolic dimensions operators. If it's true, then you can have a unit promotion step for the auto ones. I would just have it convert to Dimensions (with `uexpand`) and then symbolic-ify it again.

We would also need a macro. Maybe just `ua"km"` for automatic symbolic units?

MilesCranmer:
For these things you would just define new promotion rules here: https://github.com/SymbolicML/DynamicQuantities.jl/blob/9625403f76385603a3acd0fa1119fada3912a5ba/src/symbolic_dimensions.jl#L542-L556. between AutoSymbolicDimensions and everything else. Then everything should just work automatically. (which is why the following operation already works:

```julia
julia> 1u"m" == 1us"m"
true
```

I think the order of precedence would be:

Dimensions > AutoSymbolicDimensions > SymbolicDimensions

Meaning that

- Dimensions + AutoSymbolicDimensions => Dimensions
- SymbolicDimensions + AutoSymbolicDimensions => AutoSymbolicDimensions

At least, I think this makes the most sense.

PR #159, opened 2025-01-31
Affine Units Functionality

Deduction42:
This adds a new type of Dimension (AffineDimension) that can be used to describe units with offsets (like Celsius and Fahrenheit). Like most other packages, operations on AffineDimensions are forbidden except subtraction. This is done by throwing an error on convert if the offset is zero; operations are allowed if the offset is zero (by using promote/convert to SI units).


MilesCranmer:
Thanks! Can you walk me through the implementation a bit? How does it work, what are the fields used for, etc?

MilesCranmer:
And FYI I would like to get things to 100% unit-test coverage if possible (particularly important as I'm unfamiliar with the code). My expectation is that there will be many unexpected errors that will pop out when going through this process.

MilesCranmer:
I think the Aqua and JET tests will likely take a bit of work to get passing but are also important

Deduction42:
Sure! As for how it works. Almost all of the changes are in "affine_dimensions.jl", which somewhat mirrors the patterns found in "symbolic_dimensions.jl" (or as much as reasonably possible). The basic structure is the AffineDimension <: AbstractAffineDimension <: AbstractDimension which has the following fields:
```
@kwdef struct AffineDimensions{R} <: AbstractAffineDimensions{R}
    scale::Float64 = 1.0
    offset::Float64 = 0.0
    basedim::Dimensions{R}
    symbol::Symbol = :nothing
end
```
The `scale` and `offset` parameters are used to convert this unit to its SI equivalent. So the SI quantitiy is basically 
```
si_quantity = affine_quantity*scale + offset
```
This is essentially what uexpand(affine_quantity) does when it converts things to SI dimensions. The `symbol` field is optional and is for displaying units. If no symbol is present, it will merely display (scalse+offset basedim). For example:
```
kelvin  = AffineDimensions(scale=1.0, offset=0.0, basedim=u"K")
rankine = AffineDimensions(scale=5/9, offset=0.0, basedim=kelvin)
fahrenheit = AffineDimensions(scale=1.0, offset=459.67, basedim=rankine)
>> (0.5555555555555556+255.37222222222223 K)
```
But if we used 
```
fahrenheit = AffineDimensions(scale=1.0, offset=459.67, basedim=rankine, symbol=:°F)
>>°F
```

Deduction42:
Generally speaking, it's not encouraged to do mathematical operations in affine units because of potential ambiguities present in the offset. Promotion rules will attempt to convert affine quantities into basic dimensional (SI) quantities. However, the function "convert" will fail if the offset is not zero (I allow typical multiplication/division operations for affine units with zero offsets in order to make affine unit macros work with compound units like m/s). 

The affine unit marcro is "ua" and I registered the Celsius and Fahrenheit units. So you could write
```
 uconvert(ua"°C", 0ua"°F")
>> -17.77777777777777 °C
```
Becuase AffineDimensions are more general than Dimensions and SymbolicDimensions, registering a unit will also make it available to the "ua" macro, and I registered all items in UNIT_SYMBOLS and UNIT_VALUES off the bat, so you could write:
```
velocity = ua"mm/s"
>>1.0 (0.001 m s⁻¹)
```
This generality is so that you could make a single Type-Stable container (like a dictionary or vector) that represents both Affine and Non-Affine units and easily convert them all to SI or Symbolic units. Note however, compound non-affine units break, becuase offsets cannot be multiplied:
```
ua"°C/K"
ERROR: AssertionError: AffineDimensions °C has a non-zero offset, implicit conversion is not allowed due to ambiguity. Use uexpand(x) to explicitly convert
Stacktrace:
```

Deduction42:
Finally, there is a special macro for registering AffineDimensions only, because the regular registration macro doesn't handle offsets. Registration will apply to anything that can be evaluated as a dimension or quantity. For example, you could do:

```
@register_unit psi 6.89476us"kPa" #Regular registration method for PSI
@register_affine_unit psig AffineDimensions(offset=u"Constants.atm", basedim=u"psi") #Gauge-pressure subtracts atmospheric

uconvert(ua"psig", u"Constants.atm")
>> 0.0 psig
```

MilesCranmer:
Quick question - why do affine dimensions need to support algebraic operations on non-offset affine dimensions? If it's non-offset, why not just force the user to immediately convert to SI units if they want to use such operations?

Deduction42:
Anyway, I'll start writing the tests for this on Monday, I just wanted to make sure you were onboard with adding this before I put the effort into it. I was pleasantly surprised that this took me less than a week to do; your package has some fairly elegant ways to plug in this feature (although these plug-in points are a bit hard to find). To reiterate, almost all of the changes are in the new "affine_dimensions.jl" file I added. I also had to make some changes to register_units.jl to add automatic registering of any units to the AffineUnits registry, and make a separate macro for registering AffineDimensions only.

I do have some questions:
1. In the "test coverage" tools, is there any way to identify what lines are missing from the coverage? Or is this a game of educated whack-a-mole?
2. I noticed in "register_units.jl" that you use a SpinLock(), is there any reason you're using this rather than the more common (and more recommended) ReentrantLock()?
3. Last time I contributed, I was able to run all the tests by just running "run_tests.jl", but this time, there's a bunch of test-related dependencies. Is there a standard way of loading this environment before doing the tests, or do I have to just install all the test dependencies myself?

Deduction42:
> Quick question - why do affine dimensions need to support algebraic operations on non-offset affine dimensions? If it's non-offset, why not just force the user to immediately convert to SI units if they want to use such operations?

The only reason I do this is so that I can parse units like "ua"m/s"" as an Affine Unit.

MilesCranmer:
This is the command I usually run at the root of this repository:

```bash
julia --startup-file=no --depwarn=yes --threads=auto --code-coverage=user --project=. -e 'using Pkg; Pkg.test(coverage=true)'
julia --startup-file=no --depwarn=yes --threads=auto coverage.jl
```

Then I use coverage gutters in VSCode https://marketplace.visualstudio.com/items?itemName=ryanluker.vscode-coverage-gutters which highlights the missing lines inside the IDE.

MilesCranmer:
> * I noticed in "register_units.jl" that you use a SpinLock(), is there any reason you're using this rather than the more common (and more recommended) ReentrantLock()?

ReentrantLock lets the same thread acquire a lock multiple times and I didn't see any use of that in this context. Also I expect the SpinLock to basically never get hit, it's just for the rare case where someone actually tries to register two units simultaneously, which I would expect to be exceedingly rare.

Deduction42:
Excellent, thanks!

Deduction42:
Is there any trick to making "coverage gutters in VSCode" work consistently? When I first installed it, it said that none of the files were covered even though there were ".cov" files that I could read myself. I added "*.cov" to the list of `"coverage-gutters.coverageFileNames":`, which worked for a while, but it only showed the files that existed and won't update if I run new tests.

lcov.info
```diff
@@ -0,0 +1,1426 @@
+SF:src\DynamicQuantities.jl
```
MilesCranmer, line 1: (needs to be removed)

test/affine_tests.jl
```diff
@@ -0,0 +1,39 @@
+using Revise
```
MilesCranmer, line 1: please put the tests in `test/unittests.jl` rather than a new file

MilesCranmer:
Can there be aliases for other common ways to write temperature, like degC?

Also what other affine units are out there that are common? I think it would be useful to have a couple others, just to ensure the structure itself captures all cases other than just temperature.

Deduction42:
> Can there be aliases for other common ways to write temperature, like degC?

Yes, that's totally possible. I was simply following the Unitful.jl convention for the label. Did you want me to add "degC" and "degF" to the default registry?

> Also what other affine units are out there that are common? I think it would be useful to have a couple others, just to ensure the structure itself captures all cases other than just temperature.

Honestly, it's almost only temperature, because we've been reading temperatures far longer than we've known about "absolute zero". Temperature units are the only default affine units in [Unitful.jl](https://github.com/PainterQubits/Unitful.jl/blob/d869c3614969186ae21f9029216eadaf99e3e5c8/src/pkgdefaults.jl#L272)

In some cases we use "gauge pressure" which is pressure relative to atmosphere; this is sometimes denoted using a "g" at the end as "kPag" or "psig" but the notations for those aren't exactly standard. I've seen a lot of variation in how these units are represented (it's almost as bad as the hideous "standard cubic feet" that I often run into in the oil and gas sector). Because of this variation, I might prefer leave it up to the user to register it (this very example has been given for "psig").

Deduction42:
Hello @MilesCranmer, just checking in to see if there are any other blockers for merging. I just added a small update to the pull request to ignore symbols when checking for affine dimension equality. 

Some of my projects that use this package are starting to get some traction in my organization, and I'd really like to be able to simplify the workflow with this PR before bringing more people onboard with these projects. DynamicQuantities has the simplest and most expressive string parsing and unit registry tools of all the packages I've seen; I'm currently handling affine units with the Unitful.jl extension, and the kludges I had to use to parse affine unit strings in CSV config files are an unsightly, type-unstable abomination.

Deduction42:
The biggest simplification would be to allow `uexpand` to work on Quantities with `dimension` (where it just returns the value). A lot of the conversion code funnels `SymbolicDimensions |> Dimensions |> AffineDimensions` so having a universal simple function to convert any dimension into `Dimensions` would be nice. I currently defined a function `siunits` that's basically just this, a `uexpand` that works on regular dimensions.

I really tried to simplify this to the maximum extent possible. My main goals of this PR were: 
1. Support creating AffineDimensions from regular Dimensions
2. Support registering/parsing of AffineUnits
3. Support unit conversion from AffineUnits
4. Make sure that when people try to do operations on AffineUnits, they get useful errors telling them not to

Moreover, I wanted to do this through maximizing code reuse, minimizing new code, and following the package's existing conventions where possible. Unfortunately, most of the code's complexity is a direct consequence of the design decisions of the rest of the package.

Lines 15-63 deal with the fact that the macro `u"kg/s"` gives a quantity, not a dimension. Thus, if I'm building an AffineDimension, I have to support using quantities as a dimensional input.

Lines 90-106 deal with the fact that `uexpand` doesn't work on `Quantities` with `Dimensions`

Lines 132-158 are basically pulled from `symbolic_dimensions.jl` and deal with the fact that there are numerous `Quantity` types. If you can think of a way to simplify this, I'd be open to hear it.

Lines 160-175 is basically telling the promotion rules to just convert any AffineDimensions to Dimensions. If they have a non-zero offset, it will throw an error explicitly telling people to convert to non-affine units. It's a way of catching cases where people are trying to perform mathematical manipulation of affine quantities.

Lines 180-216 are the _raison d'etre_ of this pull request, allowing you to convert to and from AffineDimensions.

Lines 218-265 deal with the fact that `map_to_dimensions` doesn't account for scale, outside actual quantities. Supporting these operations is necessary for parsing affine units in a type-stable way.

Lines 267-282 basically throw errors when adding two numbers with the same affine dimensions. This is necessary to prevent silently returning potentially wrong results.

Lines 293-436 is a mirror image of the SymbolicDimensions unit registry and units parsing. Unfortunately, a separate registry is required because AffineDimensions are more general than SymbolicDimensions. This means a new macro is required for registering units, but this macro doesn't affect the other registries.

MilesCranmer:
Thanks for talking me through this!

For many of these that are replicas of the original code, I would definitely be open to refactoring both the replica and the new copy to be instances of a more general code pattern, so long as it's still fairly clean. Like, for example:

> Lines 293-436 is a mirror image of the SymbolicDimensions unit registry and units parsing. Unfortunately, a separate registry is required because AffineDimensions are more general than SymbolicDimensions. This means a new macro is required for registering units, but this macro doesn't affect the other registries.

Here, maybe there's a way to refactor both so they can be an instance of a single registry somehow? i.e., I would like to avoid needing to make ["shotgun surgery"](https://refactoring.guru/smells/shotgun-surgery) in the future (want to guard my overall maintenance load). Especially since this feature by itself is something I would very rarely if ever use, I want to make sure that I can still [dogfood](https://en.wikipedia.org/wiki/Eating_your_own_dog_food) the relevant code. If the code is shared with functionality I do use, it's a lot easier for me to hit it from my normal use in other contexts.

Deduction42:
I think the biggest opportunity to reduce "shotgun surgery" by refactoring is in lines 398-423.

`map_to_scope` and `lookup_units` are the biggest culprits. They make use of different module-level constants. If I was able pass a "registry" argument like this
```
map_to_scope(ex::Expr, registry::Dict{Symbol, <:Quantity})
```
I could import this function independent of any module. I'm also wondering why you're not looking things up in a dictionary directly. Dictionaries are slow when looking things up in small arrays, but the registry for Symbolic is likely big enough to make dictionaries more efficient. Moreover, you're doing the lookup twice. For example in symbolic_dimensions.jl Lines 496-505, you have
```
    function map_to_scope(sym::Symbol)
        if sym in UNIT_SYMBOLS
            # return at end
        elseif sym in CONSTANT_SYMBOLS
            throw(ArgumentError("Symbol $sym found in `Constants` but not `Units`. Please use `us\"Constants.$sym\"` instead."))
        else
            throw(ArgumentError("Symbol $sym not found in `Units` or `Constants`."))
        end
        return lookup_unit(sym)
    end
```
The first branch iterates through all of UNIT_SYMBOLS and then lookup_unit is then called again. I'm not sure why all the fuss is done to find the element in UNIT_SYMBOLS rather than just looking things up in the dictionary like this:
```
    function map_to_scope(sym::Symbol, unitregistry::Dict{Symbol, <:Quantity}, constregistry::Dict{Symbol, <:Quantity})
        q = get(unitregistry, sym) do
            if haskey(constregistry, q)
                throw(ArgumentError("Symbol $sym found in `Constants` but not `Units`. Please use `us\"Constants.$sym\"` instead."))
            else
                throw(ArgumentError("Symbol $sym not found in `Units` or `Constants`."))
            end
        end
        return q
    end
```
Now AffineDimensions only has a unit registry, no constants, so it would look more like
```
    function map_to_scope(sym::Symbol, unitregistry::Dict{Symbol, <:Quantity})
        q = get(unitregistry, sym) do
                throw(ArgumentError("Symbol $sym not found in `AffineUnits` registry."))
        end
        return q
    end
```
Either way, this would eliminate the need for having two constants "UNIT_SYMBOLS" and "UNIT_VALUES" as well as the need for multiple lookups.

Deduction42:
The only thing is that the above refactoring would require quite a bit of effort. Shotgun surgery could go both ways as well. Because this parsing utility is independent of SymbolicDimensions (which is arguably more complex), then updating that codebase is less likely to break something here. In fact, you probably could refactor your part of the code that way and then I could follow suit later if you like.

I really don't mind maintaining this. I already contributed to this repo half a year ago to improve performance of Unitful.jl integration. I'm also going to be putting this functionality through the ringer on some really big projects over the next couple of months, so if anything goes wrong, I'll likely be the first to hear about it and fix it.

MilesCranmer:
Thanks for the reassurance, that is good to hear and makes me much more positive about the increased code from this! Ok, I will try to take a look later this week. This week in particular is a bit insane for me; are you ok with waiting until this weekend?

Deduction42:
How is your schedule looking this week? I'm planning on building out the new workflow this week for my industrial AI app, which should be a pretty rigorous training ground for this extension.

src/affine_dimensions.jl
```diff
@@ -0,0 +1,466 @@
+
+const AbstractQuantityOrArray{T,D} = Union{UnionAbstractQuantity{T,D}, QuantityArray{T,<:Any,D}}
+
+abstract type AbstractAffineDimensions{R} <: AbstractDimensions{R} end
+
+const AffineOrSymbolicDimensions{R} = Union{AbstractAffineDimensions{R}, AbstractSymbolicDimensions{R}}
+
+@kwdef struct AffineDimensions{R} <: AbstractAffineDimensions{R}
```
MilesCranmer, line 8: Could you add a docstring for this?

src/affine_dimensions.jl
```diff
@@ -0,0 +1,466 @@
...
+    scale::Float64 = 1.0
+    offset::Float64 = 0.0
+    basedim::Dimensions{R}
+    symbol::Symbol = :nothing
+end
+
+#Inferring the type parameter R ========================================================================================================================
```
MilesCranmer, line 15: Could you remove these long `===` lines

src/affine_dimensions.jl
```diff
@@ -0,0 +1,466 @@
...
+
+#Base case when everyting is convrted to si units (offset is assumed to be in SI units)
+function AffineDimensions{R}(s::Real, o::UnionAbstractQuantity{<:Any,<:Dimensions}, q::UnionAbstractQuantity{<:Any,<:Dimensions}, sym::Symbol=:nothing) where R
+    dimension(o) == dimension(q) || throw(DimensionError(o, q))
+    o_val = ustrip(o)
+    q_val = ustrip(q)
+    return AffineDimensions{R}(s*q_val, o_val, dimension(q), sym)
+end
+
+#If a quantity is used only for the dimension, the offset is assumed to be in the same scale as the quantity
```
MilesCranmer, line 60: Please format comments with a space immediately following the `#` character

src/affine_dimensions.jl
```diff
@@ -0,0 +1,466 @@
...
+siunits(q::UnionAbstractQuantity{<:Any, <:Dimensions}) = q
+siunits(q::UnionAbstractQuantity{<:Any, <:AbstractSymbolicDimensions}) = uexpand(q)
+function siunits(q::Q) where {T, R, D<:AbstractAffineDimensions{R}, Q<:UnionAbstractQuantity{T,D}}
+    return force_convert(with_type_parameters(Q, T, Dimensions{R}), q)
+end
+siunits(q::QuantityArray) = siunits.(q)
+
+
+"""
+uexpand(q::Q) where {T, R, D<:AbstractAffineDimensions{R}, Q<:UnionAbstractQuantity{T,D}}
```
MilesCranmer, line 100: Please format docstrings with 4 leading spaces preceding the signature, similar to the other docstrings

src/affine_dimensions.jl
```diff
@@ -0,0 +1,466 @@
...
+        end
+
+        #Forced conversion of (AbstractQuantity){T,R<:AffineDimensions} to (AbstractQuantity){T,R<:Dimensions} (zero offset requirement overridden)
+        function force_convert(::Type{Q}, q::UnionAbstractQuantity{<:Any,<:AbstractAffineDimensions}) where {T,D<:Dimensions,Q<:$type{T,D}}
+            d = dimension(q)
+            v = ustrip(q)*scale(d) + offset(d)
+            return constructorof(Q)(convert(T, v), basedim(d))
+        end
+
+        #Conversion of (AbstractQuantity){T,R<:AffineDimensions} to (AbstractQuantity){T,R<:Dimensions}
```
MilesCranmer, line 152: Comments like this are not needed since it is obvious from the code

src/affine_dimensions.jl
```diff
@@ -0,0 +1,466 @@
...
+    uconvert(qout::UnionAbstractQuantity{<:Any, <:AbstractAffineDimensions}, q::UnionAbstractQuantity{<:Any, <:Dimensions})
+
+Convert a quantity `q` with base SI units to the affine units of `qout`, for `q` and `qout` with compatible units.
+You can also use `|>` as a shorthand for `uconvert`
+"""
+function uconvert(qout::UnionAbstractQuantity{<:Any, <:AffineDimensions}, q::UnionAbstractQuantity{<:Any, <:Dimensions})
+    @assert isone(ustrip(qout)) "You passed a quantity with a non-unit value to uconvert."
+    dout = dimension(qout)
+    dimension(q) == basedim(dout) || throw(DimensionError(q, qout))
+    vout = (ustrip(q)-offset(dout))/scale(dout)
```
MilesCranmer, line 191: Please use a space for infix operators, like `ustrip(q) - offset(dout)`

README.md
```diff
@@ -284,6 +283,55 @@ julia> 3us"V" |> us"OneFiveV"
 2.0 OneFiveV
 ```
 
+### Affine units
```
MilesCranmer, line 286: Can this be removed from the README and instead put in the docs file (`affine_units.md`)? Or otherwise combined?

src/DynamicQuantities.jl
```diff
@@ -32,6 +33,7 @@ using DispatchDoctor: @stable
     include("complex.jl")
     include("register_units.jl")
     include("disambiguities.jl")
+    include("affine_dimensions.jl")
```
MilesCranmer, line 36: I think this should go after `symbolic_dimensions.jl`

src/affine_dimensions.jl
```diff
@@ -0,0 +1,466 @@
...
+scale(d::AffineDimensions)  = d.scale
+offset(d::AffineDimensions) = d.offset
+basedim(d::AffineDimensions) = d.basedim
+
+with_type_parameters(::Type{<:AffineDimensions}, ::Type{R}) where {R} = AffineDimensions{R}
+constructorof(::Type{AffineDimensions}) = AffineDimensions{DEFAULT_DIM_BASE_TYPE}
+constructorof(::Type{AffineDimensions{R}}) where R = AffineDimensions{R}
+
+function Base.show(io::IO, d::AbstractAffineDimensions)
+    addsign = ifelse(offset(d)<0, "-" , "+")
```
MilesCranmer, line 75: Why is this needed? I see `abs(offset(d))` is printed already. Why not just print `offset(d)` directly?

src/affine_dimensions.jl
```diff
@@ -0,0 +1,466 @@
...
+    end
+end
+
+Base.:(==)(q1::UnionAbstractQuantity{<:Any, <:AffineDimensions}, q2::UnionAbstractQuantity{<:Any, <:AffineDimensions})   = (siunits(q1) == siunits(q2))
+Base.:(==)(q1::UnionAbstractQuantity{<:Any, <:AffineDimensions}, q2::UnionAbstractQuantity{<:Any, <:AbstractDimensions}) = (siunits(q1) == siunits(q2))
+Base.:(==)(q1::UnionAbstractQuantity{<:Any, <:AbstractDimensions}, q2::UnionAbstractQuantity{<:Any, <:AffineDimensions}) = (siunits(q1) == siunits(q2))
+Base.:(≈)(q1::UnionAbstractQuantity{<:Any, <:AffineDimensions}, q2::UnionAbstractQuantity{<:Any, <:AffineDimensions})    = (siunits(q1) ≈ siunits(q2))
+Base.:(≈)(q1::UnionAbstractQuantity{<:Any, <:AffineDimensions}, q2::UnionAbstractQuantity{<:Any, <:AbstractDimensions})  = (siunits(q1) ≈ siunits(q2))
+Base.:(≈)(q1::UnionAbstractQuantity{<:Any, <:AbstractDimensions}, q2::UnionAbstractQuantity{<:Any, <:AffineDimensions})  = (siunits(q1) ≈ siunits(q2))
+Base.:(==)(d1::AffineDimensions, d2::AffineDimensions) = (d1.scale==d2.scale) & (d1.offset==d2.offset) & (d1.basedim == d2.basedim)
```
MilesCranmer, line 290: Should use `&&` not `&`

src/affine_dimensions.jl
```diff
@@ -0,0 +1,466 @@
...
+        return _no_offset_expand(q1) - _no_offset_expand(q2)
+    end
+end
+
+Base.:(==)(q1::UnionAbstractQuantity{<:Any, <:AffineDimensions}, q2::UnionAbstractQuantity{<:Any, <:AffineDimensions})   = (siunits(q1) == siunits(q2))
+Base.:(==)(q1::UnionAbstractQuantity{<:Any, <:AffineDimensions}, q2::UnionAbstractQuantity{<:Any, <:AbstractDimensions}) = (siunits(q1) == siunits(q2))
+Base.:(==)(q1::UnionAbstractQuantity{<:Any, <:AbstractDimensions}, q2::UnionAbstractQuantity{<:Any, <:AffineDimensions}) = (siunits(q1) == siunits(q2))
+Base.:(≈)(q1::UnionAbstractQuantity{<:Any, <:AffineDimensions}, q2::UnionAbstractQuantity{<:Any, <:AffineDimensions})    = (siunits(q1) ≈ siunits(q2))
+Base.:(≈)(q1::UnionAbstractQuantity{<:Any, <:AffineDimensions}, q2::UnionAbstractQuantity{<:Any, <:AbstractDimensions})  = (siunits(q1) ≈ siunits(q2))
+Base.:(≈)(q1::UnionAbstractQuantity{<:Any, <:AbstractDimensions}, q2::UnionAbstractQuantity{<:Any, <:AffineDimensions})  = (siunits(q1) ≈ siunits(q2))
```
MilesCranmer, line 289: Can these be wrapped up into a for loop containing an `@eval`?

src/affine_dimensions.jl
```diff
@@ -0,0 +1,466 @@
...
+    ex = :($AffineUnits.as_quantity($ex))
+    return esc(ex)
+end
+
+
+
+
+
+
+
```
MilesCranmer, line 466: Extra newlines at end (should only be 1)

src/affine_dimensions.jl
```diff
@@ -0,0 +1,466 @@
...
+            return ex
+        end
+    end
+
+    function map_to_scope(sym::Symbol)
+        if sym in AFFINE_UNIT_SYMBOLS
+            return lookup_unit(sym)
+        else
+            throw(ArgumentError("Symbol $sym not found in `AffineUnits`."))
+        end
```
MilesCranmer, line 413:
```suggestion
        sym in AFFINE_UNIT_SYMBOLS || throw(ArgumentError("Symbol $sym not found in `AffineUnits`."))
        return lookup_unit(sym)
```

MilesCranmer (changes requested):

src/affine_dimensions.jl
```diff
@@ -0,0 +1,466 @@
...
+    import ..WriteOnceReadMany
+    import ..SymbolicUnits.as_quantity
+
+    #Constants are not imported
+    const AFFINE_UNIT_SYMBOLS = WriteOnceReadMany([UNIT_SYMBOLS...])
+    const AFFINE_UNIT_VALUES  = WriteOnceReadMany(affine_unit.([UNIT_VALUES...]))
+    const AFFINE_UNIT_MAPPING = WriteOnceReadMany(Dict(s => INDEX_TYPE(i) for (i, s) in enumerate(AFFINE_UNIT_SYMBOLS)))
+
+    # Used for registering units in current module
+    function update_external_affine_unit(name::Symbol, q::UnionAbstractQuantity{<:Any,<:AffineDimensions{R}}) where R
```
MilesCranmer, line 328:
```suggestion
    function update_external_affine_unit(name::Symbol, q::UnionAbstractQuantity{<:Any,<:AffineDimensions{R}}) where {R}
```

src/affine_dimensions.jl
```diff
@@ -0,0 +1,466 @@
...
+    assert_no_offset(l)
+    return AffineDimensions(
+        scale   = inv(scale(l)),
+        offset  = offset(l),
+        basedim = inv(basedim(l))
+    )
+end
+
+# Operations on self-values ======================================================================================
+function _no_offset_expand(q::Q) where {T, R, D<:AbstractAffineDimensions{R}, Q<:UnionAbstractQuantity{T,D}}
```
MilesCranmer, line 268: Current style is no spaces in type expression, so this should be
```suggestion
function _no_offset_expand(q::Q) where {T,R,D<:AbstractAffineDimensions{R},Q<:UnionAbstractQuantity{T,D}}
```

MilesCranmer:
Also, how should I interpret the printout? I see this:

```julia
julia> 1ua"km/s"
1.0 (1000.0 m s⁻¹)

julia> 2ua"km/s"
2.0 (1000.0 m s⁻¹)
```

Wasn't sure what this meant (Maybe this is a standard format?)

src/affine_dimensions.jl
```diff
@@ -0,0 +1,466 @@
...
+    import ..AffineDimensions
+    import ..UnionAbstractQuantity
+
+    import ..DEFAULT_DIM_BASE_TYPE
+    import ..WriteOnceReadMany
+    import ..SymbolicUnits.as_quantity
+
+    #Constants are not imported
+    const AFFINE_UNIT_SYMBOLS = WriteOnceReadMany([UNIT_SYMBOLS...])
+    const AFFINE_UNIT_VALUES  = WriteOnceReadMany(affine_unit.([UNIT_VALUES...]))
```
MilesCranmer, line 324:
```suggestion
    const AFFINE_UNIT_VALUES  = WriteOnceReadMany(map(affine_unit, UNIT_VALUES))
```

src/affine_dimensions.jl
```diff
@@ -0,0 +1,466 @@
...
+    import ..AbstractDimensions
+    import ..AffineDimensions
+    import ..UnionAbstractQuantity
+
+    import ..DEFAULT_DIM_BASE_TYPE
+    import ..WriteOnceReadMany
+    import ..SymbolicUnits.as_quantity
+
+    #Constants are not imported
+    const AFFINE_UNIT_SYMBOLS = WriteOnceReadMany([UNIT_SYMBOLS...])
```
MilesCranmer, line 323:
```suggestion
    const AFFINE_UNIT_SYMBOLS = WriteOnceReadMany(deepcopy(UNIT_SYMBOLS))
```

src/affine_dimensions.jl
```diff
@@ -0,0 +1,466 @@
...
+function Base.:^(l::AffineDimensions{R}, r::Integer) where {R}
+    assert_no_offset(l)
+    return AffineDimensions(
+        scale   = scale(l)^r,
+        offset  = offset(l),
+        basedim = basedim(l)^tryrationalize(R, r)
+    )
+end
+
+function Base.:inv(l::AffineDimensions{R}) where {R}
```
MilesCranmer, line 258:
```suggestion
function Base.inv(l::AffineDimensions{R}) where {R}
```

MilesCranmer:
I think since the `AbstractAffineDimensions` isn't used much, I would just delete it and let `AffineDimensions <: AbstractDimensions` directly.

Also, why is there an `AffineOrSymbolicDimensions`?

src/affine_dimensions.jl
```diff
@@ -0,0 +1,466 @@
...
+# Multiplication and division of AffineDimensions ===============================================================
+function Base.:*(l::AffineDimensions, r::AffineDimensions) 
+    assert_no_offset(l)
+    assert_no_offset(r)
+    return AffineDimensions(
+        scale  = scale(l)*scale(r),
+        offset = offset(l),
+        basedim = basedim(l)*basedim(r)
+    )
+end
```
MilesCranmer, line 227: Can these be automatically defined by overloading `map_dimensions`? This is what is done for symbolic dimensions

src/affine_dimensions.jl
```diff
@@ -0,0 +1,466 @@
...
+        end
+
+        #Conversion of (AbstractQuantity){T,Dimensions{R}} to (AbstractQuantity){T,AffineDimensions{R}}
+        function Base.convert(::Type{Q}, q::UnionAbstractQuantity{<:Any,<:Dimensions}) where {T,R,Q<:$type{T,AffineDimensions{R}}}
+            dims = AffineDimensions{R}(scale=1, offset=0, basedim=dimension(q))
+            return constructorof(Q)(convert(T, ustrip(q)), dims)
+        end
+
+        #Forced conversion of (AbstractQuantity){T,R<:AffineDimensions} to (AbstractQuantity){T,R<:Dimensions} (zero offset requirement overridden)
+        function force_convert(::Type{Q}, q::UnionAbstractQuantity{<:Any,<:AbstractAffineDimensions}) where {T,D<:Dimensions,Q<:$type{T,D}}
```
MilesCranmer, line 146: Would prefer `_unsafe_convert`

Deduction42:
> Also, how should I interpret the printout? I see this:
> 
> ```julia
> julia> 1ua"km/s"
> 1.0 (1000.0 m s⁻¹)
> 
> julia> 2ua"km/s"
> 2.0 (1000.0 m s⁻¹)
> ```
> 
> Wasn't sure what this meant (Maybe this is a standard format?)
> 
> Also
> 
> ```julia
> julia> 2ua"°F"
> 2.0 °F
> 
> julia> 2ua"degF"
> 2.0 degF
> ```
> 
> I think these two things should have the same printout, no?

The default symbol of AffineDimensions is :nothing (often occurs from multiplication), if this is the case, then the standard format is printed out (basically, SI dimensions with a scale and offset). For registered units, if there is a symbol, it will print the symbol instead. For some reason, the symbols aren't assigned for these. I'll have a look.

Deduction42:
Hello @MilesCranmer! I think I managed to cover all of your asks. I even found a way to make map_dimensions more efficient and accurate, so now I'm using that for operations. You can also check for better printouts of existing units in the ua"..." macro. The only thing is that you will still see the default behavior of
```
julia> 1ua"km/s"
1.0 (1000.0 m s⁻¹)
```
Because compound units require multiplication/division during parsing and such operations lose the symbolic information of the original units; they eagerly evaluate to SI units with scale and offset. The only reason I support these operations is that it's necessary for parsing.

MilesCranmer:
Can you mark individual comments as “resolved” (see button at bottom), then I can see which ones are left, if any

Deduction42:
I resolved all the comments except one. Did you need me to add all affine units as constants inside the AffineUnits module?

MilesCranmer:
Ah, I just wasn't sure what that `begin ... end` block was for. If those are temporary variables (?) maybe use a `let ... end` so that the variables are scoped?

Deduction42:
> Ah, I just wasn't sure what that `begin ... end` block was for. If those are temporary variables (?) maybe use a `let ... end` so that the variables are scoped?

Ah shoot. That's what I meant to use... a "let" block. I keep getting the two confused. Just a minute, I'll change that and I'll be good to go.

Deduction42:
Okay, I updated it, I'm good to go.

MilesCranmer:
Yes, sorry, I just haven't yet found a block of time long enough to give this a full review. Will hopefully get to it soon...

Deduction42:
Okay, no worries. In fact, what I planned to do got pushed back for other reasons. Moreover, I'm going to be offline until March 6th, so it may be best to avoid merging until hearing back from me after then.

Deduction42:
Okay, I'm back online, feel free to review at your earliest convenience.

src/affine_dimensions.jl
```diff
@@ -0,0 +1,452 @@
...
+    end
+end
+
+assert_no_offset(d::AffineDimensions) = iszero(offset(d)) || throw(AssertionError("AffineDimensions $(d) has a non-zero offset, implicit conversion is not allowed due to ambiguity. Use uexpand(x) to explicitly convert"))
+siunits(q::UnionAbstractQuantity{<:Any,<:Dimensions}) = q
+siunits(q::UnionAbstractQuantity{<:Any,<:AbstractSymbolicDimensions}) = uexpand(q)
+function siunits(q::Q) where {T,R,D<:AbstractAffineDimensions{R},Q<:UnionAbstractQuantity{T,D}}
+    return _explicit_convert(with_type_parameters(Q, T, Dimensions{R}), q)
+end
+siunits(q::QuantityArray) = siunits.(q)
```
MilesCranmer, line 102: Sorry if I already asked this, but why not just use `uexpand` directly? It seems like it's identical in purpose, no?

src/affine_dimensions.jl
```diff
@@ -0,0 +1,452 @@
...
+
+#Generic fallback (slower and less accurate but works for more cases)
+function map_dimensions(op::F, args::AffineDimensions...) where {F<:Function}
+    assert_no_offset.(args)
+    return AffineDimensions(
+        scale=exp(op(log.(scale.(args))...)),
+        offset=zero(Float64),
+        basedim=map_dimensions(op, basedim.(args)...)
+    )
+end
```
MilesCranmer, line 261: Sorry; what is this for?

src/affine_dimensions.jl
```diff
@@ -0,0 +1,452 @@
...
+
+# If a quantity is used only for the dimension, the offset is assumed to be in the same scale as the quantity 
+function AffineDimensions{R}(s::Real, o::Real, q::Q, sym::Symbol=:nothing) where {R, Q<:UnionAbstractQuantity}
+    return AffineDimensions{R}(s, o*q, q, sym)
+end
+
+
+scale(d::AffineDimensions) = d.scale
+offset(d::AffineDimensions) = d.offset
+basedim(d::AffineDimensions) = d.basedim
```
MilesCranmer, line 74: I think it would be better to name these `affine_scale`, `affine_offset`, and `affine_base_dim`

src/affine_dimensions.jl
```diff
@@ -0,0 +1,452 @@
...
+    q_si  = siunits(q)
+    dims  = AffineDimensions{R}(scale=1.0, offset=0.0, basedim=dimension(q_si))
+    q_val = convert(T, ustrip(q_si))
+    return constructorof(Q)(q_val, dims)
+end
+
+"""
+    affine_unit(q::UnionAbstractQuantity, symbol::Symbol=:nothing)
+
+Converts a quantity to its nearest affine unit (with scale=ustrip(q) and offset=0.0)
```
MilesCranmer, line 130: Not sure what "nearest" means here, worth elaborating on

src/affine_dimensions.jl
```diff
@@ -0,0 +1,452 @@
...
+
+AffineDimensions adds a scale and offset to Dimensions{R} allowing the expression of affine transformations of units (for example °C)
+The offset parameter is in SI units (i.e. having the dimension of basedim)
+"""
+@kwdef struct AffineDimensions{R} <: AbstractAffineDimensions{R}
+    scale::Float64 = 1.0
+    offset::Float64 = 0.0
+    basedim::Dimensions{R}
+    symbol::Symbol = :nothing
+end
```
MilesCranmer, line 19: Could you please avoid using `@kwdef` here? See how the other structs are manually defined. I think `@kwdef` is a bit dangerous because it permits too many possible constructors, many of which we don't actually want

src/affine_dimensions.jl
```diff
@@ -0,0 +1,452 @@
...
+
+# Inferring the type parameter R
+AffineDimensions(s, o, dims::AbstractDimensions{R}, sym::Symbol=:nothing) where {R} = AffineDimensions{R}(s, o, dims, sym)
+AffineDimensions(s, o, q::UnionAbstractQuantity{<:Any,<:AbstractDimensions{R}}, sym::Symbol=:nothing) where {R} = AffineDimensions{R}(s, o, q, sym)
+AffineDimensions(d::Dimensions{R}) where R = AffineDimensions{R}(scale=1.0, offset=0.0, basedim=d, symbol=:nothing)
+
+# Affine dimensions from other affine dimensions 
+function AffineDimensions{R}(s::Real, o::Real, dims::AbstractAffineDimensions, sym::Symbol=:nothing) where {R}
+    new_s = s*scale(dims)
```
MilesCranmer, line 28: `s * scale(dims)`

src/affine_dimensions.jl
```diff
@@ -0,0 +1,452 @@
...
+end
+
+function map_dimensions(op::typeof(-), args::AffineDimensions...) 
+    assert_no_offset.(args)
+    return AffineDimensions(
+        scale=/(scale.(args)...),
+        offset=zero(Float64),
+        basedim=map_dimensions(op, basedim.(args)...)
+    )
+end
```
MilesCranmer, line 232: Please minimize writing the same code twice. Here, you could write this out like:

```
+for (op, combine) in ((:+, :*), (:-, :/))
+    @eval function map_dimensions(op::typeof($op), args::AffineDimensions...)
+        map(assert_no_offset, args)
+        return AffineDimensions(
+            scale=($combine)(map(scale, args)...),
+            offset=zero(Float64),
+            basedim=map_dimensions(op, map(basedim, args)...) 
+        )
+    end
 end
```

Also note the `map` instead of `.`. Using `map` tends to be safer unless you are dealing with a multi-dimensional array.

src/affine_dimensions.jl
```diff
@@ -0,0 +1,452 @@
...
+    end
+end
+
+for op in (:(==), :(≈))
+    @eval Base.$op(q1::UnionAbstractQuantity{<:Any,<:AffineDimensions}, q2::UnionAbstractQuantity{<:Any,<:AffineDimensions})   = $op(siunits(q1), siunits(q2))
+    @eval Base.$op(q1::UnionAbstractQuantity{<:Any,<:AffineDimensions}, q2::UnionAbstractQuantity{<:Any,<:AbstractDimensions}) = $op(siunits(q1), siunits(q2))
+    @eval Base.$op(q1::UnionAbstractQuantity{<:Any,<:AbstractDimensions}, q2::UnionAbstractQuantity{<:Any,<:AffineDimensions}) = $op(siunits(q1), siunits(q2))
+end
+
+Base.:(==)(d1::AffineDimensions, d2::AffineDimensions) = (d1.scale==d2.scale) & (d1.offset==d2.offset) & (d1.basedim==d2.basedim)
```
MilesCranmer, line 286: These should be `&&`, not `&`.

src/affine_dimensions.jl
```diff
@@ -0,0 +1,452 @@
...
+# This function works like uexpand but will throw an error if the offset is 0
+function _no_offset_expand(q::Q) where {T,R,D<:AbstractAffineDimensions{R},Q<:UnionAbstractQuantity{T,D}}
+    return convert(with_type_parameters(Q, T, Dimensions{R}), q)
+end
+
+# Addition will return Quantity{T, Dimensions}
+Base.:+(q1::UnionAbstractQuantity{<:Any,<:AffineDimensions}, q2::UnionAbstractQuantity{<:Any,<:AffineDimensions}) = _no_offset_expand(q1) + _no_offset_expand(q2)
+
+# Subtraction will return Quantity{T, Dimensions}, in special cases, differences between offsetted AffineDimensions is allowed as offsets cancel out
+function Base.:-(q1::UnionAbstractQuantity{<:Any,<:AffineDimensions}, q2::UnionAbstractQuantity{<:Any,<:AffineDimensions})
```
MilesCranmer, line 272: By convention you should write this as `Base.:(-)`

MilesCranmer (commented):
The degC/degF are currently not importable:

```
julia> using DynamicQuantities.AffineUnits: degC
WARNING: could not import AffineUnits.degC into Main

julia> using DynamicQuantities.AffineUnits: °C
WARNING: could not import AffineUnits.°C into Main
```

src/affine_dimensions.jl
```diff
@@ -0,0 +1,452 @@
...
+end
+
+
+scale(d::AffineDimensions) = d.scale
+offset(d::AffineDimensions) = d.offset
+basedim(d::AffineDimensions) = d.basedim
+
+with_type_parameters(::Type{<:AffineDimensions}, ::Type{R}) where {R} = AffineDimensions{R}
+constructorof(::Type{AffineDimensions}) = AffineDimensions{DEFAULT_DIM_BASE_TYPE}
+constructorof(::Type{AffineDimensions{R}}) where {R} = AffineDimensions{R}
```
MilesCranmer, line 78: These are incorrect; see the other instances of `constructorof`

Deduction42 (commented):
There are still some open questions regarding to your comments. Please answer them and I'll try to implement our resolutions as soon as possible.

MilesCranmer:
> I can see why the first one is incorrect (which I changed) but the second one is pulled directly from [symbolic_dimensions.jl](https://github.com/SymbolicML/DynamicQuantities.jl/blob/3386209ae76cac6d9a2f130f500a6474119001a5/src/symbolic_dimensions.jl#L107)
>
> Can you explain to me why the second one is wrong?

`SymbolicDimensionsSingleton{R}` is a special case because it doesn’t actually store any field with `R`, so it requires it in the constructor. Other ones can get it from the contents.

MilesCranmer:
```julia
uexpand(q::UnionAbstractQuantity{<:Any,<:Dimensions}) = q
```
This seems pretty reasonable to add! In fact I wish we had it already.

src/affine_dimensions.jl
Deduction42: I saw the convention you used in "utils.jl" and decided to follow that:
`scale(d)` is now `uscale(d)`
`offset(d)` is now `uoffset(d)`
`basedim(d)` is now `ubasedim(d)`

src/affine_dimensions.jl
Deduction42: I'll keep it just in case someone wants to add new functions for "map_dimensions".

src/affine_dimensions.jl
Deduction42: It turns out that this was fixed since I initially wrote this. I'm changing all `siunits(q)` to `uexpand(q)`

README.md
MilesCranmer: Unresolving this as it seems it’s still on the readme. Can you please delete the affine units from the README? It is a very small part of the library so I don’t think needs to be mentioned here. Advanced users can find it on the docs.

MilesCranmer:
> This is a generic fallback for map_dimensions; it's slower and less accurate than the specific versions but is much more flexible. I've made specific versions for exponentiation, multiplication and division, but just in case someone defines another method, this is there as a catch-all. It's technically not needed for core functionality and I only test this once. Let me know if you want me to remove it due to YAGNI
> I'll keep it just in case someone wants to add new functions for "map_dimensions".

Please remove this generic version of `map_dimensions`. Note that map dimensions is an internal function and none of the other dimensions types have a generic version. (Please keep the PR as simple as possible)

src/affine_dimensions.jl
```diff
@@ -0,0 +1,456 @@
...
+function Base.:(-)(q1::UnionAbstractQuantity{<:Any,<:AffineDimensions}, q2::UnionAbstractQuantity{<:Any,<:AffineDimensions})
+    if dimension(q1) == dimension(q2)
+        return uexpand(q1) - uexpand(q2)
+    else
+        return _no_offset_expand(q1) - _no_offset_expand(q2)
+    end
+end
+
+for op in (:(==), :(≈))
+    @eval Base.$op(q1::UnionAbstractQuantity{<:Any,<:AffineDimensions}, q2::UnionAbstractQuantity{<:Any,<:AffineDimensions})   = $op(uexpand(q1), uexpand(q2))
```
MilesCranmer, line 278: Are these needed? I would have thought the promotion rules would already get it to work

README.md
Deduction42: I'll remove the other two sections, but I have people on my team still suffering through Unitful type-instabilities because this feature is missing, and this PR already has already been mentioned in another Github discussion. I think I can remove the custom affine units and operations and tell them to refer to the docs, but the readme should at least mention this.

Deduction42:
I renamed the functions to affscale, affoffset, and affbasedim (these functions are used frequently and in some long expressions, so I still prefer shorter forms). I also removed the generic fallback of map_dimensions

src/affine_dimensions.jl
```diff
@@ -0,0 +1,387 @@
...
+    return _no_offset_expand(q1) + _no_offset_expand(q2)
+end
+
+# Subtraction
+function Base.:(-)(q1::UnionAbstractQuantity{<:Any,<:AffineDimensions}, q2::UnionAbstractQuantity{<:Any,<:AffineDimensions})
+    if dimension(q1) == dimension(q2)
+        return uexpand(q1) - uexpand(q2)
+    else
+        return _no_offset_expand(q1) - _no_offset_expand(q2)
+    end
```
MilesCranmer, line 255: Why is there special behavior here?

src/affine_dimensions.jl
```diff
@@ -0,0 +1,390 @@
...
+        print(io, "(", affine_scale(d), " ", affine_base_dim(d),")")
+    else
+        addsign = affine_offset(d) < 0 ? "-" : "+"
+        if isone(affine_scale(d))
+            print(io, "(", addsign, abs(affine_offset(d)), affine_base_dim(d), ")")
+        else
+            print(io, "(", affine_scale(d), addsign, abs(affine_offset(d)), " ", affine_base_dim(d),")")
+        end
+    end
+end
```
MilesCranmer, line 107: The printing for generic affine quantities is a bit confusing:

```julia
julia> x = AffineDimensions(1.0, 2.0, Dimensions(length=2, time=-2))
(+2.0m² s⁻²)

julia> x = AffineDimensions(10.0, 2.0, Dimensions(length=2, time=-2))
(10.0+2.0 m² s⁻²)
```

The last equation looks to be dimensionally invalid.

Can this be polished? How do other packages indicate affine units?

src/affine_dimensions.jl
```diff
@@ -0,0 +1,374 @@
...
+# Base case with SI units
+function AffineDimensions{R}(s::Real, o::UnionAbstractQuantity{<:Any,<:Dimensions}, q::UnionAbstractQuantity{<:Any,<:Dimensions}, sym=:nothing) where {R}
+    dimension(o) == dimension(q) || throw(DimensionError(o, q))
+    return AffineDimensions{R}(s * ustrip(q), ustrip(o), dimension(q), sym)
+end
+
+# Offset from real
+function AffineDimensions{R}(s::Real, o::Real, q::Q, sym=:nothing) where {R, Q<:UnionAbstractQuantity}
+    return AffineDimensions{R}(s, o * q, q, sym)
+end
```
MilesCranmer, line 83: Having trouble understanding the use of this function

src/affine_dimensions.jl
Deduction42: Subtraction between two values of the same affine unit is typically an allowable operation that returns a non-affine unit because the offsets cancel each other out. If they're different affine units with non-zero offset, an error is thrown.

src/affine_dimensions.jl
Deduction42: If someone uses a real value for the offset and a quantity as the base dimension, it assumes that the offset has hte same scale as the quantity. For example:

AffineDimensions{R}(1.0, 101.3, u"kPa", sym=:nothing)

Means that the offset is 101.3"kPa" (scale is calculated by converting u"kPa" to SI units)

MilesCranmer: I think it is a bit dangerous to have this be argument-based, since it is very similar to the signature of creating a new `AffineDimensions` from a dimensions object. Perhaps if it were keyword based it would be safer. Or if there was a separate function entirely for this

Deduction42: You can use keywords for this as well. Keyword versions simply pass to the argument-based functions which trigger the dispatch (functions will specialize on keywords but you can't dispatch on them, which is why I get around this with forwarding to dispatched argument-based functions)

src/affine_dimensions.jl
MilesCranmer: Just to check, is this type stable or unstable?

Deduction42: Yes, it should always produce an SI quantity, it's just that `_no_offset_uexpand(q)` will throw an error if the offset is not zero.

Deduction42:
So I noticed something when I tested at local version of this package in one of my industrial projects. Our unit registration macros behave very differently.

`@register_unit` will throw an error if you try to re-register a unit, even if it has the same value. This is horribly limiting because I can't re-run a script that includes unit registration unless I completely restart Julia (which is so painful that Revise.jl was invented). What's worse is that any attempt to put `@register_unit` inside a control-flow block seems to cause unit registration to not take. The only way I could make this work is to put the @register_unit statements in another file, and conditionally `include` the file if the units don't exist.

`@register_affine_unit` will throw a warning, saying that the unit is already registered and that no changes will be made. This is perhaps a bit verbose if you're just re-running a script and the values don't change but was way easier to implement in my solution.

Do you think it may be best to just throw warnings, or to just split the difference and only throw an error if one attempts to re-register a unit with a _**different**_ value and do nothing otherwise (kid of like how struct definitions work)?

Deduction42:
Other than the unit registration behavior, this update has made a huge quality-of-life improvement for one of my projects (even with the annoying limitations of using a local module). Being able to register the "gauge-psi" sensor and handle that other "fahrenheit" sensor without using Unitful.jl removed hundreds of lines of code and dramatically reduced precompile times when building a docker image. I'm so excited for this!

src/affine_dimensions.jl
```diff
@@ -0,0 +1,368 @@
...
+        return uexpand(q1) - uexpand(q2)
+    else
+        return _no_offset_expand(q1) - _no_offset_expand(q2)
+    end
+end
+
+for op in (:(==), :(≈))
+    @eval begin
+        function Base.$op(q1::UnionAbstractQuantity{<:Any,<:AffineDimensions}, q2::UnionAbstractQuantity{<:Any,<:AffineDimensions})
+            $op(uexpand(q1), uexpand(q2))
```
MilesCranmer, line 231: One worry I have here is that this slightly differs from the behavior of symbolic dimensions, which otherwise seem similar:
```julia
julia> x = 22ua"degC"
22.0 °C

julia> y = 71.6ua"degF"  # Same quantity; different units
71.6 °F

julia> x ≈ y
true

julia> xs = 1us"m"
1.0 m

julia> ys = xs |> us"mm"  # Same quantity; different units
1000.0 mm

julia> xs == ys
false
```

Note that symbolic units don't automatically convert at the moment. Is it bad that they don't, but affine units do? I'm not sure. What are your thoughts.

In some ways, is `==` supposed to be literal "is identical", or is it meant to be "equal under a normalization to the same units". If it's the latter, it is in conflict with symbolic units. Which is fine; I guess some of the use of symbolic units is for explicitly checking such things. And affine units just happen to be symbolic for some other reason.

But at the same time I wonder if affine units should also have this behavior where `==` checks equality of dimensions AND equality of values. And only if someone converts to the same unit system, does it become equal. I'm really not sure about this.

MilesCranmer: I think this is one area where I feel like DynamicQuantities.jl is supposed to be more strict than Unitful.jl, and have less convenience functions that are hard to debug.

Deduction42: I thought that was what "===" was for. I actually only used this behaviour because you subtyped AbstractQuantity as "Number". What I would expect to happen then is that "==" tests for _**equality of value**_, while "===" tests for **_identicality_**.

For example, for Float/Int
julia> 1 == 1.0
true
julia> 1 === 1.0
false

Then, if units are numbers I would expect:
julia> -40ua"°C" == -40ua"°F"
true
julia> -40ua"°C" === -40ua"°F"
false

MilesCranmer: > Clearly, the symbolic quantity is the odd one out here.

I mean, one of these is not yet part of DQ :-)

MilesCranmer: It's not true that `===` measures identical values. For example:

```julia
julia> 5us"m" === 5us"m"
false
```

This is because there is mutable storage, so these objects have different `objectid`.

MilesCranmer: The default equality checks every field for equality:

```julia
julia> struct A
           value::Float64
           dimension::Symbol
       end

julia> A(1.0, :km) == A(1000.0, :m)
false
```

Neither `SymbolicDimensions` nor `Dimensions` modify this behavior. Only
when all fields, recursively, are equal, are the objects marked equal.

The only reason that `1000u"m"` and `1u"km"` are marked equal is
because the underlying objects are literally equal in all respects:

```julia
julia> x = 1000u"m"
1000.0 m

julia> x = 1u"km"
1000.0 m
```

The reason someone would use `SymbolicDimensions` is because they don't
wish to convert to automatically convert to SI base units. So I personally
find it safer and more in-line with the Julia standard behavior to not
change this.

Deduction42: Okay, ultimately it's your call. Did you want me to make the change? I can do that. Also, did you see my comment on @regitster_unit behaviour? I don't think it should error if the value is the same. It wrecks havoc with re-running scripts and even Revise.jl.

Deduction42 (commented):
So I noticed something when I tested at local version of this package in one of my industrial projects. Our unit registration macros behave differently when attempting to re-register a unit.

`@register_unit` will throw an error if you try to re-register a unit, even if it has the same value. This is horribly limiting because I can't re-run a script that includes unit registration unless I completely restart Julia (which is so painful that Revise.jl was invented). What's worse is that any attempt to put `@register_unit` inside a control-flow block seems to cause unit registration to not take. The only way I could make this work is to put the `@register_unit` statements in another file, and put the include statement inside the control flow (so that it only includes if the units don't exist).

`@register_affine_unit` will throw a warning, saying that the unit is already registered and that no changes were made. This is perhaps a bit verbose if you're just re-running a script and the values don't change but was much easier to implement in my solution.

Do you think it may be best to just throw warnings, or to just split the difference and only throw an error if one attempts to re-register a unit with a _different_ value and do nothing otherwise (kid of like how struct definitions work)?

src/affine_dimensions.jl
Deduction42: Anyway, I'm in Western Canada, so I'll be out sleeping for the next 8 hours so if you want to so anything yourself, I won't be doing anything. However, if you want me to make some changes, I'll be ready in about 10 or so hours.

MilesCranmer:
What does Unitful do for trying to register twice?

MilesCranmer:
You shouldn’t be re-running a script though. You should use `Revise.includet` rather than `include`, and re-run a function, not include a file

MilesCranmer:
If two packages register the same unit symbol, but in different ways, a warning would introduce some really sinister bugs. If you really want to include a script twice, just wrap the registration with a try-catch.

Deduction42:
I tried to wrap it in a try-catch and an if-statement but units don't register for some reason unless I do an "include". For example, I try to register "lb" and then "psi" but the second registration fails because the unit "lb" doesn't exist in that scope for some reason.

The way Unitful tries to get you to do things is to put all your unit registrations inside a module and then use @register on the module. This is because units are type-unstable so you will only be able to know the unit type if it's a constant in a module. What's weirder is that you can actually change and re-register units up until the point that you use them. But once you use them, trying to re-register them as something else silently fails. For example:
```
julia> @unit M "M" Molar 2u"mol/L" true #Incorrect value
julia> @unit M "M" Molar 1u"mol/L" true #Corrected the value
julia> uconvert(u"mol/L", Quantity(1,u"M")) #The correction takes
1 mol L^-1
julia> @unit M "M" Molar 2u"mol/L" true #Try to re-register the incorrect value
M
julia> uconvert(u"mol/L", Quantity(1,u"M")) #The incorrect value doesn't take
1 mol L^-1
```
I think the best approach on registering a unit is to
1. Check to see if the unit exists, if not, register the unit
2. If the unit exists, check if the "value to be registered" is different
3. If the attempted unit is something that is different than what exists, throw an error, otherwise, do nothing

Currently, neither @register_unit nor @register_affine_unit work this way. However, I think this way is not likely going to break anything (because re-registering the same unit doesn't introduce bugs because nothing changes), and it doesn't break Revise.jl things if you're trying to register units inside a package. I'm going to change @register_affine_unit to have this behaviour.

Deduction42:
I made the changes so that attempting to re-register different affine units will produce an error (but re-registering the same ones will do nothing). If you could mimic the same behaviour, I think this is the best tradeoff between safety and usability.

src/affine_dimensions.jl
Deduction42: I just realized another problem with your approach to equality. When you compare quantities with different unit types, promotion happens.
julia> 1000u"mm" == 1u"m"
true
julia> 1000u"mm" == 1us"m"
true
julia> 1000us"mm" == 1us"m"
false
I find the last result "astonishing" given the second result. Should not the second result also give "false"?

MilesCranmer (commented):
There are still some issues related to the other mathematical operators. For example, `mod` does not error

src/affine_dimensions.jl
```diff
@@ -0,0 +1,361 @@
...
+        #Check to make sure the unit's name is not :nothing (default)
+        name = newdims.symbol
+        if name == :nothing
+            error("Cannot register a unit if its symbol is :nothing")
+        end
+
+        ind = get(AFFINE_UNIT_MAPPING, name, INDEX_TYPE(0))
+        if !iszero(ind)
+            olddims = dimension(AFFINE_UNIT_VALUES[ind])
+            if (olddims.scale != newdims.scale) || (olddims.offset != newdims.offset) || (olddims.basedim != newdims.basedim)
```
MilesCranmer, line 273: Please use the getters (affine_scale, etc.). I also added one for the symbol

MilesCranmer:
As a heads up I'm making some quick changes as this is taking a long time to go back-and-forth. I removed the definition of `==` and also removed the `-` behaviour when the dimensions are the same. The implementation of affine dimensions is already really complex, so I want to avoid these kinds of convenience functions and value-dependent branches.

src/affine_dimensions.jl
MilesCranmer: Yeah this is just due to promotion rules.

The behavior of `==` not auto-expanding is actually one of the main reasons for the existence of SymbolicDimensions. The JuMP.jl people wanted it so they could error if the SI prefixes on values were different, because I guess the internals are all unit free. 

On my end I kind of just want to throw out everything and go back to `Dimensions` when it was nice and simple and safe and easy to maintain... Ugh. Oh well.

test/unittests.jl
```diff
@@ -2097,25 +2237,49 @@ end
...
+        @test length(UNIT_MAPPING) == map_count_before_registering + 4
+        @test length(ALL_MAPPING) == all_map_count_before_registering + 4
+        @test length(AFFINE_UNIT_MAPPING) == affine_count_before_registering + 7
     end
 
     for my_unit in (MySV, MyV)
         @test my_unit in UNIT_VALUES
         @test my_unit in ALL_VALUES
         @test my_unit in SYMBOLIC_UNIT_VALUES
+        @test my_unit in AFFINE_UNIT_VALUES #Non-affine units should also be registered
```
MilesCranmer, line 2268: Please always use a space before the text in a comment. This is done throughout the package

test/unittests.jl
```diff
@@ -2087,8 +2211,24 @@ end
...
     @eval @register_unit MySV2 us"km/h"
 end
+if :psi ∉ UNIT_SYMBOLS
+    @eval @register_unit psi 6.89476us"kPa"
+end
+
+if :psig ∉ AFFINE_UNIT_SYMBOLS
+    @eval @register_affine_unit psig AffineDimensions(offset=u"Constants.atm", basedim=u"psi")
+else
+    skipped_register_unit = true
```
MilesCranmer, line 2221: I don't think this does anything. At this point in the code it should already be true. Please delete

test/unittests.jl
```diff
@@ -2097,25 +2237,49 @@ end
...
     @test MyV === u"V"
     @test MyV == us"V"
     @test MySV == us"V"
     @test MySV2 == us"km/h"
+    @test MySV == ua"V"
+    @test MySV2 == ua"km/h"
+    @test psi == ua"psi"
+    @test psi == u"psi"
+    @test psig == ua"psig"
+    @test_throws AffineOffsetError 0*psig == u"Constants.atm"
```
MilesCranmer, line 2254: Please use single spaces between infix operators. So `0 * psig`

test/unittests.jl
```diff
@@ -2000,6 +2002,126 @@ end
...
+    @test isnothing(DynamicQuantities.update_external_affine_unit(dimension(°C))) #same value yields for dimension
+    @test_throws "Cannot register a unit if its symbol is :nothing" DynamicQuantities.update_external_affine_unit(celsius) #cannot register :nothing
+
+    #Cannot re-register a unit if its value changes
+    @test_throws "Unit `°C` already exists as `(scale = 1.0, offset = 273.15, basedim = K)`, its value cannot be changed to `(scale = 2.0, offset = 273.15, basedim = K)`" DynamicQuantities.update_external_affine_unit(:°C, 2*°C)
+
+    # Test map_dimensions
+    @test map_dimensions(+, dimension(ua"m/s"), dimension(ua"m/s")) == AffineDimensions(Dimensions(length=2, time=-2))
+    @test map_dimensions(-, dimension(ua"m"), dimension(ua"s"))     == AffineDimensions(Dimensions(length=1, time=-1))
+    @test map_dimensions(Base.Fix1(*,2), dimension(ua"m/s"))        == AffineDimensions(Dimensions(length=2, time=-2))
```
MilesCranmer, line 2119: Please avoid these kinds of stylistic patterns with multiple spaces to line up the `==`

MilesCranmer:
I am worried we are playing whack-a-mole with these issues and I am concerned with how many reviews I have had to do of this PR which is supposed to be for a fairly simple purpose. Maybe the current design of the AffineDimensions is just not compatible with DQ, if we need to keep monkey patching stuff like this?

For example, what if a user has their own function that they define for quantities, and, as done throughout the entirety of DynamicQuantities, they simply do a `dimension(q) == dimension(q2)` check before doing an `op(ustrip(q), ustrip(q2))` on the units and computing the result. If someone were to pass `AffineDimensions` into that function, it would give incorrect result, and be completely silent about it. It seems like (at the moment) the solution is to just patch stuff manually so that it works for `AffineDimensions`.

I'm not sure how to solve this though. It seems really difficult...

MilesCranmer:
The clearest example of this is `mod`. We shouldn't need to patch `mod` within the affine_dimensions.jl code. It should just work with the existing code. If it doesn't work I worry there is a deeper incompatibility, or maybe a redesign is needed. 

Like maybe a different approach could be that `ua"degC"` returns a special quantity which is NOT a subtype of `AbstractQuantity`, and is only defined for multiplication, at which point it immediately starts using `Dimensions`. So you would be forced to write something like `22ua"degC"` - but the result would be in `Dimensions`, not `AffineDimensions`. So we wouldn't need to worry about this other stuff.

The downside is that a raw `ua"degC"` wouldn't be a quantity. But maybe that's precisely what we want.

test/unittests.jl
Deduction42: Addressed these where I saw them

MilesCranmer:
Or maybe just have ustrip return the scaled value, and dimension return the base dimension? Then I guess stuff would work out-of-the-box?

Deduction42:
> For example, what if a user has their own function that they define for quantities, and, as done throughout the entirety of DynamicQuantities, they simply do a `dimension(q) == dimension(q2)` check before doing an `op(ustrip(q), ustrip(q2))` on the units and computing the result. If someone were to pass `AffineDimensions` into that function, it would give incorrect result, and be completely silent about it. It seems like (at the moment) the solution is to just patch stuff manually so that it works for `AffineDimensions`.

The most "Julian" solution to this is to properly constrain the parameters and define the operation on those dimensions
```
function my_op(q1::Quantity, q2::Quantity)
    return Quantity(my_op(ustrip(q1), ustrip(q2)), my_op(dimension(d1), dimension(d2)))
end
function my_op(d1::ScaledDimension, d2::ScaledDimension)
   return d1==d2 ? d1 : DimensionError(l, r)
end
```
As long as the op isn't defined for AffineDimension, it should work. Defining a function for AbstractDimension is insane, because you don't know what kind of whak-a-doodle dimensions you can get.

Deduction42:
You don't even have to redefine "my_op" for dimensions. You could also simply have a function called `scaled_dim_check`
```
function my_op(q1::Quantity, q2::Quantity)
    return Quantity(my_op(ustrip(q1), ustrip(q2)), scaled_dim_check(dimension(q1), dimension(q2)))
end
scaled_dim_check(d1::AbstractScaledDimension, d2::AbstractScaledDimension) =  (d1==d2) ? d1 : throw(DimensionError(l, r))
```
This will also throw errors on AffineDimensions. Or you could just constrain the the quantity dimension types
```
function my_op(q1::Quantity{<:Number, <:AbstractScaledDimension}, q2::Quantity{<:Number, <:AbstractScaledDimension)
    (d1==d2) || throw(DimensionError(l, r))
    return Quantity(my_op(ustrip(q1), ustrip(q2)), d1)
end
```
The key is using the correct level of abstraction.

Deduction42:
I think in this case, it's best to define `AbstractDimension` as a subtype of `AbstractScaledOrAffine` and just not export the latter. That way, people are less likely to fall into the affine trap. If people want to write their own functions, warn them to make sure they use `Quantity{<:Number, <:AbstractDimension}`. As it stands now, as long as we keep the internal set of operators constrained, whack-a-mole is a legitimate strategy.

MilesCranmer:
Thanks for the detailed thoughts. Given that DynamicQuantities is already at v1 with a stable API, we can’t ask users to restrict their function definitions to `Quantity{<:Number, <:AbstractDimension}` without breaking the API contract. For me this is a non-option.

I appreciate the convenience of treating `AffineDimensions` like a typical `AbstractDimension` and incrementally patching uncovered methods. However, I'm concerned this approach would silently introduce correctness issues—especially in cases where users assume compatibility based on the current API.

A safer route would be to either ensure `AffineDimensions` reliably integrate with the existing `dimension`/`ustrip` API by simply throwing errors when those methods are used (with custom `affine_dimension` or `affine_ustrip` to be explicit), or use `ua"degC"` to produce a specialized, restricted type which is completely distinct from AbstractQuantity. Those ideas would prevent some of the more sinister correctness errors.

Deduction42:
At this point, I'm not sure what kind of problems this would fix because how you implemented `SymbolicDimensions` and various quantities is rather confusing to me. I'm also not sure how to validate if these changes fix your problems. You're welcome to attempt this but I'm not sure I can help you here.

MilesCranmer:
I'm confused. Why wouldn't the ideas above work? These ones:

- Throwing errors when `ustrip` or `dimension` are used; instead requiring `affine_dimension` or `affine_ustrip` to be called.
  - This automatically errors for any functions with `ustrip(x)`, which is exactly the part of the API that would cause silent bugs with affine dimensions. So this would I think basically solve the issue entirely.
- Have `ua"degC"` by a specialized, restricted type which is completely distinct from AbstractQuantity, and only define `/` and `*` which would convert to a regular `Quantity{_, <:Dimensions}`

Deduction42:
It would technically work but would require a whole separate API for affine units, where the API in Unitful.jl is pretty straightforward. I'm already trying to use this version of the API in some of my codebases with local versions of this repo and having to make separate functions for affine units will just make this package a headache to use. I'm now having to play whack-a-mole to break an API that makes perfect sense.

Deduction42:
So from my end:
1. I need to be able to have generic `Quantity{T, <:AffineDimensions}` and `ua"degC"` needs to return that. I already have an more general affine unit registry that can contain both symbolic and affine units so that I can have type-stable objects containing both affine and non-affine units. Messing with this behaviour is a non-starter. 
2. Breaking `dimension(q::Quantity{T, <:AffineDimensions}` is drastic, and is also a non-starter. Grabbing `AffineDimensions` from a quantity is a reasonable operation. I'd sooner break "==" on two affine dimensions.
3. I wholeheartedly agree with breaking `ustrip` and creating `affine_ustrip` because `ustrip` on an affine dimension is dangerous operation anyway. I think this will solve all the real problems we're having.

MilesCranmer:
Ok maybe let’s try 3? Maybe that will help fix some of the issues

Deduction42:
Okay, I'll give it a shot.

Deduction42:
Welp, breaking ustrip really broke a lot of things. I'm at the point where I don't know if I can fix the remaining issues. I'm having a hard time with the Julia 1.10 implementation, the problem seems to be in utils.jl file around line 358, I'm not sure why Julia 1.10 is breaking here. This part of the code is hard for me to understand so I'm not sure if it's safe to patch. Are you able to look into this one?

MilesCranmer:
I can take a look. Before that can you try to fix your merged changes from edits? I think some of my edits were erased; stuff like adding `affine_symbol` and `PLACEHOLDER_SYMBOL`

Deduction42:
Shoot, I thought I force-pulled from your changes. I'm not sure how to fix that.

Deduction42:
I'm probably going to be hands-off this for a while. You can safely edit at will.

MilesCranmer:
Before handing over the reins, can you put back 31c8bfea96e77322aa6747f4aa4d216283f62079? I don’t know how you merged the changes but that commit seems to be gone and also `affine_symbol`. Also some of the other commits around that time look to be overwritten by your merge

Deduction42:
Okay, I did the best I could; there were a lot of merge conflicts.

Deduction42:
I'm having second thoughts on this though, breaking ustrip really broke a lot of things and it required a _lot_ of code to be copied over, so I'm worried about shotgun surgery. Maybe this package just isn't ready for affine units.

MilesCranmer:
Breaking `ustrip` is kind of *correct* though, no? Since the interpretation is essentially ambiguous when we are talking about affine units.

What does Unitful do for stripping units in an affine context?

Deduction42:
In Unitful.jl, it defines `ustrip(q::Quantity)=q.val` so it would work for all units, because there are a lot of valid times to use `ustrip` on affine quantities (a lot more than I previously thought).

For numerical functions, Unitful.jl defines a couple of useful derived classes:
```
const AffineQuantity{T,D,U} = AbstractQuantity{T,D,U} where U<:AffineUnits
const ScalarQuantity{T,D,U} = AbstractQuantity{T,D,U} where U<:ScalarUnits
```
Users can therefore define operations that only work on ScalarQuantity, which I think is the correct approach for this.

MilesCranmer:
What is the practical use of defining functions for affine quantities? I thought the main objective is to move stuff to SI quantities and then operate there. I mean there’s only degC and degF; what’s the main benefit here. I don’t want to shoot it down, but like, why does it justify so much complexity? I think Unitful is overly complex at the moment, and they can’t go back on that; they are stuck with increased complexity.

MilesCranmer:
If I scroll through https://github.com/search?q=/AffineQuantity/%20language:julia&type=code, it looks like most appearances of AffineQuantity in libraries is just to convert it to regular quantities or throw errors…

MilesCranmer:
Actually it looks even less frequently used than I thought. Do any libraries use affine quantities internally for computing things? Most uses seem like just compatibility layers for handling all possible Unitful inputs

Deduction42:
That's correct, it's basically an add-on. The thing is though, since every unit is it's own distinct type, adding AffineUnits is pretty easy. However, with DynamicQuantities, we want to make sure we can represent more units with a single type. However, you can't represent an AffineDimension with a SymbolicDimension (but fortunately, AffineDimensions can represent SymbolicDimensions). This means that if I'm going to read a spreadsheet of units from a technician, it's best if I read it with AffineDimensions because all units will work even if there's the odd Fahrenheit in there. This also means that I now need to be able to parse raw input as AffineDimensions, in fact, I probably will always use `ua"..."` and `aff_uparse("...")` because it will work for everything, and then convert to SI at the first chance I get.

Deduction42:
As far as end users go, you only ever do two things with affine units: (1) Take differences, (2) Convert them to scalar units. Affine units are usually only temperatures (sometimes gauge pressure), and people who do calculations with temperature are sufficiently acquainted with thermodynamics to know that you only do operations in Celsius/Fahrenheit for temperature differences, otherwise you use absolute temperatures. If a function does actual math with Celsius/Fahrenheit, it's for an empirical model that's been trained on those units.

MilesCranmer:
Sorry for a maybe dumb question, but why not handle this dataset with a regular expression? What does your dataset look like?

Deduction42:
Anyway, I guess what I'm saying is that users should know how to use affine units, so I'm not sure we have to worry too much about users defining functions and silently returning bad results, because if there are calculations, affine units will break pretty quickly (you won't get very far without +-*./) even without breaking ustrip (I know, because it happens to me with Unitful and Pint (python)).

MilesCranmer:
Alternative implementation here: https://github.com/SymbolicML/DynamicQuantities.jl/pull/168. Just leaves the absolute basics

MilesCranmer:
Moving to #168

PR #161, opened 2025-03-06
Add `sibasevalue` function

j-fu:
* Define uexpand for quantities with Dimensions
* Define `sibasevalue(q)=ustrip(uexpand(q))

When obtaining the numerical value of a quantity (e.g. for computations with sparse direct solvers implemented in other languages) via `ustrip`, there is danger of mixing up the right values when quantities with symbolic dimensions come into play.

By adding a `uexpand` method for quantities with dimensions, which just passes through the value one can write code consistent for both flavors of quantities.

`sibasevalue` shall help to remove a little bit of boilerplate and explicitely hint at the characteristics of the returned value.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -48,7 +48,9 @@ makedocs(;
         "Symbolic Units" => "symbolic_units.md",
         "Types" => "types.md",
     ],
-    warnonly = [:missing_docs]
+    warnonly = [:missing_docs],
+    clean = false,
+
 )
 
 deploydocs(;
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -48,9 +48,7 @@ makedocs(;
         "Symbolic Units" => "symbolic_units.md",
         "Types" => "types.md",
     ],
-    warnonly = [:missing_docs],
-    clean = false,
-
+    warnonly = [:missing_docs]
 )
 
 deploydocs(;
--- a/docs/src/symbolic_units.md
+++ b/docs/src/symbolic_units.md
@@ -32,8 +32,8 @@ To convert a quantity in regular base SI units to corresponding symbolic units, 
 uconvert
 ```
 
-In order to obtain the numerical value of the quantity in its SI base units, use `sibasevalue`.
+In order to obtain the numerical value of the quantity in its SI base units, use `ustripexpand`.
 
 ```@docs
-sibasevalue
+ustripexpand
 ```
--- a/docs/src/symbolic_units.md
+++ b/docs/src/symbolic_units.md
@@ -24,6 +24,7 @@ To convert a quantity to its regular base SI units, use `uexpand`:
 
 ```@docs
 uexpand
+ustripexpand
 ```
 
 To convert a quantity in regular base SI units to corresponding symbolic units, use `uconvert`:
@@ -32,8 +33,3 @@ To convert a quantity in regular base SI units to corresponding symbolic units, 
 uconvert
 ```
 
-In order to obtain the numerical value of the quantity in its SI base units, use `ustripexpand`.
-
-```@docs
-ustripexpand
-```
--- a/docs/src/symbolic_units.md
+++ b/docs/src/symbolic_units.md
@@ -31,3 +31,9 @@ To convert a quantity in regular base SI units to corresponding symbolic units, 
 ```@docs
 uconvert
 ```
+
+In order to obtain the numerical value of the quantity in its SI base units, use `sibasevalue`.
+
+```@docs
+sibasevalue
+```
--- a/src/DynamicQuantities.jl
+++ b/src/DynamicQuantities.jl
@@ -8,7 +8,7 @@ export AbstractDimensions, Dimensions, NoDims
 export AbstractSymbolicDimensions, SymbolicDimensions, SymbolicDimensionsSingleton
 export QuantityArray
 export DimensionError
-export ustrip, dimension
+export ustrip, dimension, sibasevalue
 export ulength, umass, utime, ucurrent, utemperature, uluminosity, uamount
 export uparse, @u_str, sym_uparse, @us_str, uexpand, uconvert, @register_unit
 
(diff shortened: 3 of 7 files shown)


MilesCranmer:
Great idea! Two quick comments:

1. I'm not sure about the name. Is there an equivalent function in Unitful that we can use, or maybe in another units package? Any kind of precedent would be helpful.
2. The PR seems to have some unintended changes attached to it (like `docs/Project.toml`), can you please clean those up?

test/unittests.jl
```diff
@@ -855,7 +855,14 @@ end
     @test dimension(inv(us"s") * us"km") == dimension(us"km/s")
     @test dimension(inv(us"s") * us"m") != dimension(us"km/s")
     @test dimension(uexpand(inv(us"s") * us"m")) == dimension(uexpand(us"km/s"))
+    
+
+    @test sibasevalue(u"h") == 3600.0
+    @test sibasevalue(us"h") == 3600.0
+    @test sibasevalue(QuantityArray(fill(1.0,5),u"h")) == fill(3600.0,5)
+    @test sibasevalue(QuantityArray(fill(1.0,5),us"h")) == fill(3600.0,5)
```
MilesCranmer, line 863:
```suggestion
    @test sibasevalue(QuantityArray(fill(1.0,5), u"h")) == fill(3600.0, 5)
    @test sibasevalue(QuantityArray(fill(1.0,5), us"h")) == fill(3600.0, 5)
```

j-fu:
AFAIK there is nothing like this in Unitful. I tried to make a similar PR  there (https://github.com/PainterQubits/Unitful.jl/pull/448), but this was not accepted. As a consequence, I made a whole package around this (https://github.comj-fu/LessUnitful.jl). This essentially provides a string macro which returns the "unit factor" (did'nt call it "si..." due to the preferred units thing) E.g.  `ufac"km"`  would give `1000.0`.  Basically I would like to sunset LessUnitful in favor of DQ, if this workflow can be realized without much boilerplate.

As for other packages, I don't know. I recently looked for some good units package for python and found astropy. I asked DeepSeek about this and it ended up (amazingly!) with proposing
```python
def get_si_value(quantity):
    return quantity.decompose().value
``` 

As for the name - I am open for other ideas.

MilesCranmer:
Could we maybe call it `usivalue` or `ubasevalue`? I think it should have the `u` prefix in any case, like the other functions of similar behavior.

MilesCranmer:
Or just `ubase`

j-fu:
Among those I would prefer `ubasevalue`.

MilesCranmer:
I'm rethinking my suggestion, sorry. So, `base` and `value` both don't really have any existing meaning in the codebase. Whereas for `ustripexpand`, both `expand` and `strip` already mean something. In other words, the body is a `ustrip(uexpand(x))`, not a `uvalue(ubase(x))`. If the latter methods were what we had gone with originally, then it might make more sense to call it that. But I would prefer we don't increase the dictionary of concepts without good reason.

If it is ok with you, can we please call it `ustripexpand`? I guess it doesn't roll off the tongue, but it's very clear what it does.

I also prefer `ustripexpand` over `uexpandstrip` because the former can be tab-completed from `ustrip` which is the non-safe form of it. So it seems kinda nice.

j-fu:
Ok from the logic of the other names in this package `ustripexpand` seems to be a good compromise.

After sleeping about it, I came to my mind that we also might want to have the reverse operation.

We could define
```
uconvert(q::AbstractQuantity, x::Real)= uconvert(q,(x/sibasevalue(q))*q)
```

This gives

```
julia> uconvert(us"h", 3600.0)
1.0 h
```

possibly this could be done also for units with Dimensions.

It also could be named differently.  What do you think ?

MilesCranmer:
I'm not sure about that one, it is a bit unintuitive to me (perhaps just me). But I am totally on board with the main function in this PR.

docs/make.jl
```diff
@@ -48,7 +48,9 @@ makedocs(;
         "Symbolic Units" => "symbolic_units.md",
         "Types" => "types.md",
     ],
-    warnonly = [:missing_docs]
+    warnonly = [:missing_docs],
+    clean = false,
+
 )
```
MilesCranmer, line 54: Please remove these changes

docs/src/symbolic_units.md
```diff
@@ -31,3 +31,9 @@ To convert a quantity in regular base SI units to corresponding symbolic units,
 ```@docs
 uconvert
 ```
+
+In order to obtain the numerical value of the quantity in its SI base units, use `ustripexpand`.
+
+```@docs
+ustripexpand
+```
```
MilesCranmer, line 39: I think it would be better to just add this immediately under `uexpand`. No need for the extra sentence

src/DynamicQuantities.jl
```diff
@@ -8,7 +8,7 @@ export AbstractDimensions, Dimensions, NoDims
 export AbstractSymbolicDimensions, SymbolicDimensions, SymbolicDimensionsSingleton
 export QuantityArray
 export DimensionError
-export ustrip, dimension
+export ustrip, dimension, sibasevalue
```
MilesCranmer, line 11:
```suggestion
export ustrip, dimension, ustripexpand
```

src/symbolic_dimensions.jl
```diff
@@ -251,6 +256,15 @@ function Base.:(|>)(
     return uconvert(qout, q)
 end
 
+"""
+    ustripexpand(q::AbstractQuantity)
+    ustripexpand(q::QuantityArray)
+
+Return the value of `q` in SI base units. This is equivalent with `ustrip(uexpand(q))`.
+"""
+ustripexpand(q::AbstractQuantity) = ustrip(uexpand(q))
```
MilesCranmer, line 265: Shouldn't this be UnionAbstractQuantity? (To pick up RealQuantity too)

j-fu:
Ok I will think about the uconvert stuff and possibly address this in a different PR.

MilesCranmer:
Done in #162

PR #168, opened 2025-03-16
Simple affine units implementation

MilesCranmer:
@Deduction42 

Here is a reduced version of #159 with only the basic functionality. Writing out `degC` will auto-convert to dimensions when you multiply it with a number:

```julia
julia> x = 22ua"degC"
295.15 K

julia> x = 0ua"degC"
273.15 K

julia> 50ua"degC" - 40ua"degF"
45.55555555555554 K
```

That's basically it. I think this gets the main feature of unit parsing but without needing to have a separate `AffineDimensions` object that requires a lot of patching throughout the library.


Deduction42:
What kind of object does `ua"degC"` return? Does `ua"m/s"` work too?

MilesCranmer:
`ua"degC"` returns an `AffineUnits` object. It's outside of the dimensions hierarchy and thus can't be used inside a `Quantity`. Only when you multiply this against a number does it become a quantity.

`ua"m/s"` at the moment does work; it triggers a fall-back method that calls `uparse`.

(Though I was thinking of deleting that, so users would need to be very explicit about getting an `AffineUnits` object)

Deduction42:
I'd actually prefer a string macro that produced units, so this looks like a very clever fix. As long 'ua"..."' and 'aff_uparse("...")' produces a consistent object (AffineUnits) and works on any units in the registry, this is probably good enough.

MilesCranmer:
There’s no conversion to affine units, only from affine to SI. Right now there’s not even a display in affine units, because you would never use a naked `AffineUnits` for anything.

But I feel the reverse from SI to affine + printing might be ok to do in user code. Always good to keep the core library lean.

Also, why do you need the other units accessible in `ua_str`? Since they are incompatible with affine anyways. Why not just do a 

```julia
try
    uparse(s)
catch
    aff_uparse(s)
end
```

I don’t know why DQ would need to do this internally.

MilesCranmer:
Ok I am pretty happy with this. From 744$\rightarrow$217 lines compared to #159, and gets the main functionality without the difficulties of treating affine units as regular `Quantity`. Will merge soon

PR #171, opened 2025-03-26
Add `ustrip(unit, quantity)`

### LINKED ISSUE #170: Feature suggestion: two-argument ustrip(units, quantity)
I've done some looking, and it doesn't seem like this exists yet in DynamicQuantities. If it does (or there is an obvious workaround), happy to hear it.

One of the pieces of Unitful's API that I lean on quite a lot is that the single-argument `ustrip` is discouraged, so that `ustrip(unit, quantity) -> value` is the preferred pattern. In the case of Unitful, that is essential because the precise units (unlike dimensions) often depend on what math was done on it previously.

I understand now that in DynamicQuantities the default is to convert everything to base SI (which tripped me up the first time I tried to make a switch), and I see the reasoning for that; in that light, it makes sense to use a single-argument `ustrip`.

But in the case where I am stripping units to feed into something else, and the quantity should be in non-base-SI units, the two-argument `ustrip` makes a very clean interface for guaranteeing that (no matter what weird units came in) I get a value with the proper units out.

Is the lack of two-argument `ustrip` an explicit design decision? If not, would it be useful for me to prepare a PR?

MilesCranmer:
Fixes #170

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -411,7 +411,7 @@ julia> ustrip(u"km", [1000u"m", 2000u"m"])
 @inline function ustrip(unit::UnionAbstractQuantity, q::UnionAbstractQuantity)
     unit, q = promote_except_value(unit, q)
     dimension(unit) == dimension(q) || throw(DimensionError(unit, q))
-    return ustrip(q / unit)
+    return ustrip(q) / ustrip(unit)
 end
 
 """
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -390,7 +390,6 @@ ustrip(::AbstractDimensions) = error("Cannot remove units from an `AbstractDimen
 
 """
     ustrip(unit::UnionAbstractQuantity, q::UnionAbstractQuantity)
-    ustrip(unit::UnionAbstractQuantity, q::AbstractArray{<:UnionAbstractQuantity})
 
 Convert quantity `q` to the units specified by `unit`, then strip the units.
 This is equivalent to `ustrip(q / unit)`, but also verifies the dimensions are compatible.
@@ -410,12 +409,9 @@ julia> ustrip(u"km", [1000u"m", 2000u"m"])
 ```
 """
 @inline function ustrip(unit::UnionAbstractQuantity, q::UnionAbstractQuantity)
+    unit, q = promote_except_value(unit, q)
     dimension(unit) == dimension(q) || throw(DimensionError(unit, q))
     return ustrip(q / unit)
-end
-@inline function ustrip(unit::UnionAbstractQuantity, q::AbstractArray{<:UnionAbstractQuantity})
-    all(qi -> dimension(qi) == dimension(unit), q) || throw(DimensionError(unit, q))
-    return @. ustrip(q / unit)
 end
 
 """
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -389,6 +389,36 @@ ustrip(::AbstractDimensions) = error("Cannot remove units from an `AbstractDimen
 @inline ustrip(q) = q
 
 """
+    ustrip(unit::UnionAbstractQuantity, q::UnionAbstractQuantity)
+    ustrip(unit::UnionAbstractQuantity, q::AbstractArray{<:UnionAbstractQuantity})
+
+Convert quantity `q` to the units specified by `unit`, then strip the units.
+This is equivalent to `ustrip(q / unit)`, but also verifies the dimensions are compatible.
+
+# Examples
+```julia
+julia> ustrip(u"km", 1000u"m")
+1.0
+
+julia> ustrip(u"s", 1u"minute")
+60.0
+
+julia> ustrip(u"km", [1000u"m", 2000u"m"])
+2-element Vector{Float64}:
+ 1.0
+ 2.0
+```
+"""
+@inline function ustrip(unit::UnionAbstractQuantity, q::UnionAbstractQuantity)
+    dimension(unit) == dimension(q) || throw(DimensionError(unit, q))
+    return ustrip(q / unit)
+end
+@inline function ustrip(unit::UnionAbstractQuantity, q::AbstractArray{<:UnionAbstractQuantity})
+    all(qi -> dimension(qi) == dimension(unit), q) || throw(DimensionError(unit, q))
+    return @. ustrip(q / unit)
+end
+
+"""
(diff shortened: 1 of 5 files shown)


MilesCranmer:
@Ickaser let me know if this is good?

Ickaser:
For reference, I am doing this in https://github.com/LyoHUB/LyoPronto.jl/tree/dynamicquantities . Since I built this with Unitful and DifferentialEquations, the current logic is to add units inside my RHS, and strip off units before handing it back to the ODE solver. That's not a very demanding operation as far as the units go, but it's a start.

As far as running the simulations, this seems to work. Next to try (and possibly the harder question?) will be the parameter fitting routines and the plot recipes.

Notes and questions so far:
- `ustrip` to nondimensional: I might expect `us""` to produce a unitless quantity, like Unitful's `NoUnits`. Is there existing API for something like this? I found myself doing `ustrip(us"m/m", q)`, but that's not as clear as I would like.
- I will want to plot temperatures in Celsius; stripping to Celsius is probably avoidable in most math code (unless people have ugly correlations that take arguments in Celsius), but somewhere in a plot recipe pipeline it would be nice to have somethign like `ustrip(ua"°C", T)`, which currently fails.
- Error message for `ustrip(u"...", q)` should maybe say "you can only uconvert *or ustrip* to quantities with SymbolicDimensions..."

- In several places, had `uconvert(u"K", q)` to catch possible Celsius and make sure was math friendly. Since `-40ua"°C"` returns in Kelvin already, these were unnecessary 👍 
- No `μPa`. Not killer, only showed up as `8.1u"μPa*s"` anyway so just replaced with `8.1e-6u"Pa*s"`

MilesCranmer:
> * `ustrip` to nondimensional: I might expect `us""` to produce a unitless quantity, like Unitful's `NoUnits`. Is there existing API for something like this? I found myself doing `ustrip(us"m/m", q)`, but that's not as clear as I would like.

You can do `us"1"` for unitless quantity.

> * I will want to plot temperatures in Celsius; stripping to Celsius is probably avoidable in most math code (unless people have ugly correlations that take arguments in Celsius), but somewhere in a plot recipe pipeline it would be nice to have somethign like `ustrip(ua"°C", T)`, which currently fails.

Since affine units are a custom struct that live outside the abstract dimensions hierarchy (for safety reasons), this seems like a good way to add a conversion back to degC. @Deduction42 wdyt? We basically just add a method `ustrip(unit::AffineUnit, ::UnionAbstractQuantity)`.

> * Error message for `ustrip(u"...", q)` should maybe say "you can only uconvert _or ustrip_ to quantities with SymbolicDimensions..."

I don't understand this question^ Can you clarify?

> * In several places, had `uconvert(u"K", q)` to catch possible Celsius and make sure was math friendly. Since `-40ua"°C"` returns in Kelvin already, these were unnecessary 👍

:+1:

> * No `μPa`. Not killer, only showed up as `8.1u"μPa*s"` anyway so just replaced with `8.1e-6u"Pa*s"`

Yeah we can totally add that. My rule for adding units is basically just "does someone need it?" and then I add it. What I wanted to avoid was adding a ton of esoteric units that nobody used, which this system avoids.

src/utils.jl
```diff
@@ -388,6 +388,32 @@ Remove the units from a quantity.
...
+julia> ustrip(u"km", [1000u"m", 2000u"m"])
+2-element Vector{Float64}:
+ 1.0
+ 2.0
+```
+"""
+@inline function ustrip(unit::UnionAbstractQuantity, q::UnionAbstractQuantity)
+    unit, q = promote_except_value(unit, q)
+    dimension(unit) == dimension(q) || throw(DimensionError(unit, q))
+    return ustrip(q / unit)
```
MilesCranmer, line 414: Since we've already verified dimensions, I _think_ we can do

```julia
    return ustrip(q) / ustrip(unit)
```

which is faster

MilesCranmer:
Ok I have added a method for `ustrip` with affine units!

```julia
julia> ustrip(ua"degC", 27ua"degC")
27.0

julia> ustrip(ua"degF", 300u"K")
80.33000000000003
```

Ickaser:
> > * Error message for `ustrip(u"...", q)` should maybe say "you can only uconvert _or ustrip_ to quantities with SymbolicDimensions..."
> 
> I don't understand this question^ Can you clarify?

My original question here was unrelated to `ustrip` it turns out (just an error I ran into while moving to DQ). However, it does feel a little inconsistent to me in terms of API that e.g.
```
ustrip(u"km", 10u"cm") # ->0.0001
ustrip(u"m", 10u"cm")# -> 0.1
```
works, where `uconvert(u"km", 10u"cm")` errors. 

And in fact, although the behavior
```
ustrip(10u"km", 10u"cm") # -> 1.0e-5
```
might sometimes be desirable, it does feel odd to me. I think since my goal is to be sure of what the units are before stripping, I might prefer that `ustrip` take an object that represents only the units, not a quantity in those units.

That said, I don't feel so strongly about this that it should block the current API, just that it should be considered and perhaps documented that `ustrip` and `uconvert` have different behavior (in contrast to Unitful, where `ustrip(u, q)` calls `uconvert` then strips.)

Ickaser:
I think partly it's taking me a long time to adjust my thinking from Unitful, where the units can (and do) exist independently of Quantity objects. I see the good reasons for why DynamicQuantities works how it does but it might have helped me to have a docs section with some tips on how/why the behavior is different (perhaps I can write up some of the things that tripped me up).

MilesCranmer:
Yeah this is kind of the fundamental difference between the designs.

With `Dimensions`, it doesn't make sense to convert a quantity to non-SI units, because the non-SI units don't "exist" within `Dimensions`. There are only SI units, and everything else is just a multiplicative factor in front.

However, this is why `SymbolicDimensions` was created. That does let you do this type of thing:

```julia
julia> x = 1.5u"km"
1500.0 m

julia> uconvert(us"km", x)
1.5 km

julia> x |> us"km"  # shorthand
1.5 km
```

And it still gets the nice type stability properties of DQ.

However, you shouldn't do use symbolic dimensions internally—it's moreso for interactivity or display—because it's much less efficient than `Dimensions`.

MilesCranmer:
And I think `ustrip(unit, quantity)` actually makes sense to allow non-unit `Dimensions` quantities in the first argument purely because the output is not a quantity anyways.

PR #175, opened 2025-04-21
Update docs and add doctests

abhro:
Please see the commmit messages for relevant changes


docs/make.jl
```diff
@@ -31,14 +32,15 @@ end
...
-    repo="https://github.com/SymbolicML/DynamicQuantities.jl/blob/{commit}{path}#{line}",
+    repo=GitHub("SymbolicML/DynamicQuantities.jl"),
     sitename="DynamicQuantities.jl",
     format=Documenter.HTML(;
         prettyurls=get(ENV, "CI", "false") == "true",
         canonical="https://ai.damtp.cam.ac.uk/dynamicquantities/stable",
         edit_link="main",
         assets=String[]
     ),
+    doctest = false,
```
MilesCranmer, line 43: Doesn't this mean the doctests aren't ran?

docs/make.jl
```diff
@@ -31,14 +32,15 @@ end
 makedocs(;
     modules=[DynamicQuantities, DynamicQuantities.Units],
     authors="MilesCranmer <miles.cranmer@gmail.com> and contributors",
-    repo="https://github.com/SymbolicML/DynamicQuantities.jl/blob/{commit}{path}#{line}",
+    repo=GitHub("SymbolicML/DynamicQuantities.jl"),
```
MilesCranmer, line 35: What's the compat required for `GitHub` object?

docs/make.jl
abhro: For the `makedocs()` call, the doctests aren't ran, but there's already an existing call to `doctest()` near the top of doc/make.jl

docs/make.jl
abhro: In Documenter.jl's changelog, it seems they offered this starting 1.0.0 (https://github.com/JuliaDocs/Documenter.jl/blob/master/CHANGELOG.md#version-v100---2023-09-15, "The handling of remote repository...") so a compat specifier of "1" should work

MilesCranmer: Cool, thanks

docs/make.jl
MilesCranmer: I guess I'm confused why the CI isn't failing then. Because one of the doctests has an error at the moment. We should fix it but we can use it first to verify that the doctests are getting ran.

abhro: In the documentation CI step, it says the doctests did run and passed? I think I remember having local errors as well, so I'm not sure what the difference in the CI environment is.

src/utils.jl
```diff
@@ -395,14 +395,14 @@ Convert quantity `q` to the units specified by `unit`, then strip the units.
...
-```julia
+```jldoctest
 julia> ustrip(u"km", 1000u"m")
 1.0
 
 julia> ustrip(u"s", 1u"minute")
 60.0
 
-julia> ustrip(u"km", [1000u"m", 2000u"m"])
+julia> ustrip.(u"km", [1000u"m", 2000u"m"])
```
MilesCranmer, line 405: Doesn't need to be changed

docs/make.jl
MilesCranmer: Oh wait that's my bad, sorry! I didn't actually try the modified `ustrip.` example and forgot that syntax works.

src/utils.jl
MilesCranmer: Wait sorry it's fine as is!

abhro: Oh ok haha

MilesCranmer: I think I'm a bit sleep deprived this week, sorry for the repeated mistakes in this review. I was running it locally like

```julia
julia> ustrip(u"km", QuantityArray([1000u"m", 2000u"m"]))
2-element Vector{Float64}:
 1.0
 2.0
```

which works. But the Array one does not, as you correctly figured out!

abhro: No worries!

docs/make.jl
MilesCranmer, line 35: Does the `deploydocs` call later in the file (two of them) also need a `GitHub` object for consistency? I'm assuming the `repo` kwarg takes the same type?

docs/src/examples.md
```diff
@@ -183,6 +182,7 @@ function Base.convert(::Type{Quantity{T,AngleDimensions{R}}}, q::Quantity{<:Any,
         )
     )
 end
+nothing # hide
```
MilesCranmer, line 185: Can this be done with `end;` instead?

abhro: Maybe? I've seen the standard practice be a `nothing # hide` line, but an `end;` should also work. Checking.

src/register_units.jl
```diff
@@ -23,14 +23,14 @@ a particular value.
...
-```julia
+```jldoctest register-unit
 julia> @register_unit MyVolt 1.5u"V"
 ```
 
 This will register a new unit `MyVolt` with a value of `1.5u"V"`.
 You can then use this unit in your calculations:
 
-```julia
+```jldoctest register-unit
```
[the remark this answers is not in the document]
abhro: The names are here because they're in a shared environment. Otherwise, each doctest is run isolated. This lets us refer back variables defined in previous code blocks

docs/make.jl
abhro: It seems `deploydocs()` is asking for a string for `repo=` rather than a `GitHub` object (https://documenter.juliadocs.org/stable/lib/public/#Documenter.deploydocs)

docs/src/examples.md
abhro: They do seem to work. I've updated it on ee24a9f.

abhro:
Hi! I think I've addressed all the changes, it's most likely just the CI checks now

abhro:
> Have you been able to generate docs with the [`Mirror to DAMTP`](https://github.com/abhro/DynamicQuantities.jl/blob/033d428274044bdfd4c1a978a4a74a1eeb260977/docs/make.jl#L59-L65) bit btw? I've been finding that since it it [relies on an ENV in CI](https://github.com/SymbolicML/DynamicQuantities.jl/blob/b209b7c52d9a6bc9c14afddd49d06bcde435e724/.github/workflows/CI.yml#L60), it fails when I try to build locally, so I've just been commenting it our for now. I'm wondering if it would make sense to hide that bit behind a conditional flag that's called during CI?

I'm guessing this part is meant for @MilesCranmer? I'm not sure I can offer much help on the deployment side of things

icweaver:
Cool how about just wrapping the deploydocs bit in something like this?

```julia
in_CI_env = get(ENV, "CI", nothing) == "true"

if in_CI_env
    deploydocs(;
        repo="github.com/SymbolicML/DynamicQuantities.jl",
        devbranch="main"
    )

    # Mirror to DAMTP:
    ENV["DOCUMENTER_KEY"] = ENV["DOCUMENTER_KEY_CAM"]
    ENV["GITHUB_REPOSITORY"] = "ai-damtp-cam-ac-uk/dynamicquantities.git"
    deploydocs(;
        repo="github.com/ai-damtp-cam-ac-uk/dynamicquantities.git",
        devbranch="main"
    )
end
```

MilesCranmer:
That looks ok but the environment variable should be more specific. Like “DEPLOY_DOCUMENTER” or something. And just have the default be “false” rather than nothing

docs/make.jl
```diff
@@ -31,7 +31,7 @@ end
...
```
MilesCranmer, line 34: should be updated to new slug

icweaver: Hi @abhro, are you able to update this? I do not think that I have access from my end

abhro: When you say new slug, what's the proposed change? I'm not sure what I'm supposed to update it to 😅

icweaver:
Howdy folks, just wanted to check if there was anything else still needed for this PR. Did we want `in_CI_env` renamed to `DEPLOY_DOCUMENTER`? cc: @MilesCranmer

MilesCranmer (approved):
Thanks for the PR!

docs/Project.toml
icweaver, line 1: Hi Miles, I just noticed that you removed this. For docs development, is the preferred workflow to i) activate the docs environment, ii) dev DQ, and then iii) remove the DQ dep added to the docs Project.toml before committing changes upstream? This seems a bit circuitous to me, but I could be misunderstanding your intended workflow. Is there a better way that I should be doing this? Thanks!

MilesCranmer: Right. This is the standard workflow for Documenter.jl. If you put it in `docs/Project.toml`, even if you add a `[sources]` section, it will use the latest registered version of your package, rather than the code that actually lives in the directory. The only way to build the docs on the current version of the code is to dev the package. There is no way to indicate "dev MyPackage" in Project.toml. It has to be done via the Manifest - but the Manifest isn't stored, so the action has to do a dev command. Hope this helps

MilesCranmer: Specifying path is not the same as specifying the development version

PR #178, opened 2025-05-31
Additional `QuantityArray` constructions

icweaver:
Closes #166

**To-do**
- [x] docs
- [x] method ambiguities (remaining addressed in #181)

PR #181, opened 2025-08-02
Dates.jl + `QuantityArray`

icweaver:
Dependent PR split out from #178

Covers remaining method ambiguities introduced in that PR


icweaver:
False alarm, we can get away with just using tighter typing =] 046b8d4

Will go ahead and close this

PR #188, opened 2025-09-01
README + docs enhancement

icweaver:
Following up on https://github.com/JuliaPhysics/DynamicQuantities.jl/pull/178#issuecomment-3070575740

* Fixes anchor links issue (links in GitHub readme are case-insensitive while links in Documenter.jl are not)
    * I opted to not just titlecase the links in the source README.md doc to keep things consistent with how GitHub automatically lowercases anchor links
* Enables syntax highlighting for julia-repl blocks

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -4,6 +4,8 @@ using DynamicQuantities: constructorof, with_type_parameters, dimension_names
 using Documenter
 using Documenter.Remotes: GitHub
 
+import Base: titlecase
+
 DocMeta.setdocmeta!(DynamicQuantities, :DocTestSetup, :(using DynamicQuantities); recursive=true)
 
 readme = open(dirname(@__FILE__) * "/../README.md") do io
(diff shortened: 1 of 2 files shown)


docs/make.jl
```diff
@@ -16,6 +16,9 @@ readme = replace(readme, r"<img src=\"([^\"]+)\"[^>]+>.*" => s"![](\1)")
 # Then, we remove any line with "<div" on it:
 readme = replace(readme, r"<[/]?div.*" => s"")
 
+# Then, we titlecase the ToC entries to be compatible with Documenter.jl
+readme = replace(readme, r"\[(.*?)\]\(#(.*?)\)" => titlecase)
```
MilesCranmer, line 20: Where is `titlecase` defined? Is it some function? If so, maybe do an explicit import. I think for less-common imports it's very helpful for readability to have explicit imports

icweaver: For sure, it's exported in Base. Added an explicit import 2b397a6

MilesCranmer: Oh, sorry, I had no idea this was in Base... my bad. You can remove the import then

PR #189, opened 2025-09-09
prevent adding new methods to functions `>`, `>=`, `isgreater`

nsajko:
As documented in the doc string of each of these three functions, these functions are not intended to get any new methods from packages.

Preventing the definition of these unnecessary methods also prevents most of the sysimage invalidation that happens when loading this package. On nightly Julia, commit JuliaLang/julia@3681bb37ec7, the unique invalidation count upon `using DynamicQuantities` goes from $1381$ to $261$.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -208,7 +208,7 @@ Base.keys(q::UnionAbstractQuantity) = keys(ustrip(q))
 end
 
 # Numeric checks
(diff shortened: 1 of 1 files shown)


nsajko:
Data on each invalidation:

* [before.json](https://github.com/user-attachments/files/22244736/before.json)

* [after.json](https://github.com/user-attachments/files/22244738/after.json)

MilesCranmer:
Thanks!

PR #191, opened 2025-09-13
Create alias `FRInt32` for `FixedRational{Int32, 25200}`

MilesCranmer:
Makes the printouts easier to read. Also this creates `FRInt8=FixedRational{Int8,12}` which makes it less challenging to swap the dimensions type.

@icweaver would you be up for a review?

### DIFF
--- a/README.md
+++ b/README.md
@@ -457,9 +457,13 @@ the type you wish to use as the second argument to `Quantity`:
 ```julia
 julia> using DynamicQuantities
 
-julia> q8 = [Quantity{Float64,FRInt8}(randn(), length=rand(-2:2)) for i in 1:1000];
+julia> R8 = Dimensions{FRInt8};
 
-julia> q32 = [Quantity{Float64,FRInt32}(randn(), length=rand(-2:2)) for i in 1:1000];
+julia> R32 = Dimensions{FRInt32};
+
+julia> q8 = [Quantity{Float64,R8}(randn(), length=rand(-2:2)) for i in 1:1000];
+
+julia> q32 = [Quantity{Float64,R32}(randn(), length=rand(-2:2)) for i in 1:1000];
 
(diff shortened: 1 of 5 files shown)


icweaver:
Thanks Miles, love the simplifications here! Should this bit of the README also be updated now?

https://github.com/JuliaPhysics/DynamicQuantities.jl/blob/5668cf2c218291ea2bb4b337a0fa661cecc86864/README.md?plain=1#L331

```diff
- 3-element QuantityArray(::Vector{Float64}, ::Quantity{Float64, Dimensions{FixedRational{Int32, 25200}}}):
+ 3-element QuantityArray(::Vector{Float64}, ::Quantity{Float64, Dimensions{FRInt32}}):
```

I also did some reading on why the type alias printing tests are failing, and my understanding is that this is [actually by design](https://github.com/JuliaLang/julia/issues/40448)? I'm still getting familiar with TestItems, but are things failing because it is creating its own module, leading to the fully qualified types we're seeing in CI, but not interactively?

MilesCranmer:
I don't think so b/c these ones are normal `@testset` and `FRInt32` is in the namespace... Weird

MilesCranmer:
And thx, good catch

MilesCranmer:
This is what shows up in the REPL:

image

Idk. Maybe the testset turns this off somehow... But even doing `string(FRInt32)` works correctly elsewhere.

MilesCranmer:
Oh wait. You're totally right!!! My bad. Just saw this behavior:

```julia-repl
julia> module A
       using DynamicQuantities
       f() = println(FRInt32)
       end
Main.A

julia> A.f()
DynamicQuantities.FRInt32

julia> A.f()
FRInt32
```

Which is ... sorta weird. But I guess means the tests need to accommodate.

Issue #25, opened 2023-06-15
CGS units type / other type systems


MilesCranmer:
Some domains use non-SI unit systems. For example, astronomers like to use [CGS units](https://en.wikipedia.org/wiki/Centimetre%E2%80%93gram%E2%80%93second_system_of_units) (also known as Gaussian units). Although CGS units are falling out of use (and have been dead in most fields for a long time) it could be nice if these other domains could use DynamicQuantities.jl.

One should generally not mix cgs units with SI because:

1. There are only 3 base dimensions: length, mass, and time. All others are derived based on these.
2. Some equations need to be rewritten. i.e., you could not write one version of Maxwell's equations and have it work for both SI units and cgs units – there is an extra $4\pi$ factor in the cgs version.

I think all that would be required is a `CGSQuantity <: AbstractQuantity` and `CGSDimensions <: AbstractQuantity` (requiring #24), with one fewer dimension.

However, maybe it's just easier to define `erg` as one of the available units, and be clear that the SI convention is used.

PR #196, opened 2025-11-25
Adding some more astro units

icweaver:
Following up from #25

Added a few cgs and other units that show up in astro fairly regularly


icweaver:
Gentle ping @MilesCranmer

icweaver:
Happy new year! Just checking back in on if anything else is needed for this

PR #215, opened 2026-04-22
fix: preserve symbolic external units across precompile

### LINKED ISSUE #212: Proper way to support "g" (acceleration) unit
Hello, I'm fiddling with a lot of different acceleration units, and I was wondering what would be the best way to support all of them with `DynamicQuantities`. 
I am pretty sure this is out of the scope of the package as it clearly conflicts with the "g" (gram) unit, although both are derived units. 
However, I am wondering how to support this at my level without generating conflicts. 
Thanks.

MilesCranmerBot:
## Summary
Preserve external symbolic unit registrations across precompile and import boundaries.

## Changes
- thread the caller module through `us"..."` symbolic macro expansion
- lazily re-register external bindings when a symbolic unit exists in the caller module but the symbolic registry has not been rebuilt yet
- add a regression test covering the precompile/import case for external units

## Testing
- `/usr/local/bin/julia --project -e 'using Pkg; Pkg.test()'`

Fixes #212


MilesCranmer:
@MilesCranmerBot we need to get to 100% coverage of the diff please

test/precompile_test/ExternalUnitRegistration.jl
```diff
@@ -7,15 +7,21 @@ using Test
...
+
 @testset "Register Unit Inside a Module" begin
     for collection in (UNIT_SYMBOLS, ALL_SYMBOLS, keys(ALL_MAPPING._raw_data), keys(UNIT_MAPPING._raw_data))
         @test :MyWb ∈ collection
     end
 
-    w = u"MyWb"
-    ws = us"MyWb"
+    w = expanded_mywb()
+    ws = symbolic_mywb()
```
MilesCranmer: @MilesCranmerBot preferably this would not break, why is this needed? We can't make breaking changes please

MilesCranmer:
@MilesCranmerBot code coverage is not yet 100% for the diff. And also see comment about making breaking changes please.

test/precompile_test/ExternalUnitRegistration.jl
MilesCranmerBot: The added `expanded_mywb()` / `symbolic_mywb()` helpers are just there to exercise the actual failing path, not because the API is supposed to change.

The breakage only showed up once the symbolic unit was referenced from compiled code after import/precompile. A top-level `us"MyWb"` inside the module body still evaluates while the module is being built, so it was not catching the bug we care about. By contrast, `u"MyWb"` bakes the expanded quantity directly into the compiled code, while `us"MyWb"` was still depending on the symbolic registry being present again at runtime.

So the intent of this PR is not to make users rewrite code or change registered-unit semantics, only to make `us"..."` keep working for externally registered units across the precompile/import boundary. The function wrappers in the test are just the smallest regression I found that actually reproduces that path.

MilesCranmer: @MilesCranmerBot please revert this change. If you want to hit the part of the code you should add a new test

MilesCranmerBot: Reverted the change to the existing precompile test and left the dedicated wrapper regression test in `test/unittests.jl` instead. Verified with `/usr/local/bin/julia --project=. -e 'using Pkg; Pkg.test(test_args=["unittests"])'

MilesCranmer:
@MilesCranmerBot needs to add a test where we have a fake package that registers units, and that has a function which uses those units. See the recent issue for the workflow I'd like to enable.

MilesCranmer:
@codex review

MilesCranmer:
@MilesCranmerBot Please address the above feedback

src/symbolic_dimensions.jl
```diff
@@ -495,39 +504,55 @@ module SymbolicUnits
...
+    function map_to_scope(mod::Module, sym::Symbol)
         if sym in UNIT_SYMBOLS
             # return at end
         elseif sym in CONSTANT_SYMBOLS
             throw(ArgumentError("Symbol $sym found in `Constants` but not `Units`. Please use `us\"Constants.$sym\"` instead."))
+        elseif _has_quantity_binding(mod, sym)
+            # return at end
         else
             throw(ArgumentError("Symbol $sym not found in `Units` or `Constants`."))
         end
```
MilesCranmer: @MilesCranmerBot 
A bit messy. Would be nicer if had negated if statements like `!(sym in UNIT_SYMBOLS)` etc. And only two branches.

MilesCranmerBot: Done, I collapsed this to two branches using the negated `!(sym in UNIT_SYMBOLS)` guard and kept the success path as the shared return at the end. Pushed in dde2975, and `Pkg.test()` passed locally.

src/symbolic_dimensions.jl
```diff
@@ -495,39 +504,53 @@ module SymbolicUnits
...
+        return nothing
     end
-    function lookup_constant(ex::Symbol)
-        i = findfirst(==(ex), CONSTANT_SYMBOLS)::Int
-        return as_quantity(SYMBOLIC_CONSTANT_VALUES[i])
+    function lookup_unit(mod::Module, ex::Symbol)
+        _ensure_registered(mod, ex)
+        return as_quantity(symbolic_unit_from_symbol(ex))
     end
+    lookup_unit(ex::Symbol) = lookup_unit(@__MODULE__, ex)
```
MilesCranmer: @MilesCranmerBot I'm a bit uncomfortable with all of the dynamism going on here. Stuff like `Base.invokelatest(getproperty, mod, sym)` is being used on a call that will occur quite frequently (looking up a constant value). Surely there's a cleaner way to do all of this? It feels quite messy at the moment and I wouldn't be comfortable merging this.

MilesCranmerBot: Good call. I cleaned this up in 982937d.

The lazy external-unit registration path is still there, but the dynamic bits are a lot smaller now:
- no `Base.invokelatest(getproperty, mod, sym)`
- no `Base.invokelatest` on the updater either
- direct `Core.getglobal(mod, sym)` for the module binding read
- regular call back into `update_all_values_unlocked`

So we keep the precompile/import fix, but the lookup path is less magical and less messy.

I also reran `Pkg.test(test_args=["unittests"])`, including the external-unit precompile regression, and it passed.

MilesCranmer: @MilesCranmerBot `Core.getglobal(mod, sym)` is the same and arguably even worse. Please clean up

MilesCranmerBot: Cleaned this up in b06c809.

The symbolic-unit path now treats built-in units separately, captures external quantity bindings directly via `GlobalRef(mod, sym)` and `getfield`, and registers from the captured unit instead of doing a reflective module lookup in the hot path.

I also reran `/usr/local/bin/julia --project=. -e 'using Pkg; Pkg.test(test_args=["unittests"])'` after rebasing, and it passes locally.

MilesCranmer:
@codex review

MilesCranmer:
Hm, I still think this feels too hacky... @MilesCranmerBot could you please change the PR in the following way? Here's a detailed description:

> This can be fixed in a cleaner and more symmetric way.
> 
> The current PR appears to solve the immediate `us"..."` failure across precompile/import boundaries, but it does so by making only the symbolic parser caller-module-aware. That leaves `u"..."` and `us"..."` on different resolution paths:
> 
> - `u_str` still expands through `map_to_scope(Meta.parse(s))`
> - `us_str` is changed to expand through `map_to_scope(__module__, Meta.parse(s))`
> 
> That asymmetry is hard to justify at the parser API level. External units should have one consistent resolution story, regardless of whether the target is ordinary or symbolic.
> 
> A cleaner fix would be:
> 
> 1. Make both macros caller-module-aware.
>    - Change `@u_str` in `src/uparse.jl` to call `map_to_scope(__module__, Meta.parse(s))`
>    - Keep the same for `@us_str`
> 
> 2. Add matching module-aware `map_to_scope(mod::Module, ...)` overloads on both the ordinary and symbolic sides.
>    - `map_to_scope(ex::Expr) = map_to_scope(@__MODULE__, ex)`
>    - `map_to_scope(sym::Symbol) = map_to_scope(@__MODULE__, sym)`
>    - `map_to_scope(mod::Module, ex::Expr)` for recursive descent
>    - `map_to_scope(mod::Module, sym::Symbol)` for symbol resolution
> 
> 3. Share the external-binding detection and lazy registration logic instead of implementing it only for symbolic parsing.
>    - The external binding check should live in shared internal code
>    - The lazy registration path should also be shared
>    - Both `u"..."` and `us"..."` should be able to recover an external unit from the caller module when the binding exists but the registration state has not yet been rebuilt in the current import/precompile lifecycle
> 
> 4. Keep symbolic lookup independent of `SYMBOLIC_UNIT_VALUES[i]` cache alignment for correctness.
>    - Reconstructing the symbolic quantity from the symbol is a reasonable approach
>    - That part of the current PR looks directionally good
>    - However, it should be part of a symmetric design rather than paired with a parser-only special case for `us_str`
> 
> 5. Add regression tests for both ordinary and symbolic macros through an external module boundary.
>    - `u"MyExternalUnit"`
>    - `us"MyExternalUnit"`
>    - direct use
>    - imported helper function use
>    - precompile/import style loading
>    - threaded first-use if concurrency is relevant
> 
> The key point is that the problem does not appear to be that symbolic registration should stop being updated. The issue is that symbolic parsing currently has a more fragile lifecycle around external registrations, especially across precompile/import boundaries. The fix should therefore make the registration and recovery story more robust, but without introducing a macro-level asymmetry between `u_str` and `us_str`.
> 
> In short, the suggested direction is:
> 
> - make `u_str` and `us_str` symmetric
> - share external-unit recovery logic
> - keep symbolic lookup robust to cache rebuild timing
> - test both macros the same way
> 
> That should produce a clearer and more maintainable design than fixing only `us_str` in isolation.

MilesCranmer:
@MilesCranmerBot CI is failing, please fix.

MilesCranmer:
@MilesCranmerBot code coverage of the diff is under 100%. Are there methods which aren't possible to hit in a realistic scenario? If so then please remove them. If it is possible to hit them, then ensure 100% diff coverage with testing.

src/symbolic_dimensions.jl
```diff
@@ -458,15 +472,12 @@ module SymbolicUnits
 
     update_symbolic_unit_values!(w::WriteOnceReadMany) = update_symbolic_unit_values!.(w._raw_data)
     update_symbolic_unit_values!(UNIT_SYMBOLS)
+    const BUILTIN_UNIT_SYMBOLS = Tuple(UNIT_SYMBOLS._raw_data)
```
MilesCranmer: @MilesCranmerBot how about defining BUILTIN_UNIT_SYMBOLS at the definition site, and then informing UNIT_SYMBOLS as a copy of it... Instead of this hacky solution

src/symbolic_dimensions.jl
```diff
@@ -495,39 +506,40 @@ module SymbolicUnits
...
+        if sym in BUILTIN_UNIT_SYMBOLS
+            return Expr(:call, GlobalRef(@__MODULE__, :lookup_unit), QuoteNode(sym))
         elseif sym in CONSTANT_SYMBOLS
             throw(ArgumentError("Symbol $sym found in `Constants` but not `Units`. Please use `us\"Constants.$sym\"` instead."))
+        elseif !(mod === @__MODULE__) && _external_quantity_binding(mod, sym)
+            return Expr(:call, GlobalRef(@__MODULE__, :lookup_external_unit), QuoteNode(sym), GlobalRef(mod, sym))
         else
             throw(ArgumentError("Symbol $sym not found in `Units` or `Constants`."))
         end
-        return lookup_unit(sym)
```
MilesCranmer: I think it hurts readability to have mix of `returns` and `throw`. Instead, you should have the `throw`s within one block. And if none of the if branches get hit, then it hits the returns block. Much cleaner that way.

src/uparse.jl
```diff
@@ -23,6 +24,7 @@ macro generate_units_import()
 end
 
 @generate_units_import
+const BUILTIN_UNIT_SYMBOLS = Tuple(UNIT_SYMBOLS._raw_data)
```
MilesCranmer: Uh, this was defined multiple times? Code smell...

src/uparse.jl
```diff
@@ -59,38 +61,62 @@ the quantity corresponding to the speed of light multiplied by Hertz,
...
+function map_to_scope(mod::Module, sym::Symbol)
+    if sym in BUILTIN_UNIT_SYMBOLS
+        return Expr(:call, GlobalRef(@__MODULE__, :lookup_unit), QuoteNode(sym))
     elseif sym in CONSTANT_SYMBOLS
         throw(ArgumentError("Symbol $sym found in `Constants` but not `Units`. Please use `u\"Constants.$sym\"` instead."))
+    elseif !(mod === @__MODULE__) && _external_quantity_binding(mod, sym)
+        return Expr(:call, GlobalRef(@__MODULE__, :lookup_external_unit), QuoteNode(sym), GlobalRef(mod, sym))
     else
         throw(ArgumentError("Symbol $sym not found in `Units` or `Constants`."))
     end
```
MilesCranmer: Similar comment about `throw` and `return` being in separate chunks of the code rather than mixed up like this.

src/uparse.jl
```diff
@@ -59,38 +61,62 @@ the quantity corresponding to the speed of light multiplied by Hertz,
...
-    i = findfirst(==(ex), UNIT_SYMBOLS)::Int
-    return UNIT_VALUES[i]
+map_to_scope(::Module, ex) = ex
+
+function _external_quantity_binding(mod::Module, sym::Symbol)
+    return isdefined(mod, sym) && getfield(mod, sym) isa UnionAbstractQuantity
+end
+function _ensure_registered_external_unit(sym::Symbol, unit::UnionAbstractQuantity)
+    unit_update_lock = getfield(parentmodule(@__MODULE__), :UNIT_UPDATE_LOCK)
+    update_all_values_unlocked = getfield(parentmodule(@__MODULE__), :update_all_values_unlocked)
```
MilesCranmer: `getfield(parentmodule(@__MODULE__), :UNIT_UPDATE_LOCK)` is a really bad code smell. Something has gone wrong while writing this it seems...

MilesCranmer (changes requested):
@MilesCranmerBot see new comments

src/uparse.jl
MilesCranmer: I guess this is trying to do an import??

src/symbolic_dimensions.jl
MilesCranmerBot: Yep, agreed. I moved `BUILTIN_UNIT_SYMBOLS` into `Units` at the definition site and now import that shared snapshot from both parse paths instead of recomputing it locally.

Pushed in c5c1868, and `/usr/local/bin/julia --project=. -e 'using Pkg; Pkg.test()'` passed locally.

src/external_unit_lookup.jl
MilesCranmer, line 1: Doesn't need to be a separate file.

MilesCranmer:
@codex review

src/uparse.jl
```diff
@@ -59,38 +62,50 @@ the quantity corresponding to the speed of light multiplied by Hertz,
...
+
+    if !has_builtin_binding && sym in CONSTANT_SYMBOLS
         throw(ArgumentError("Symbol $sym found in `Constants` but not `Units`. Please use `u\"Constants.$sym\"` instead."))
-    else
+    elseif !has_builtin_binding && !has_external_binding
         throw(ArgumentError("Symbol $sym not found in `Units` or `Constants`."))
     end
+
+    if has_builtin_binding
+        return Expr(:call, GlobalRef(parentmodule(@__MODULE__), :lookup_registered_unit), QuoteNode(sym))
```
MilesCranmer: Bad code smell on this line. Looks like something that should have been an import...

src/uparse.jl
```diff
@@ -59,38 +62,50 @@ the quantity corresponding to the speed of light multiplied by Hertz,
...
+    else
+        return Expr(:call, GlobalRef(@__MODULE__, :lookup_external_unit), QuoteNode(sym), GlobalRef(mod, sym))
```
MilesCranmer: Similar comment here

src/symbolic_dimensions.jl
```diff
@@ -495,39 +506,45 @@ module SymbolicUnits
...
-        return ex
-    end
-    function lookup_unit(ex::Symbol)
-        i = findfirst(==(ex), UNIT_SYMBOLS)::Int
-        return as_quantity(SYMBOLIC_UNIT_VALUES[i])
+
+        if has_builtin_binding
+            return Expr(:call, GlobalRef(@__MODULE__, :lookup_unit), QuoteNode(sym))
+        else
+            return Expr(:call, GlobalRef(@__MODULE__, :lookup_external_unit), QuoteNode(sym), GlobalRef(mod, sym))
```
MilesCranmer: See other comments

src/symbolic_dimensions.jl
```diff
@@ -495,39 +506,45 @@ module SymbolicUnits
...
             throw(ArgumentError("Unexpected expression: $ex. Only `:call` and `:.` (for `SymbolicConstants`) are expected."))
         end
         if ex.head == :call
-            ex.args[2:end] = map(map_to_scope, ex.args[2:end])
+            ex.args[2:end] = map(arg -> map_to_scope(mod, arg), ex.args[2:end])
             return ex
         else # if ex.head == :. && ex.args[1] == :Constants
             @assert ex.args[2] isa QuoteNode
-            return lookup_constant(ex.args[2].value)
+            return Expr(:call, GlobalRef(@__MODULE__, :lookup_constant), QuoteNode(ex.args[2].value))
```
MilesCranmer, line 518: See other comments

MilesCranmer (changes requested):
@MilesCranmerBot Please address these comments + those of codex bot + ensure that code coverage for the diff is 100% (failing CI at the moment)

MilesCranmer:
@codex review

PR #221, opened 2026-05-12
feat: add THz

icweaver:
Howdy, we needed this unit for an astropy tutorial on [analyzing spectra](https://learn.astropy.org/tutorials/units-and-integration.html) that we are translating for our [learn.JuliaAstro](https://learn.juliaastro.org/) site. Would it be ok if we add this?


MilesCranmer (approved):

icweaver:
thx!

